PR #51, opened 2019-03-15
If file not choosen getFile() should return null

pjkroll:


miguelpruivo:
Hello. Glad you’ve spotted it. Can you please change the base branch to `beta` instead? Thank you.

Issue #58, opened 2019-03-21
Selecting a single image from Google Photos does not work


Boehrsi:
If FilePicker.getFilePath() is used and an image is added via the default file chooser + Google Photos an error occurs.

`E/flutter: [ERROR:flutter/lib/ui/ui_dart_state.cc(148)] Unhandled Exception: type 'List<dynamic>' is not a subtype of type 'String'`
`#0      FilePicker.getFile (package:file_picker/file_picker.dart:42:18)`

The reason for this problem is Google Photos and not this plugin. Even if only one file is requested Google Photos returns a list of elements, with one entry. It seems like a good idea to verify the returned data and wrap it if necessary. I will attach a PR which fixes the problem.

miguelpruivo:
Thank you for your input, can you post your full log and the steps to reproduce?

Boehrsi:
1. Call await FilePicker.getFilePath(type: fileType, fileExtension: extension);
2. Select Google Photos (see Screenshot)
3. Select any image
4. See error

![Screenshot_20190321-144535_Files](https://user-images.githubusercontent.com/7944013/54756749-2616c380-4be9-11e9-8ff2-cbf4375a379a.jpg)

Error:
```
2019-03-21 14:47:27.713 4484-4586/com.openxchange.oxtalk E/flutter: [ERROR:flutter/lib/ui/ui_dart_state.cc(148)] Unhandled Exception: type 'List<dynamic>' is not a subtype of type 'FutureOr<String>'
#0      FilePicker.getFilePath (package:file_picker/file_picker.dart:35:7)
<asynchronous suspension>
#1      _BlocScreenState._getFilePath (package:justAnApp/src/some_code/just_a_file.dart:401:40)
<asynchronous suspension>
#2      _BlocScreenState._showAttachmentChooser.<anonymous closure>.<anonymous closure> (package:justAnApp/src/some_code/just_a_file.dart:371:28)
#3      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:513:14)
#4      _InkResponseState.build.<anonymous closure> (package:flutter/src/material/ink_well.dart:568:30)
#5      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:120:24)
#6      TapGestureRecognizer._checkUp (package:flutter/src/gestures/tap.dart:242:9)
[...]

PR #59, opened 2019-03-21
Selecting a single image from Google Photos does not work #58

Boehrsi:
Fix for https://github.com/miguelpruivo/plugins_flutter_file_picker/issues/58

I just added a small check if _isMultipleSelection_ and "map" the list to a String if not.

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
@@ -60,7 +60,12 @@ public class FilePickerPlugin implements MethodCallHandler {
               Log.i(TAG, "[MultiFilePick] File #" + currentItem + " - URI: " +currentUri.getPath());
               currentItem++;
             }
-            result.success(paths);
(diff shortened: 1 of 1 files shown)


miguelpruivo:
This is should be protected in the Android side like is being already done [in iOS](https://github.com/miguelpruivo/plugins_flutter_file_picker/blob/43796d2a2186b9d5c82dd383ab7dc7c276ceab51/ios/Classes/FilePickerPlugin.m#L141) whereas a single String is being returned if the list contains only one item.

Boehrsi:
> This is should be protected in the Android side like is being already done [in iOS](https://github.com/miguelpruivo/plugins_flutter_file_picker/blob/43796d2a2186b9d5c82dd383ab7dc7c276ceab51/ios/Classes/FilePickerPlugin.m#L141) whereas a single String is being returned if the list contains only one item.

Yep would also work for that case :+1:

miguelpruivo:
This has now been fixed in [1.3.3](https://pub.dartlang.org/packages/file_picker#-changelog-tab-) by protecting in the Android native callback so this PR is not longer required. 

Anyhow, thank you for contributing.

PR #110, opened 2019-07-19
Fixes #107

### LINKED ISSUE #107: NullPointerException on FilePickerPlugin$1.onActivityResult in certain Android devices
My app is crashing in some Android devices with a NullPointerException.

Unfortunately I've been unable to reproduce the issue in my device. The crash logs came from Google's Developer Console and points to a NPE in `FilePickerPlugin$1.onActivityResult` but without a line number.

Taking a look into this method on version 1.3.7 pointed me to this suspect piece of code. On line 59  of `com.mr.flutter.plugin.filepicker.FilePickerPlugin` you have `if(data.getClipData() != null) {` yet on line 78 you have an else for that if written as `} else if (data != null) {`. This `else` block won't ever be reached and Android Studio even warns about that.

I don't know if `data` being null on line 59 is the cause of this NPE but the rest of the code seems fine so I would start taking a look there.

Below is the stack trace of the crash. If I'm able to reproduce this in my device I'll post more details.

> java.lang.RuntimeException: 
>   at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3469)
>   at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:3509)
>   at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2772)
>   at android.app.ActivityThread.-wrap12 (ActivityThread.java)
>   at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1507)
>   at android.os.Handler.dispatchMessage (Handler.java:102)
>   at android.os.Looper.loop (Looper.java:154)
>   at android.app.ActivityThread.main (ActivityThread.java:6236)
>   at java.lang.reflect.Method.invoke (Method.java)
>   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:891)
>   at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:781)
> Caused by: java.lang.RuntimeException: 
>   at android.app.ActivityThread.deliverResults (ActivityThread.java:4141)
>   at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3443)
>   at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:3509)
>   at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2772)
>   at android.app.ActivityThread.-wrap12 (ActivityThread.java)
>   at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1507)
>   at android.os.Handler.dispatchMessage (Handler.java:102)
>   at android.os.Looper.loop (Looper.java:154)
>   at android.app.ActivityThread.main (ActivityThread.java:6236)
>   at java.lang.reflect.Method.invoke (Method.java)
>   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:891)
>   at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:781)
> Caused by: java.lang.NullPointerException: 
>   at com.mr.flutter.plugin.filepicker.FilePickerPlugin$1.onActivityResult (FilePickerPlugin.java)
>   at io.flutter.app.FlutterPluginRegistry.onActivityResult (FlutterPluginRegistry.java)
>   at io.flutter.app.FlutterActivityDelegate.onActivityResult (FlutterActivityDelegate.java)
>   at io.flutter.app.FlutterActivity.onActivityResult (FlutterActivity.java)
>   at br.com.cosmiceffect.app.cosmiccollection.MainActivity.onActivityResult (MainActivity.java)
>   at android.app.Activity.dispatchActivityResult (Activity.java:6937)
>   at android.app.ActivityThread.deliverResults (ActivityThread.java:4137)
>   at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3443)
>   at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:3509)
>   at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2772)
>   at android.app.ActivityThread.-wrap12 (ActivityThread.java)
>   at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1507)
>   at android.os.Handler.dispatchMessage (Handler.java:102)
>   at android.os.Looper.loop (Looper.java:154)
>   at android.app.ActivityThread.main (ActivityThread.java:6236)
>   at java.lang.reflect.Method.invoke (Method.java)
>   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:891)
>   at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:781)

Dancovich:
Fixes the issue of dealing with some buggy Android file managers that return from the `Activity` with a `RESULT_OK` but no data, causing a NPE.


android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
```diff
@@ -117,7 +123,7 @@ public boolean onRequestPermissionsResult(int requestCode, String[] strings, int
 
   @Override
   public void onMethodCall(MethodCall call, Result result) {
-    this.result = result;
+    FilePickerPlugin.result = result;
```
miguelpruivo: @Dancovich why this change? Do you find it more readable?

Dancovich: In Java static fields are supposed to be accessed through their classes. The original call is actually translated to `FilePickerPlugin.result` by the compiler and lint gives a warning about that.

One thing I noticed is that the `result` field is static in this class. Is there a particular reason for that? It seems to me Flutter will create a new instance of the plugin when it calls `FilePicker.registerWith` and I assumed it would create one instance of the plugin for the entire app life cycle, so the `result` field could be an instance field to `FilePickerPlugin`.

Anyway, since this was way out of scope for this bug I just changed the instance access to a class access like lint suggests, but this change won't actually affect the code in anyway if it's reverted.

miguelpruivo: Yes, I know that. I only questioned because I was wondering if you changed it to be more accordingly to Java practices or if it somehow was related to the purposed fix, which would surprise me. 😁

Anyway, could you please change the target branch to `beta` instead? I usually merge to master only after testing and when I’m about to publish on pub.dev as well.

Thank you!

Dancovich:
Done: #111 
I'll close this request then

PR #180, opened 2019-11-20
fix usage sample for single file

lecaros:
Sample for single file is the same than multiple files. This is a minor fix but could confuse someone.


miguelpruivo:
Thank you for noticing @lecaros, it's suppose to be like that, it was a copy/paste daw.

PR #231, opened 2020-04-07
Ignore permission requests from other plugins

wardes:
In my app, the file picker popup was getting triggered from use of the permission_handler plugin to request WRITE_EXTERNAL_STORAGE permissions.

This small change checks if requestCode matches file_picker's own code, and ignores the call to onRequestPermissionsResult if they don't match.


miguelpruivo:
Thank you @wardes. Good catch!

PR #256, opened 2020-04-27
import fix

ethael:
when building for web, it fails on import mismatch, because file_picker.dart is taking File object from dart:io but your web implementation is taking it from dart:html. It is correctly declared in file_picker_stub.dart, but not here


miguelpruivo:
@ethael thank you! I'm still working on this and probably a few changes will come in play, but I'll merge yours as of now nevertheless.

Issue #233, opened 2020-04-08
[Feature] Select multiple images/videos from gallery app on iOS


miguelpruivo:
Self explanatory.

DebugTheCode:
Great. Awaiting for updates :)

jamesdixon:
@miguelpruivo thanks for this awesome package!

Interestingly enough, I landed here because I was looking for this very feature 😄 

Do you have any idea of the perceived effort involved for this?

Cheers.

miguelpruivo:
@jamesdixon thanks! Actually, I'm afraid that not much. I was thinking on also supporting compression parameters hence the delay, but for the time being, I guess I'll implement it right away so people can start picking multiple pictures/videos from gallery with original quality and leave the compression for now (after all, there are other packages that can be used to compression afterwards with a lot of options).

DebugTheCode:
There are some compression libraries (that use native compression) that are very easily to use and therefor also good to combine with your package. Compression straight out of your package would be a nice feature tho.
Let me know where I can buy you a cup of coffee.

jamesdixon:
@miguelpruivo sweet! Yes, honestly, the multiple picking feature is something that it seems a lot of people are looking for. There are a couple of other libraries that do it, but your library is the cleanest and most versatile with regards to picking files, so having multi-select would be 🔥 

More than happy to help where I can!

By the way, saw you're from Portugal. My wife and I had a trip planned for May and it obviously was cancelled. We're so bummed. We hear nothing but amazing things about your country!

miguelpruivo:
@jamesdixon thank you! It turns out that I've been working on the implementation for this since yesterday and I thought that it would be out-of-the-box but then found out that on iOS, the `ImagePickerViewController` doesn't allow you to pick multiple items, thus, I'm working around that — probably will use a good iOS lib with the features I need or else will have to create my own. Nevertheless, I think I'll have this on the plugin really soon (will work a bit on it today as well).

As for Portugal, you are right, this virus got us all and I still definitely recommend you to come when all of this goes away, but preferably, by the spring/summer, where you can really take advantage of this amazing country weather, beaches, food and people, of course.

jamesdixon:
@miguelpruivo thanks for the update! Did you end up picking an existing iOS library?

Yes, we're definitely planning on rebooking our trip for next spring. Stay safe!

miguelpruivo:
[...]

PR #258, opened 2020-04-29
fix conflict with flutter_image_cropper

jamesdixon:
This resolves the [issue](https://github.com/miguelpruivo/flutter_file_picker/issues/233#issuecomment-620264779) we discussed regarding [flutter_image_cropper](https://pub.dev/packages/image_cropper) and [DKImagePicker](https://github.com/zhangao0086/DKImagePickerController) both importing [TOCropViewController](https://github.com/TimOliver/TOCropViewController) and causing a conflict.

`DKImagePicker` exposes subspecs -- from those, I was able to select this package to only import the `PhotoGallery` subspec, which excludes `TOCropViewController`.

I'm new to this world of Pods, etc so if I messed anything up, be gentle 🤣

### DIFF
--- a/ios/file_picker.podspec
+++ b/ios/file_picker.podspec
@@ -15,7 +15,7 @@ A new flutter plugin project.
   s.source_files = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
-  s.dependency 'DKImagePickerController'
+  s.dependency 'DKImagePickerController/PhotoGallery'
   s.ios.deployment_target = '8.0'
 end
 


miguelpruivo:
Wow! That’s huge @jamesdixon, you can’t imagine how hard I tried to find a way to only select that subspec _only_ without success and it turns out that you need to use only the path directly. May I ask you where did you find that info? I’ve googled a lot of things and neither provide that answer 😂

Thanks!

jamesdixon:
@miguelpruivo honestly, it was pure luck! I kept finding things about creating subspecs but not consuming them. Then I randomly stumble on [this](https://github.com/RestKit/RestKit#via-cocoapods) and realized that you could reference by path.

It was NOT obvious in any documentation so I understand why you couldn't find anything!

Anyways, glad it worked and thank you again for your work on this. It's invaluable. If I ever make it to Portugal, I owe you a drink!

PR #291, opened 2020-06-14
Fix error when closing app on Android

DEMndz5:
Fixes error: java.lang.RuntimeException: Unable to destroy activity {package.name/package.name.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mr.flutter.plugin.filepicker.FilePickerDelegate.setEventHandler(io.flutter.plugin.common.EventChannel$EventSink)' on a null object reference


miguelpruivo:
Good catch, thank you.

PR #298, opened 2020-06-18
Fix NullPointerException in tearDown()

josephx86:


josephx86:
When starting an app, the UI hangs when tearDown() crashes. I checked the event logs.

josephx86:
I used jadx and information from logs to pinpoint the line causing the crash

miguelpruivo:
Do you have the logs? It's weird because it shouldn't ever be `null`.

josephx86:
that is what I thought too. It should never be null but the logs are showing it is. I have screenshots.

josephx86:
.... and from jadx

miguelpruivo:
Does that happen on multiple devices?

josephx86:
The problem only happens when an app is installed from the Play Store in my case, so that is why I ended up tracking this issue hacky way reversing an apk.

I have this issue on a Galaxy Note9, Nexus 5X, and a Blu Advance 5.2 (A290Q)

miguelpruivo:
@josephx86 which Flutter version were used to build that release?

josephx86:
I tested with both stable and dev last week. At the moment is am using dev 1.20.0-1.0.pre

miguelpruivo:
@josephx86 can you try with latest 1.12.0 running on Flutter stable and let me know if you are still experiencing the same issue? Thanks.

vlefort:
I i'm also facing the same issue.
image

All work on emulator but not when my user use the Play store App...

I was with flutter dev channel, I try to use flutter stable channel (1.20.4) and give use a feedback when published.

miguelpruivo:
@vlefort does that happen with latest version (2.0.0)?

vlefort:
Stangely it looks the same. 
image.

Flutter 1.20.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision fba99f6cf9 (3 days ago) • 2020-09-14 15:32:52 -0700
Engine • revision d1bc06f032
Tools • Dart 2.9.2

  file_picker: ^2.0.0

miguelpruivo:
I'll eventually merge this, however, with the uncertain of what might be causing it. It shouldn't be `null` at that point.

miguelpruivo (approved):

miguelpruivo:
Fix added. Closing.

Issue #234, opened 2020-04-08
[Feature] Support scoped storage on Android


miguelpruivo:
# Motivation
[Scoped storage](https://developer.android.com/training/data-storage#scoped-storage) was introduced with API 29 and was motived to abstract devs from directly handling absolute paths.

I'm aware that as convenience, having absolute paths allows you to instance a `File` object directly on Flutter and handle it from there, however, this is not always possible and cached files (copies) act as fallback. 

This may result in some devs being confused whether they are picking the original file or copies and also, in the later, prevents original file from being manipulated directly. Supporting scoped storage will allow files to always being directly manipulated through platform calls, instead of creating a `File` instance.

A lot of discussion has been made around this on some issues, such as #169. This solution aims to act as an alternative and absolute paths will still exist for those that want or below API 29.

hpoul:
Out of curiosity, do you already have any plans how to implement this (API wise), or have you started?
I'm thinking about having a go on it as well.. I have a few user reports ([@AuthPass](https://github.com/authpass/authpass/)) because of that problem and not much to solve it without touching the plugin.

miguelpruivo:
@hpoul yes, I have been drafting something but's not yet on origin, this is on my top priorities alongside with #229 and #233, so that's probably next on the line.

hpoul:
that is great to hear, so i'll just wait it out.. one thing, do you know that the same exact problem exists on iOS devices? Right now the plugin uses: `UIDocumentPickerModeImport` when opening `UIDocumentPickerViewController`. This basically only returns a copy in the applications `Inbox` directory. As far as I understand it, to be able to read and (re)write. it, one would have to use `UIDocumentPickerModeImport` and then [use secure bookmarks](https://developer.apple.com/documentation/foundation/nsurl/1417795-bookmarkdata) to keep the. permission to write the data on subsequent launches.

So maybe when planning the API for android it would make sense to keep this in mind.

miguelpruivo:
@hpoul good point. Anyway, I’m not sure if that will be necessary on iOS side as you _always_ an absolute path to handle. That makes it easier for Dart to manipulate a `File` object directly.

hpoul:
[...]

PR #328, opened 2020-07-18
Web support serialize file object

lennhv:
Hi Miguel, i implemented the function requested in number 326, tested it with Chrome and Firefox with successful results.

Changes:
- Create a class to wrap the `html.File` named `File` with a method named `fileAsBytes` to request the serialized object.
- Change the type object from `html.File` to File in all plugin code.
- Add in the sample code an example of using the new function.

These changes allow to get the file at a low level and decouple it from the platform type, also to avoid importing the `dart: html` package outside the web plugin when the development is cross-platform.

Regards.

### DIFF
--- a/file_picker_web/lib/file_picker_web.dart
+++ b/file_picker_web/lib/file_picker_web.dart
@@ -1,8 +1,33 @@
 import 'dart:async';
-
 import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'dart:html' as html;
+
+
+/// File Object wrapper 
+/// 
+/// [file]  `html.File` object that contains picked file
+class File {
+  final html.File file;
+
+  File({this.file});
+  
+  String get name => file.name;
+  String get type => file.type;
+  int get size => file.size;
+  String toString() => file.toString();
+
+  /// Serialize `html.File` object
+  /// 
+  /// Returns a `<List<int>>`
+  Future<List<int>> fileAsBytes() async {
+    final Completer<List<int>> bytesFile = Completer<List<int>>();
+    final html.FileReader reader =  html.FileReader();
+    reader.onLoad.listen((event) => bytesFile.complete(reader.result) );
+    reader.readAsArrayBuffer(file);
+    return await bytesFile.future;
+  }
+}
 
 class FilePicker extends FilePickerPlatform {
   FilePicker._();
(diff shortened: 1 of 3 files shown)


miguelpruivo:
Ok thank you @lennhv I'll review it as soon as possible.

file_picker_web/lib/file_picker_web.dart
```diff
@@ -1,9 +1,34 @@
...
-
 import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'dart:html' as html;
 
+
+/// File Object wrapper 
+/// 
+/// [file]  `html.File` object that contains picked file
+class File {
```
miguelpruivo, line 10: I'm not sure if naming this to `File` would confuse some users as it's just wrapping a `html.File`. What advantages would you see with this?

lennhv:
Hi Miguel, answer your question: avoid expose `dart:html` library outside of plugin and prevent errors when develop  `cross-platform` applications. 

I agree with you, can be confuse but is the same case when expose `html.File`. I was happy developing an app until I found out that File object is not the same object type for mobile that in web. I think a better approach is expose the same object type for files regardless of whether it is mobile or web.

miguelpruivo:
But how would you use File from dart:io when on web? It's just a matter of being careful when picking your imports I'd say. The same goes for people importing dart:html when on mobile, which is invalid.

lennhv:
`File` from` dart: io` is not a valid object in the web environment, using` File` from` dart: html` is fine for web, but not optimal when the app is cross-platform, I think the best approach is to provide the same type of object, but it's a major change to the plugin, so I only implemented the wrapper for the web plugin. We can plan to make the plugin hide the type of object linked to the platform and expose a more generic one that allows us to work without worrying about the platform where the application is running.

miguelpruivo:
The thing is that you can't have html libs when running on mobile, neither you can't have io imports when running on web.

But since the package is already separated in two (web and io) there shouldn't be any issue, even without the wrapper.

I'm just trying to understand what is this actually trying to solve.

aliasgarlabs:
I ran into the same issue as @lennhv. I see @lennhv 's point. The issue the plugin should also expose a lower-level object something like Uint8List to allow abstract the platform during the processing the file once picked.

miguelpruivo:
@aliasgarlabs how can you come across that issue if you are using `file_picker_web` exclusively? The fact that the plugin is split in IO and Web serves exactly this purpose, to avoid inference conflicts.

My point is, if you are working with a web project you should never experience any issue. Are you both working on projects that could work on both Web and Mobile? If so, this might make sense and can actually fit with #234 where I’m going to create a PlatformFile object that abstracts this logic.

lennhv:
Hi @miguelpruivo, sorry for the late response, in my case i implementing a cross-platform application, mobile (android an IOS) and web.   

Basically i need upload files to a server, i know that it can be accomplished doing some validations based on platform type but i understand that major plugin advantage is provide a common interface without depending on extra validations in the app because the plugin do that work. Obviusly the approach that i implemented is not complet to accomplish that but is a good aproximation to that.

miguelpruivo:
Update is now live with [2.0.0](https://github.com/miguelpruivo/flutter_file_picker/blob/master/CHANGELOG.md#200) please, [see the changelog for details](https://github.com/miguelpruivo/flutter_file_picker/blob/master/CHANGELOG.md#200) of what have changed.

So this is no longer required as the introduced `PlatformFile` actually now does this and you have a unified plugin.

Thank you!

PR #350, opened 2020-08-04
Added File Picker Support for Android Q and above

lakshyab1995:
Opening a PR for File Picker Support for devices with Android Q and above.
Implemented OS level check to get file path for the selected file.
Verified this implementation on devices with Android Q and below as well, it is working as expected.
#234

### DIFF
--- a/file_picker/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
+++ b/file_picker/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
@@ -87,9 +87,11 @@ public class FilePickerDelegate implements PluginRegistry.ActivityResultListener
                             final ArrayList<String> paths = new ArrayList<>();
                             while (currentItem < count) {
                                 final Uri currentUri = data.getClipData().getItemAt(currentItem).getUri();
-                                String path = FileUtils.getPath(currentUri, FilePickerDelegate.this.activity);
-                                if (path == null) {
+                                String path;
+                                if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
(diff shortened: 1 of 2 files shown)


lakshyab1995:
Hi @miguelpruivo , 
It wasn't working before in Android Q when I opt in for scoped storage. It was showing cannot open file-permission denied (FileSystemException) when I select any file from file_picker.
In this PR, I made the change that if user is targeting Android Q and the scoped storage is enabled then this will create a file from uri and save it in app cache dir so the file will be accessible within the app.

miguelpruivo:
@lakshyab1995 were you using `requestLegacyStorage: true` on your manifest file?

lakshyab1995:
No, I wasn't using `requestLegacyStorage ` flag in manifest file

miguelpruivo:
@lakshyab1995 that's the reason why. Any reason why you weren't using it? I've a proposal for it on #234 that will switch to it (but requires some major refactor that I hadn't the time yet).

With your PR, what will happen if users have `requestLegacyStorage: true` set on Android Q?

lakshyab1995:
@miguelpruivo, I was thinking of an alternative way of not using this flag in manifest until the plugin is moved to scoped storage.
It will also work for users who have set `requestLegacyStorage: true` flag in manifest.

miguelpruivo:
@lakshyab1995 have you tested it out? Currently I don't have time to make a deep test on it, but I will merge your PR if you ensure that it will work for both legacy and non legacy users.

lakshyab1995:
@miguelpruivo Yes, I have tested it out and it is working for both legacy and non legacy users.

miguelpruivo:
@lakshyab1995 I will deploy a new version later today or tomorrow with this and other bug fixes on pub.dev. Thanks!

lakshyab1995:
@miguelpruivo Thanks for merging the PR.

PR #352, opened 2020-08-05
Urgent Bug Fix for null path for file selection in Downloads folder for OS below Android Q

lakshyab1995:
@miguelpruivo , please merge this PR which resolves the issue when the user selects a file from **Downloads** folder and the content uri is not able to resolve to a file path so in turn the file path is returned as null. (happens on devices below Android Q)

This PR resolves this issue, when the file path is null then it will create a file from uri and save it in app cache dir so the file will be accessible within the app.

Thanks!


miguelpruivo:
@lakshyab1995 that's why I'm always afraid to fully merge without test it myself and asked you yesterday about it 👀

No problem, glad you found it quickly.

PR #386, opened 2020-09-12
Fix bug for directory picking on Linux: drop file type filter

philenius:
This pull request fixes a bug which I came across on Linux (this bug existed already before the new major release 2.0.0).

Previously, the file filter argument was set to `*.*` which somehow led to the effect that all directories are displayed as not clickable / selectable in the directory picker dialog on Linux (Ubuntu 19). This effect can be observed on the screenshot below. All directories are greyed-out and cannot be selected (except the bookmarks / default directories on the left side):

previously

As stated in the docs of [gen2brain/dlgs](https://github.com/gen2brain/dlgs/blob/master/file_linux.go), setting the file filter to a blank string will display all file types:

```go
// File displays a file dialog, returning the selected file or directory, a bool for success, and an
// error if it was unable to display the dialog. Filter is a string that determines 
// which extensions should be displayed for the dialog. Separate multiple file 
// extensions by spaces and use "*.extension" format for cross-platform compatibility, e.g. "*.png *.jpg".
// A blank string for the filter will display all file types.
func File(title, filter string, directory bool) (string, bool, error) {
   ...
}
```

I tested it on Ubuntu 19 and it works. Now, the directory picker dialog allows to select / click on directories. Also, the filter on the bottom right vanished:
updated

### DIFF
--- a/go/file_linux.go
+++ b/go/file_linux.go
@@ -39,7 +39,7 @@ func fileDialog(title string, filter string) (string, error) {
 }
 
 func dirDialog(title string) (string, error) {
-	dirPath, _, err := dlgs.File(title, `*.*`, true)
+	dirPath, _, err := dlgs.File(title, "", true)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to open dialog picker")
 	}


miguelpruivo:
Thank you @philenius!

philenius:
I looks like this commit got lost / overriden? Github says _This commit does not belong to any branch on this repository_.
Also the latest releases inlcuding 2.0.3 on [pub.dev](https://pub.dev/packages/file_picker/versions) do not come with my little bug fix. Could you please merge this again?

miguelpruivo:
@philenius weird that it got lost in the way. 😞 Sure, I will do it right now.

philenius:
Awesome, thank you for your :rocket: fast response.

miguelpruivo:
@philenius done and published with 2.0.4. 🚀

PR #432, opened 2020-10-06
Directory access is not available on iOS 12 or lower

allanwolski:
The getDirectoryPath() doesn't work when access to the directory is not available.
This PR changes to use NSDocumentDirectory on iOS 12 or lower.

https://developer.apple.com/documentation/uikit/view_controllers/providing_access_to_directories


miguelpruivo:
Thank you for contributing and spot this @allanwolski!

Issue #425, opened 2020-10-05
Can't make getDirectoryPath work in Android 8.1.


me21:
FilePicker version is 2.0.6. `flutter clean` and `flutter build` didn't help.

**Describe the bug**
I can't select a directory on a phone with Android 8.1, or on emulator with Android 8.1. The same code works with Android 11.

My `compileSdkVersion` is 28, `minSdkVersion` is 21 and `targetSdkVersion` is 28.

**Issue details**
1. Android
2. 8.1
3. I'm trying to select a directory
4. I call `final String fromPath = await FilePicker.platform.getDirectoryPath();`, but `fromPath` is null. No matter which directory I choose (I tried root directory, Downloads, created a `test` directory in the root, it doesn't matter).

I've read Troubleshooting page, it says there are directories on Android which can't be picked manually, but what directory should I choose then, even if my freshly created directory doesn't work?

**Error Log**
```
Launching lib\main.dart on BV5800 in debug mode...
Running Gradle task 'assembleDebug'...
√ Built build\app\outputs\flutter-apk\app-debug.apk.
Waiting for BV5800 to report its views...
Debug service listening on ws://127.0.0.1:59288/wBV95REH17s=/ws
Syncing files to device BV5800...
D/Surface ( 8429): Surface::disconnect(this=0x7635ff6000,api=1)
D/Surface ( 8429): Surface::disconnect(this=0x7635ff6000,api=-1)
D/Surface ( 8429): Surface::disconnect(this=0x7635ff5000,api=1)
V/PhoneWindow( 8429): DecorView setVisiblity: visibility = 4, Parent = ViewRoot{5a31781 com.example.hello_world/com.example.hello_world.MainActivity,ident = 0}, this = DecorView@68a9a8b[MainActivity]
I/zygote64( 8429): Do partial code cache collection, code=26KB, data=25KB
I/zygote64( 8429): After code cache collection, code=26KB, data=25KB
I/zygote64( 8429): Increasing code cache capacity to 128KB
D/FilePickerDelegate( 8429): [SingleFilePick] File URI:content://com.android.externalstorage.documents/tree/primary%3Atest/document/primary%3Atest
I/FilePickerUtils( 8429): Caching from URI: content://com.android.externalstorage.documents/tree/primary%3Atest/document/primary%3Atest
E/FilePickerUtils( 8429): Failed to retrieve path: read failed: EISDIR (Is a directory)
[...]

PR #433, opened 2020-10-07
Fixes directory pick on Android 9 or lower (#425)

allanwolski:
This PR fixes the problem with directory pick on Android 9 or lower


miguelpruivo:
Closes #425.

PR #448, opened 2020-10-17
feat: pass multiple types of events from channel

ncb000gt:
Need to use event channel to communicate between native code and dart
but existing implementation expected only a boolean value to be passed.

This creates a file event which translates to a platform event. The
setup allows for future work related to passing more events. This commit
includes the addition of the source paths as soon as they are available
as they may be usable more immediately than the copied files are in the
event of large files.

Signed-off-by: Nick Campbell <nicholas.j.campbell@gmail.com>

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -28,12 +28,19 @@ android {
         minSdkVersion 16
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
+
     lintOptions {
         disable 'InvalidPackage'
     }
-      dependencies {
+
+    dependencies {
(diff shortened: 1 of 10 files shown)


ncb000gt:
For some context, I've used this to pick files to be used in a video player. Some of the videos are large and needing to download them before showing takes a while. When possible, I want to show the videos sooner should it be possible to buffer those files in the video player. So, I used the "FileEvent" to enable this.

miguelpruivo (changes requested):
Could you please elaborate an example on how you'd use this and can you please rebase it to latest version? Thank you!

PR #453, opened 2020-10-21
add  useful comment in "allowMultiple"

j-j-gajjar:


miguelpruivo:
Hi my friend, I don’t believe that this is necessary as the parameters are verbose enough. Plus, this relates to the example app.

Thank you anyway for your contribution.

PR #488, opened 2020-11-09
Add the capability of retrieving 'lost' files after activity being de…

biessek:
Based on [image_picker](https://pub.dev/packages/image_picker) plugin, I've added the same logic to retrieve 'lost' files selection after activity destruction on Android, This can occur by different reasons and I've seen it happening a lot on huawei and xiaomi devices.

After calling `pickFiles` if the app `Activity` is destroyed, the user can retrieve the "lost selection" by calling `retrieveLostData'. I'm using this functionality in some apps and maybe it can be useful to others as well.

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerCache.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerCache.java
@@ -1,0 +1,118 @@
+package com.mr.flutter.plugin.filepicker;
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.net.Uri;
+import android.os.Build;
+
+import androidx.annotation.Nullable;
+
+import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


miguelpruivo:
@biessek hi, thank you for your PR. Although this looks like a nice idea at first, I'd rather to not have another object on the Dart side and let this logic being purely handled on the Android side (save/restore instance). 

Do you think that it can be done?

Thank you!

biessek:
I understand you @miguelpruivo , I also think that it is not the ideal way, but from my experience I don't think there is any other way for now. Since when the activity is destroyed by Android OS, the "link" between native / flutter made by `MethodCall` is lost, so this new link is established with` retrieveLostData`

miguelpruivo (changes requested):
Can you please rebase it to latest version? Thank you!

biessek:
Done @miguelpruivo

chowdhuryrahulc (approved):

Issue #491, opened 2020-11-10
How can I get extension on WEB?


michaldev:
I try like this:
```
  var result = await FilePicker.platform.pickFiles(type: FileType.custom, allowedExtensions: ['jpg', 'pdf'], allowMultiple: false);
  print(result.files.single.extension);
```

But this work only on mobile devices.

miguelpruivo:
@michaldev it should work on web as well. Which browser are you trying on?

michaldev:
Chrome. I got null.

miguelpruivo:
It shouldn't actually. Have you tried with a different one?

sultanmyrza:
have same issue path is null on flutter web
I created more detailed issue for others to follow 
https://github.com/miguelpruivo/flutter_file_picker/issues/495

miguelpruivo:
@sultanmyrza #495 is closed since it’s working as intended. You must use `bytes` property on the web since absolute paths aren’t supported by browsers.

sultanmyrza:
@miguelpruivo  I'm trying to do something like this but can't make it work on web
```
    FilePickerResult result = await FilePicker.platform.pickFiles();

    if (result != null) {
      var bytes = result.files.single.bytes;
      var fileName = result.files.single.name;

      // dart:html 
      var file = File(bytes, fileName);
      // dart: io
      var file = await File(fileName).writeAsBytes(bytes);

      // TODO: upload to firebase storage
      UploadTask task = uploadFile(file, fileName);
```

miguelpruivo:
Closing as this should be fixed on the latest version (2.0.13).

Feel free to reopen if you still experience any problem related to it.

Thank you!

PR #497, opened 2020-11-11
Fix null PlatformFIle.extension error when called on Web

markgrancapal:
Multiple people encountered an error where calling PlatformFile.extension returns a null, like issue #491 

I've created a simple fix for this one. Hope it will help other struggling to create a fix for this.


miguelpruivo:
@markgrancapal good solution. It should remove the struggle since on the web the path is always null, hence the extension won't be available. Thank you. 👍 

Merging.

PR #542, opened 2020-12-10
Fixed desktop plugin implementation

DenchikBY:


lib/src/file_picker_io.dart
```diff
@@ -8,8 +8,13 @@ import 'package:flutter/services.dart';
 
 import 'file_picker_result.dart';
 
-const MethodChannel _channel =
-    MethodChannel('miguelruivo.flutter.plugins.filepicker');
+final MethodChannel _channel = MethodChannel(
```
miguelpruivo, line 11: Why can’t it be `const`?

miguelpruivo (changes requested):
Thank you for the PR. Please, check the comments.

lib/src/file_picker_io.dart
DenchikBY: Because there is ternaty

miguelpruivo: Oh yes, never mind. I’m reviewing on my phone. 😄

kuhnroyal:
As far as I can tell, this broke mocking of the method channel. It is no longer const and it is private which means I can no longer access the same const instance just with the name to set the mock method handler.

kuhnroyal:
Probably better to create 2 const channels and switch them depending on the platform.

PR #638, opened 2021-03-08
[Android][Bug fix]use long instead of int to represent no. of bytes  when parsing file sizes

hashem78:
On Android file_picker would crash if the number of bytes for a file exceeds the JAVA integer limit (2^32 bytes) so with this PR the limit should be at 2^64 bytes or the JAVA long limit.

Minimum steps to reproduce

```dart
Future<void> pick() async {
...
await FilePicker.platform.pickFiles();
...
}
```
Then once the picker launches pick a file with more than 2.1 GB, file_picker will crash with java.lang.NumberFormatException

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
@@ -173,7 +173,7 @@ public class FileUtils {
         fileInfo
                 .withPath(path)
                 .withName(fileName)
-                .withSize(Integer.parseInt(String.valueOf(file.length())));
+                .withSize(Long.parseLong(String.valueOf(file.length())));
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
```diff
@@ -173,7 +173,7 @@ public static FileInfo openFileStream(final Context context, final Uri uri, bool
         fileInfo
                 .withPath(path)
                 .withName(fileName)
-                .withSize(Integer.parseInt(String.valueOf(file.length())));
```
hashem78, line 176: the crash would happen here

miguelpruivo (approved):
Thank you.

Issue #661, opened 2021-03-26
Pick from Gallery for Images (Android)


tmthecoder:
When utilizing the `.image` filetype on FilePicker, I'm opened to the File Explorer window and am wondering whether there is any way to utilize the Android device's default gallery as the picker defaults to the user's documents directory in my case when any file picker action is started. If this is possible, which I think should be using the MediaStore API, I could write a PR if needed. iOS already does this, but android doesn't, leaving me with where I picked in a custom file on another filepicker call

miguelpruivo:
If this is possible on both iOS and Android, feel free to purpose an API and I’d be glad to implement it. Seems useful if possible.

tmthecoder:
Just submitted a PR outlining the above

miguelpruivo:
Closing as this is already implemented.

Thank you.

PR #663, opened 2021-03-27
Utilize a Gallery-Based picker for Android for Images

tmthecoder:
References #661 where we can use the Gallery picker instead of the file manager to pick images on Android, which makes the process to pick an image more streamlined as a user won't have to navigate through files with a similar gallery presented through the MediaStore API.

I've tested the changes and made sure normal picking as well as image picking, both in single and multiple items work.

Let me know if anything needs to be changed!


miguelpruivo:
Hi, is this available with all Android versions?

tmthecoder:
Yes, as far as I know, it's available with all android versions.

miguelpruivo:
Ok, thank you.

Sulisong:
> 好的，谢谢你。

调用系统后，部分系统自带的相册无法多选。希望添加一个开关，可以选择多选模式下用文档

PR #670, opened 2021-03-30
Detect when cancel is clicked on web

letranloc:


ttpho (approved):
LGTM 💯 

@letranloc Thank you for the awesome PR.

@miguelpruivo @GeertJohan 
Please review merge this PR. (bow)

lib/src/file_picker_web.dart
```diff
@@ -108,15 +108,34 @@ class FilePickerWeb extends FilePicker {
       });
     }
 
+    void cancelledEventListener(_) {
+      window.removeEventListener('focus', cancelledEventListener);
+
+      // This listener is called before the input changed event,
+      // and the `uploadInput.files` value is still null
+      // Wait for results from js to dart
+      Future.delayed(Duration(milliseconds: 500)).then((value) {
```
miguelpruivo, line 117: Can we use a microtask here instead? (`scheduleMicrotask(callback)`)?

miguelpruivo (commented):
I see what you did here. It's sort of a workaround I'd say, but it should work (in theory, I didn't have the chance to test it out yet).

Please, check the comments above.

lib/src/file_picker_web.dart
letranloc: Thank you

Tested with `scheduleMicrotask(callback)` and it doesn't work: the `callback` is called too soon

miguelpruivo:
Thank you.

PR #742, opened 2021-06-11
Allow not load files into memory but get file by Stream.

GrapevineLin:
Sometimes we dont need to load bytes into memory,for example there is large files to upload，the file might be too big to fit device memory.
So we just need a stream,and then through this stream we can just read byte[size] on demand.

I add a boolean argument [cachedFile] to [pickFiles],If [cachedFile] is set, it will not load bytes into memory, you can use [getBytesByUri] to get the Bytes
 when you need.

And add a menber [uri] in FileInfo so that we can use this uri to run [getBytesByUri], everytime we run [pickFiles] we will get a uri in FileInfo.

The [getBytesByUri] function allows you get bytes by uri which from [pickFiles]'s uri.

There is also [closeFileInputStreamByUri] and [openInputStreamByUri] to operate the stream by uri.

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FileInfo.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FileInfo.java
@@ -7,12 +7,14 @@ import java.util.HashMap;
 public class FileInfo {
 
     final String path;
+    final Uri uri;
     final String name;
     final long size;
     final byte[] bytes;
 
-    public FileInfo(String path, String name, long size, byte[] bytes) {
+    public FileInfo(String path, Uri uri, String name, long size, byte[] bytes) {
         this.path = path;
+        this.uri = uri;
         this.name = name;
         this.size = size;
         this.bytes = bytes;
@@ -22,31 +24,37 @@ public class FileInfo {
 
         private String path;
         private String name;
+        private Uri uri;
         private long size;
         private byte[] bytes;
 
-        public Builder withPath(String path){
+        public Builder withPath(String path) {
             this.path = path;
             return this;
         }
 
-        public Builder withName(String name){
+        public Builder withUri(Uri uri) {
+            this.uri = uri;
+            return this;
+        }
+
+        public Builder withName(String name) {
             this.name = name;
             return this;
         }
 
-        public Builder withSize(long size){
+        public Builder withSize(long size) {
             this.size = size;
             return this;
         }
 
-        public Builder withData(byte[] bytes){
+        public Builder withData(byte[] bytes) {
             this.bytes = bytes;
             return this;
         }
 
         public FileInfo build() {
-            return new FileInfo(this.path, this.name, this.size, this.bytes);
+            return new FileInfo(this.path, uri, this.name, this.size, this.bytes);
         }
     }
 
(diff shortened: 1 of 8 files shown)


GrapevineLin:
sorry,its automatically format

GrapevineLin:
@miguelpruivo hey, can you check this . I think its very useful

miguelpruivo:
Hi, I’ve been on vacations for a while. I’ll take a look next week when I get back to work.

miguelpruivo (changes requested):
Can you please rebase it to latest version? Thank you!

miguelpruivo (changes requested):
Now that the URI is available and exposed in the `PlatformFile` do you mind rebasing and updating this PR as well? Also, please run `flutter format .` in the end so it follows the guidelines and doesn't mark a lot of files as changed when they didn't — this make it harder to review. Thank you.

GrapevineLin:
> Now that the URI is available and exposed in the `PlatformFile` do you mind rebasing and updating this PR as well? Also, please run `flutter format .` in the end so it follows the guidelines and doesn't mark a lot of files as changed when they didn't — this make it harder to review. Thank you.

I'm glad  to do this, but I'm to busy lately.
I will  update this PR as soon as I'm free. thank you for reply :)

shareef-dweikat:
Hey @GrapevineLin [GrapevineLin](https://github.com/GrapevineLin)

Please check the PR when you can. I am facing a an issue with picking large files

shareef-dweikat:
@GrapevineLin 
Hi, can you rebase and merge? I tried doing it my self. I did not succeed. I think you understand the code better than me.
I am working on a project. This feature is a must to have in our Alpha release.

shareef-dweikat:
Hi @miguelpruivo ,
After reading second question/answer in FQA section https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ and some of your comments, i understand that cashing a file is a must, because accessing absolute paths is forbidden starting from Android 10 (SDK 30) and above. But at the same time i see you like this PR which trying to make selecting a big file possible by not cashing it, so please can you explain it for me? is this PR going to solve the issue so we don't need to cash files anymore? or not?

miguelpruivo:
> Hi @miguelpruivo ,
> After reading second question/answer in FQA section https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ and some of your comments, i understand that cashing a file is a must, because accessing absolute paths is forbidden starting from Android 10 (SDK 30) and above. But at the same time i see you like this PR which trying to make selecting a big file possible by not cashing it, so please can you explain it for me? is this PR going to solve the issue so we don't need to cash files anymore? or not?

I wasn't aware of that since I didn't look into this for a while (the PR is almost a year old). I'm afraid it won't work if no caching is done or the file isn't read in chunks into a stream.

Maybe @GrapevineLin may explain what is being purposed in this PR as a solution and that will help both of us to understand it more in detail.

GrapevineLin:
@miguelpruivo @shareef-dweikat  I'm sorry I didn't notice the message here.
I haven't worked on flutter in a while, but I will check this PR soon, resubmit a new PR as soon as possible

shareef-dweikat:
@GrapevineLin  No worries. Thanks a lot.

GrapevineLin:
no time to continue this pr anymore.
hope someone could do it property.
sorry

PR #780, opened 2021-07-28
Check unsupported extensions correctly.

Jahn08:
fileType cannot be "custom" since it always contains file types matched with a passed method name. Thus, the condition is always false.

When the error never happens it's likely to lead to a weird behaviour. For instance, I've stumbled upon a case when the file explorer is opened successfully but there is no possibility to choose any folders except for some default ones (e.g., Recent, Photos or Drive). The problem can be reproduced on Android API < 29 with an unsupported file extension 'json' (it's been included starting from API 29):
Android 8 0-9 0_Custom

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
@@ -165,7 +165,7 @@ public class FilePickerPlugin implements MethodChannel.MethodCallHandler, Flutte
             allowedExtensions = FileUtils.getMimeTypes((ArrayList<String>) arguments.get("allowedExtensions"));
         }
 
-        if (fileType == "custom" && (allowedExtensions == null || allowedExtensions.length == 0)) {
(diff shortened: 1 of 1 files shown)


miguelpruivo:
Thank you. What's the difference that you're experiencing with this or as it is with the `fileType = FilePickerPlugin.resolveType(call.method)`?

Jahn08:
The difference is that `fileType == "*\*"` and never `"custom"`. When I get the intended error at least I know what went wrong and don't see the file explorer, otherwise one might spend some time figuring out why the file explorer doesn't look properly - akin to this normal view:
Android 8 0_Default

miguelpruivo:
Does that happen with which versions?

Jahn08:
I've run into that weirdly looking file explorer from my initial post in emulators for API 26-28. Just because the JSON file extension is not supported in those versions (and I bet in the lower ones either) and I should've got the error and seen no explorer at all - that's what would happen now with the changes I've brought here.

miguelpruivo:
Ok got it. I'll merge it.

PR #785, opened 2021-08-01
fix #782

### LINKED ISSUE #782: Custom File Type CSV not support in Android
**Platform**
- [] Android
image

mmdrez4:
fix issue and support Type CSV in Android (#782 )


miguelpruivo:
Hi, this is not actually fixing anything, just setting the `csv` as a default filter.

PR #791, opened 2021-08-04
Migrate jcenter pointers to mavenCentral

ened:
Simple PR to reduce a few warnings Android release builds will emit when compiling against this package.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,7 +4,7 @@ version '1.0-SNAPSHOT'
 buildscript {
     repositories {
         google()
-        jcenter()
+        mavenCentral()
     }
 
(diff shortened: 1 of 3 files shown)


miguelpruivo:
Hi. Which warnings does this fix?

ened:
Background is this https://stackoverflow.com/questions/66651640/jcenter-deprecation-impact-on-gradle-and-android & https://jeroenmols.com/blog/2021/02/04/migratingjcenter/

Also Android Studio would warn you:
Screenshot 2021-08-10 at 11 46 16

The Gradle warnings will sometimes appear during release builds, I can not find them currently.

PR #811, opened 2021-08-29
Add custom equality and toString

Nolence:
Hi ! I was recently trying to compare platformFiles but realized they weren't really comparing that well.

This is a small PR that adds `==`, `hashCode` and `toString` to `FilePickerResult` and `PlatformFile`.


miguelpruivo (approved):
Looks good to me. Thank you!

PR #838, opened 2021-09-27
Bug: incorrect file filter on Linux

### LINKED ISSUE #836: [Linux] Unable to pick files without extension
Calling `pickFiles` with `FileType.any` will automatically set the extension filter to `*.*` which effectively filters out files without extension.

I tried getting around it by using `FileType.custom` and `allowedExtensions: ['']` but that results in a `*.` filter which also ignores files without extension (as they're missing the trailing dot).

I only tested this on Linux, so I don't know if the issue also exists on other platforms.

Tested with file_picker 4.1.1

philenius:
Fixes #836


miguelpruivo (approved):
Thank you @philenius! 🎉

PR #869, opened 2021-11-11
Check FlutterResult _result, is not void when picking media in PHPick…

eduhcano:
Sometimes when the users select videos very fast, _didFinishPicking:_ is called twice, and in the second time _result is null and the app crashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.1.3
+
+##### iOS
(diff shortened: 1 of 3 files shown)


miguelpruivo (changes requested):
Thank you, could you please increment the build version in the pubspec and also update the changelog accordingly? Thank you.

eduhcano:
Sure! build version incremented and changelog updated.

miguelpruivo (approved):

PR #911, opened 2021-12-06
Android use RealPathUtil to get file name && copy file to cache every time.

### LINKED ISSUE #909: Android different files with same name && some file return id instead of file name
**Platform**

- [✓ ] Android

**Describe the bug**

There are 2 bug in android.
1. The first bug is, when there are 2 files in different directories with same name, the method in [FileUtils.getFileName](https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java#L58) will return same name. 
And then the method in [FileUtils.openFileStream Line 133](https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java#L133) will create file with same path in cache directory. Before copying content from real path, [it checks if the cache file exists](https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java#L135). If yes, it won't copy content.
So bang! If you select same name file last time, and the cache directory hasn't been cleared,  this time you get the wrong file. 

<strong><em>Maybe we can copy every time to solve the bug? </em></strong>

2. Some android devices in some directory, it will return the uri like `content://media/external/file/96668` instead of real path. So [getFileName](https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java#L58) returns `96668` instead of correct file name, even though the content in the cache directory is right.
I found the gist [RealPathUtil by #tatocaster](https://gist.github.com/tatocaster/32aad15f6e0c50311626). It returns file's real path in different android version. But some files have [`java.lang.IllegalArgumentException: column '_data' does not exist`](https://gist.github.com/tatocaster/32aad15f6e0c50311626#gistcomment-2760298) problem.

<strong><em>We can use RealPathUtil to get real path, if it causes error, the we will use the plugin current getFileName method to solve the bug</em></strong>.

FaKenKoala:
fix #909

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
@@ -60,8 +60,24 @@ public class FileUtils {
 
         try {
 
+            String realPath = null;
+            try {
+                realPath = RealPathUtil.getRealPath(context, uri);
+                if (realPath != null) {
+                    int cut = realPath.lastIndexOf('/');
+                    if (cut != -1) {
+                        result = realPath.substring(cut + 1);
+                        /// get file real name
+                        return result;
+                    }
+                }
+            } catch (Exception exception) {
+                Log.e(TAG, "Failed to get real path: " + exception.getMessage());
+            }
+
             if (uri.getScheme().equals("content")) {
-                Cursor cursor = context.getContentResolver().query(uri, new String[]{OpenableColumns.DISPLAY_NAME}, null, null, null);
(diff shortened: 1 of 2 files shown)


aCoolWater:
This solution solved my problem

miguelpruivo:
Thank you @FaKenKoala would you mind rebasing and also bump the patch version and changelog?

Issue #630, opened 2021-03-04
Integration with Flutter's file_selector for native desktop support


marchellodev:
Hey, is it possible to integrate this package with [file_selector](https://pub.dev/packages/file_selector) ([file_selector_windows](https://pub.dev/packages/file_selector_windows), [file_selector_linux](https://pub.dev/packages/file_selector_linux), [file_selector_macos](https://pub.dev/packages/file_selector_macos)), so that it will work with desktop operating systems in a native way?

Thanks!

miguelpruivo:
Could be interesting. @stuartmorgan what do you think?

stuartmorgan:
I don't see why it wouldn't possible. There are a lot of different things that could look like, depending on your goal:
1. Make a new meta-package using both (see `file_picker_cross`, which AFAICT is exactly that, but with `file_chooser`, whose code became the `file_selector` implementations).
2. Make your plugin depend on `file_selector` or `file_selector_platform_interface` and use it to implement desktop support as part of this package.
3. Make an adapter package that implements `file_selector_platform_interface`'s API using this plugin, which developers could use an an unendorsed implementation to add mobile support to `file_selector` in their projects.
4. Contribute this plugin's mobile implementations to flutter/plugins as endorsed mobile implementations, and maintain them there going forward, discontinuing this plugin.

1 or 3 could be done by anyone, 2 or 4 depend entirely on your interest in either of those approaches. (If you federated this plugin, the reverse of option 3 would also become possible.)

miguelpruivo:
@stuartmorgan thank you for the input.

Previously I used the interface approach to make it more modular and platform independent, however, since most of the API would require an update anyway when a new feature was added, I found it quite hard to maintain alone and some users wanted to use the app on both web + IO which would result in having to import the 2 packages and sometimes conflict with one on another.

Right now, as it is, it's cross compatible with all platforms (desktop being used only by flutter go) but it would be great to add native desktop support as well. I'll have to think about it.

stuartmorgan:
> and some users wanted to use the app on both web + IO which would result in having to import the 2 packages and sometimes conflict with one on another.

FWIW that just sounds like an implementation bug in one or both of the packages; there's no reason web and non-web implementations of a plugin can't coexist, regardless of federation. (Not too suggest that you should use the federated structure if you don't want to, just that that shouldn't be a reason not to use it.)

> Right now, as it is, it's cross compatible with all platforms (desktop being used only by flutter go)

[...]

PR #925, opened 2021-12-16
Added option to lock the parent window(flutter window) when file picker is open

vinicios-cervantes:
- Added implementation lock the main flutter window  behind the file picker window when `lockParentWindow` is set to `true`
 
without lock(default, how it is today):
without_parent_handler

with lock(when set):
with_parent_handler

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.3.0
+
+##### Desktop (Windows)
+Adds the parameter `lockParentWindow` for Windows desktop. This parameter makes the file picker dialog behave like a modal window. That is, the file picker always stays in front of the Flutter window until it is closed. Thank you @vinicios-cervantes.
+
 ## 4.2.8
 
 ##### Desktop (Windows)
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -10,6 +10,8 @@ analyzer:
 #   strong-mode:
 #     implicit-casts: false
 
-# linter:
-#   rules:
-#     - cancel_subscriptions
+linter:
+  rules:
+    # - cancel_subscriptions
+    - comment_references
+    - slash_for_doc_comments
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -44,6 +44,7 @@ class FilePickerWeb extends FilePicker {
     bool allowCompression = true,
     bool withData = true,
     bool withReadStream = false,
+    bool lockParentWindow = false,
   }) async {
     if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
       throw Exception(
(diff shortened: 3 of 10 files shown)


miguelpruivo (commented):
Hi! Thank you for this PR, it looks handy! Do you know if MacOS and Linux also supports this? If they do, it would be great to have it implemented on those as well as part of this PR.

cc: @philenius

vinicios-cervantes:
Looks like on linux, you can put zenity/qarma always on top using some bash commands, producing a similar behavior, on MacOS i dont know.

philenius:
@miguelpruivo , I can confirm that @vinicios-cervantes' implementation works on Windows. Great job! :rocket:

I have to disagree with @vinicios-cervantes on the Linux and macOS implementation. Zenity and qarma themselves have no option to keep the dialog on top of the Flutter app. The same applies to osascript / AppleScript that we use for our macOS implementation. The issue is that we use a script-based approach on Linux and macOS which comes with unsolvable (UX) limitations. Stuart Morgan has pointed out these limitation in https://github.com/miguelpruivo/flutter_file_picker/issues/630#issuecomment-896063891

Although, there exist workarounds where you have to chain shell commands and thereby can keep zenity in foreground, but this sounds completely wrong to me:
https://unix.stackexchange.com/questions/152294/keep-a-zenity-dialog-box-always-on-top-in-foreground

If we wanted to have a `lockParentWindow` option for Linux and macOS, then we'd have to switch to an implementation in C++ as Google did: https://github.com/google/flutter-desktop-embedding/tree/master/plugins/file_selector

Our Windows implementation is different as it directly interacts with Windows APIs instead of executing a shell command.

@miguelpruivo, do you still want to merge this PR? I think that the Windows implementation is a valuable contribution. I could create a feature branch, merge @vinicios-cervantes changes into this feature branch, fix the linter errors, bump the package version, and update the doc comments.

miguelpruivo:
@philenius thank you for checking on this. Yes, I’ll merge it right away. Could you please check why the CI is failing on the format stage?

miguelpruivo (approved):

philenius:
@miguelpruivo , thanks for merging and sorry for not describing my changes / the movitation behind my changes:

I activated two more linter rules. The rule `comment_references` basically states that parameters in doc comments should be wrapped in square brackets:

grafik

https://dart-lang.github.io/linter/lints/comment_references.html

https://dart.dev/guides/language/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions

That's why I replaced all the backticks in our doc comments, e.g.:
```
If `withReadStream` is set, picked files will have its byte data available as a `Stream<List<int>>`
```
became:
```
If [withReadStream] is set, picked files will have its byte data available as a [Stream<List<int>>]
```

Somehow, values such as `false`, `true`, or `null` must not be wrapped in square brackets...


As a result, Dartdoc will look up the name of parameters/types and link to its docs as can be seen in our official API documentation:

grafik

https://pub.dev/documentation/file_picker/latest/file_picker/FilePicker/pickFiles.html


Thank you for merging. :rocket:

PR #940, opened 2022-01-16
Feature/validate filename for savefile dialog

### LINKED ISSUE #926: No error or dialog if filename has an incorrect character
I was trying to run the following code on Windows. 
```dart
final String? outputFile = await FilePicker.platform.saveFile(
    dialogTitle: 'Save as:',
    fileName: "Exported: $deviceName.json"
    allowedExtensions: ["json"],
    type: FileType.custom,
  );
```
The dialog did not open and it only returned null. I figured out that it was because i was using ":" in the filename. This might be an invalid character and might be by design. Either way, I believe it should throw an error, to easily figure out the problem.

philenius:
Fixes #926.

Under Windows the save-file dialog did not open if the specified file name contained an illegal character. Windows prohibits the usage of [reserved characters in file names](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). Now the exception `IllegalCharacterInFileNameException` is thrown if the specified file name contains forbidden characters.

I also activated many more linter rules, but I had to disable [avoid_print](https://dart-lang.github.io/linter/lints/avoid_print.html) because `file_picker_io.dart` contains print statements.

@miguelpruivo are you okay with my approach of throwing this custom exception? I'm open to suggestions for improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 4.3.2
+
+#### Desktop (Windows)
+Fixes the issue under Windows that the save-file dialog did not open if the specified file name contained an illegal character. Windows prohibits the usage of [reserved characters in file names](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). Now the exception `IllegalCharacterInFileNameException` is thrown if the specified file name contains forbidden characters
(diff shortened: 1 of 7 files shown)


miguelpruivo (approved):
LGTM. There's no problem in throwing an `IllegalCharacterInFileNameException` since it should be handled in the scope of all Exceptions anyway.

Issue #970, opened 2022-02-20
initialDirectory not working


julianklose:
**Describe the bug**
I'm using the newest version (4.4.0) of the file_picker including the new `initialDirectory` option. I tried using it with the directory picker. Unfortunately, it doesn't seem to work, I always start at my home directory.

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop

**Platform OS version**
Windows 10

**How are you picking?**
```
var dir = await FilePicker.platform.getDirectoryPath(
        lockParentWindow: true, initialDirectory: Directory.current.path);
```

**Flutter Version details**
[√] Flutter (Channel stable, 2.10.1, on Microsoft Windows [Version 10.0.19044.1526], locale de-DE)
    • Flutter version 2.10.1 at C:\Users\JulianKlose\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision db747aa133 (11 days ago), 2022-02-09 13:57:35 -0600
    • Engine revision ab46186b24
    • Dart version 2.16.1
    • DevTools version 2.9.2

[√] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    • Android SDK at C:\Users\JulianKlose\AppData\Local\Android\sdk
    • Platform android-32, build-tools 32.1.0-rc1
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.11+9-b60-7590822)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop for Windows (Visual Studio Build Tools 2019 16.11.9)
    • Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools
    • Visual Studio Build Tools 2019 version 16.11.32106.194
    • Windows 10 SDK version 10.0.19041.0

[√] Android Studio (version 2021.1)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.11+9-b60-7590822)

[√] VS Code (version 1.64.2)
[...]

Issue #943, opened 2022-01-17
getDirectoryPath() doesn't work on the web


feinstein:
**Describe the bug**
getDirectoryPath() doesn't work on the web, I get an exception saying it was implemented 

**Platform**

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Desktop

**Platform OS version**
Chrome and Edge on Windows 10

philenius:
@feinstein that’s not an issue. It's simply not supported on web, as pointed out in our [wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki/api#-getdirectorypath). Duplicate of #675. Feel free to provide an implementation if you know how to do implement it :wink: 

Sorry for any inconvenience.

feinstein:
I think this should be better documented on the README then. Most people won't open the wiki.

miguelpruivo:
@feinstein the readme points to the Wiki, which as per good practices, is the right place for the docs. 😀

feinstein:
Yes, but for simple packages, where people are only going to copy paste 1 function, it's natural to only read the README.

Also, the README should at least point out that not all features have full compatibility. It gives an impression that all features are present, when in fact they aren't.

miguelpruivo:
@feinstein sure, it’s important to note and be aware that some features are not implemented because some platforms doesn’t completely support it.

feinstein:
So I think the README should have a link for the compatibility chart, at the beginning of the README. Otherwise we keep reading and thinking everything is compatible, only to find way later that it isn't, and this makes us lose precious time.

It could be something as simple as "you can check platform compatibility per feature here"

miguelpruivo:
@feinstein yes, that sounds like a good idea, I might add a summarized table for the supported features in the README while keeping the details on the wiki.

feinstein:
Excellent.

Issue #915, opened 2021-12-08
Different Directory Picker on Windows


PeterNjeim:
Hello, I (as well as most others) seriously dislike the directory chooser/picker that was chosen for the Windows implementation of this plugin. I feel completely paralyzed using it. I always think of old programs when I see this kind of directory picker. Normally, a directory chooser should look exactly the same as a file chooser, just that the files aren't visible. This is how it's done in virtually every program I've ever used. Please consider updating to the newer Vista+ folder selector using one of the methods listed in the "edit" section of the top answer from this q&a: https://stackoverflow.com/questions/1922204/open-directory-dialog.

PeterNjeim:
Well, I suppose 9 days is close enough to 14 for your bot there lol

philenius:
@PeterNjeim I do understand your request. This

![Screenshot 2022-01-26 223210](https://user-images.githubusercontent.com/9049899/151250536-53112f76-01c5-4aa1-8975-25190e1556a4.png)

looks much better than that:

![image](https://user-images.githubusercontent.com/9049899/151250639-ea4f8a02-d729-4a74-8d2d-249aff8fd814.png)


Thank you for providing the link to StackOverflow. Unfortunately, we cannot reference .NET libraries (which must be pre-installed on Windows). The Win32 API is special in that these DLLs (`comdlg32.dll`, `shell32.dll`, etc.) seem to be present on every Windows installation.

Until now I couldn't figure out which API provides the modern dialog. Currently, we're using [SHBrowseForFolderW](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shbrowseforfolderw) from Win32 API which opens this old-fashioned dialog. In the remarks it says:
> For Windows Vista or later, it is recommended that you use IFileDialog with the FOS_PICKFOLDERS option rather than the SHBrowseForFolder function.  

Maybe the solution is to switch to `IFileDialog` is the solution. We'll just have to figure out how to use the `IFileDialog` API.

PeterNjeim:
I think that's the correct next step

b34st80y:
The pub.dev entry ( https://pub.dev/packages/file_picker ) even shows an example of the modern windows file picker: 

![image](https://user-images.githubusercontent.com/19982841/153626968-1c41fae2-2c43-4c48-a00d-ea570efccdbe.png)

I thought maybe I was doing something wrong when I got the one from windows Vista. :p

[...]

PR #971, opened 2022-02-20
#915 modern directory picker on Windows

philenius:
Changes the implementation of `getDirectoryPath` on Windows so that the dialog looks like other, modern Windows dialogs as requested in #915.

| Previously, the folder picker dialog looked like this: | With this PR |
| ---------- | ------------ |
| grafik | grafik              |

**Win32:**
For this PR, I switched from `SHBrowseForFolderW` to [Microsoft's `IFileDialog`](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ifiledialog) and therefore added `win32` as a dependency. `win32` is written by Flutter's product manager (Tim Sneath) and simplifies access to the Windows APIs. In the long run, I'm planning to refactor the two remaining Windows implementations of `pickFiles()` and `saveFile()` so that they are based on `win32`, too.

**Initial directory:**
The previous implementation based on `SHBrowseForFolderW` did not allow to set the folder where the dialog should open (parameter `initialDirectory`). The new API of `IFileDialog` provides two methods `SetFolder()` and `SetDefaultFolder()` for this purpose. However, I couldn't figure out how to use any of these two methods. We could definitely need some help here.
I fixed the Dart docs of `getDirectoryPath()` because they mentioned that `initialDirectory` was supported on Linux, macOS, and Windows which is wrong. My mistake. Thank you, @julianklose, for noticing that (#970).

**Compatibility chart:**
As requested in #943 by @feinstein, I added a compatibility chart to the README.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.5.0
+
+#### Desktop (Windows)
+Changes the implementation of `getDirectoryPath()` on Windows to provide a modern dialog that looks the same as a file picker dialog ([#915](https://github.com/miguelpruivo/flutter_file_picker/issues/915)).
+
 ## 4.4.0
 
 #### Desktop (Linux, macOS, and Windows)
--- a/README.md
+++ b/README.md
@@ -36,6 +36,18 @@ A package that allows you to use the native file explorer to pick single or mult
 
 If you have any feature that you want to see in this package, please feel free to issue a suggestion. 🎉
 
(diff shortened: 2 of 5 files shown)


feinstein:
Many thanks!

philenius:
@miguelpruivo, I need your review. Thanks.

miguelpruivo (approved):
LGTM. Thank you @philenius! 🎉

PR #993, opened 2022-03-10
fix the input element display value

navaronbracke:
This PR fixes the `<input>` tag's display value. It was not set to none, which results in it showing up in some cases.

### DIFF
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -60,6 +60,7 @@ class FilePickerWeb extends FilePicker {
     uploadInput.draggable = true;
     uploadInput.multiple = allowMultiple;
     uploadInput.accept = accept;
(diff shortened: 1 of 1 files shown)


navaronbracke:
@miguelpruivo could you review this?

miguelpruivo (changes requested):
@navaronbracke could you please describe in more detail what the issue was and also update the CHANGELOG and bump the patch version on pubspec? Thank you.

navaronbracke:
> @navaronbracke could you please describe in more detail what the issue was and also update the CHANGELOG and bump the patch version on pubspec? Thank you.

Will do!

navaronbracke:
@miguelpruivo Fixed the pubspec + changelog and also added reproduction steps in the issue that should be fixed by this PR.

miguelpruivo (approved):
LGTM. Thank you.

Issue #804, opened 2021-08-21
After I picked a file, the current working directory will be modified.


sunjhtrustmobi:
**Describe the bug**
After everytime I picked a file, and I go to visit any files under my project directory by File(path), I will failed.
The system will inform me that it can't find the designated path. Because the file_picker has modified the uri of my project, which caused I can't visit files just by realitived path. So I wonder if there any ways to make it not to modify the path or use a variable to  get the value of the path. 

**Platform**
[√] Flutter (Channel stable, 2.0.4, on Microsoft Windows [Version 10.0.19042.1165], locale zh-CN)
[√] Android toolchain - develop for Android devices (Android SDK version 30.0.1)
[√] Chrome - develop for the web
[√] Visual Studio - develop for Windows (Visual Studio Community 2019 16.9.4)   
[√] Android Studio (version 3.0)
[√] Android Studio (version 3.6)

**Platform OS version**
Microsoft Windows [Version 10.0.19042.1165], locale zh-CN
**How are you picking?**
Pick any picture , and it hanppens.

**Details to reproduce the issue**
Provide all the details to reproduce the issue.

**Error Log**
This error occured after I picked a picture. My program can't find the correct path of the file because I used realitive path.
So I wonder why the FIle_Picker will modify the working directory of the program.
[ERROR:flutter/lib/ui/ui_dart_state.cc(186)] Unhandled Exception: Exception: FileSystemException: Cannot open file, path = 'E:\AndroidStdioProject\Flutter_SecEMS_master\flutter_secemsassets/thumb/img_1629534237605.jpg' (OS Error: 系统找不到指定的路径。
, errno = 3)
#0      _File.throwIfError (dart:io/file_impl.dart:635:7)
#1      _File.openSync (dart:io/file_impl.dart:479:5)
#2      _File.writeAsBytesSync (dart:io/file_impl.dart:604:31)
#3      Luban._large2SmallCompressImage
package:flutter_luban/src/flutter_luban.dart:229
#4      Luban._large2SmallCompressImage
package:flutter_luban/src/flutter_luban.dart:220
#5      Luban._lubanCompress
package:flutter_luban/src/flutter_luban.dart:185

**Screenshots and/or video**
![系统找不到指定路径](https://user-images.githubusercontent.com/77095254/130316655-b3b2eef7-7d9b-4640-8437-6737e50b42ce.png)

**Flutter Version details**
[...]

PR #1022, opened 2022-04-14
add ofnNoChangeDir to make desktop consistent for windows

Tokenyet:
Before this changes, If we used `pickFiles` on Windows, the current directory would be changed. And the behavior is inconsistent with Linux and Macos, that those never modify current directory. Issue: https://github.com/miguelpruivo/flutter_file_picker/issues/804

For example (from example):
```
void _pickFiles() async {
    _resetState();
    try {
      _directoryPath = null;
      print(Directory.current.path);
      _paths = (await FilePicker.platform.pickFiles(
        type: _pickingType,
        allowMultiple: _multiPick,
        onFileLoading: (FilePickerStatus status) => print(status),
        allowedExtensions: (_extension?.isNotEmpty ?? false)
            ? _extension?.replaceAll(' ', '').split(',')
            : null,
      ))
          ?.files;
      print(Directory.current.path);
    } on PlatformException catch (e) {
      _logException('Unsupported operation' + e.toString());
    } catch (e) {
      _logException(e.toString());
    }
    if (!mounted) return;
    setState(() {
      _isLoading = false;
      _fileName =
          _paths != null ? _paths!.map((e) => e.name).toString() : '...';
      _userAborted = _paths == null;
    });
  }
```
Will output: (If I select file from C)
```
C:\flutter_picker_file\example
C:\
```

So I add the `OFN_NOCHANGEDIR` flag to prevent current directory from changing by default on Windows. Reference [here](https://stackoverflow.com/questions/41244561/why-the-current-directory-changes).

Then using the sample code will output:
```
C:\flutter_picker_file\example
C:\flutter_picker_file\example
```

This make the same logic as MacOS and Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.5.2
+#### Desktop (Windows)
+Adds the flag `OFN_NOCHANGEDIR` to prevent current directory from changing on Windows. That makes all platform consistent.
+
 ## 4.5.1
 
 #### Web
--- a/lib/src/windows/file_picker_windows.dart
+++ b/lib/src/windows/file_picker_windows.dart
@@ -268,7 +268,8 @@ class FilePickerWindows extends FilePicker {
     openFileNameW.ref.nMaxFile = lpstrFileBufferSize;
     openFileNameW.ref.lpstrInitialDir =
         (initialDirectory ?? '').toNativeUtf16();
-    openFileNameW.ref.flags = ofnExplorer | ofnFileMustExist | ofnHideReadOnly;
+    openFileNameW.ref.flags =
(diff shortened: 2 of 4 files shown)


miguelpruivo:
@philenius can you take a look?

Tokenyet:
> @Tokenyet would you mind to also update the pubspec and changelog?
> 
> Thank you!

Done! I think It's more like a bug fix rather than a new feature. If there is any problem, please Iet me know.

arthas1888:
I have the same problem

philenius:
hey @Tokenyet, I'm sorry that I wasn't active last year when you initially created this PR. Thank you for your input. I included your changes in #1256 and gave you credit in our `CHANGELOG.md`. Thanks again for pointing me in the right direction and sorry for the delay.

PR #1033, opened 2022-05-17
Get the absolute path of the file you selected. Fix memory leaks.

xuqiqiang:


miguelpruivo:
Hi. What are you doing with this PR? It's just pointing the file's to another package (actually a fork of this one). This is not how one should contribute, hence, it doesn't actually seem to be fixing anything.

Closing.

PR #1094, opened 2022-08-14
Picking files without caching 

abhi16180:
The current version of file_picker caches all the files regardless of type of file. This is very inefficient and cause a lot of delay in fetching the filepath. If caching is completely removed like before ( versions <2. 0. 0) it may not be able to fetch some files. So  I have made changes in way that it will caches the files only if it cannot retrieve the files' paths directly. This will save a lot of time.
 
https://github.com/abhi16180/flutter_file_picker/blob/2c95ee84010b06499330abf206fa0df353e4508b/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java#L227-L293

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,12 +16,12 @@ jobs:
     name: Run unit tests on ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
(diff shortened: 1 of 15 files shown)


miguelpruivo (changes requested):
Please rebase it and clean the unnecessary files from the PR.

Issue #1106, opened 2022-09-01
Migrate to win32 ^3.0.0


d-wolf:
I get the following error when i add [bitsdojo_window](https://pub.dev/packages/bitsdojo_window) to my dependencies:

```
Because no versions of bitsdojo_window match >0.1.4 <0.2.0 and bitsdojo_window 0.1.4 depends on bitsdojo_window_windows ^0.1.4, bitsdojo_window ^0.1.4 requires bitsdojo_window_windows ^0.1.4.
And because no versions of bitsdojo_window_windows match >0.1.4 <0.2.0 and bitsdojo_window_windows 0.1.4 depends on win32 ^3.0.0, bitsdojo_window ^0.1.4 requires win32 ^3.0.0.
And because file_picker 5.0.1 depends on win32 ^2.7.0 and no versions of file_picker match >5.0.1 <6.0.0, bitsdojo_window ^0.1.4 is incompatible with file_picker ^5.0.1.
```

Would be nice if you upgrade the win32 dependency to ^3.0.0.

miguelpruivo:
It has been updated and published with 5.1.0. Thank you!

PR #1109, opened 2022-09-04
Update Dependencies to latest versions (Win32 2.7.0 to 3.0.0)

ishangavidusha:
Issue: #1106 

Add necessary dependency API updates, As mentioned in the Win32 Change Log: 

> COM and Windows Runtime methods and properties are now camelCased, not TitleCased. This is inconvenient, but it avoids a whole class of name clashes and aligns COM and WinRT APIs more closely with Dart idioms. As the projections get smarter with more helpers, we think this is the right call for the future and worth a one-time tax to fix.

**Test Passed**: test\file_picker_windows_test.dart

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.1.0
+#### Desktop (Windows)
+Update Dependencies to latest versions (Win32 2.7.0 to 3.0.0)
+
 ## 5.0.1
 #### Android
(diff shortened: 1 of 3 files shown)


Xazin (commented):
This looks great, honestly having quite a lot of trouble getting file_picker and bitsdojo_window to work together due to this missing link.

miguelpruivo (approved):

miguelpruivo:
Thank you.

PR #1123, opened 2022-09-15
fix caching issue

abhi16180:
Currently file picker caches every single file in android . This method is inefficient when the file size is too large. This fix removes unnecessary caching and loading issue. It will cache only if it cannot retrieve the paths directly (only files in downloads folder will be cached ). This will save a lot of time by eliminating delay which is caused due to caching. (Caching issue is android specific )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 5.2.0
+
+#### Android
+- Fixes issue with caching on android
+- Files will be retrieved without caching to avoid delay (Only cached if paths cannot be retrieved directly )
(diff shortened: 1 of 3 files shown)


abhi16180:
> _ No description provided. _

@miguelpruivo  Please check it out

miguelpruivo (changes requested):
Bump version to 5.2.0 and update CHANGELOG please.

abhi16180:
> Bump version to 5.2.0 and update CHANGELOG please.

Done

miguelpruivo (approved):
LGTM.

Issue #1140, opened 2022-09-27
Not work with Android target sdk version less then 33


waleedkalyar:
Hi, It has an issue while I run project for android. It says that target sdk version must b 33. Is there any solution in which in work with sdk 32 in android? I am stuck in this issue since last 3 days. Some other libraries not work with android 33 so that's way I can increase the target sdk version.

luwei888:
same problem

InAnadea:
The same error 
  ```
.pub-cache/hosted/pub.dartlang.org/file_picker-5.2.0+1/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java:268: error: cannot find symbol
        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
                                                       ^
  symbol:   variable TIRAMISU
  location: class VERSION_CODES
1 error


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':file_picker:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

itismrx:
I am also facing the same error. Here is the output of flutter doctor
` flutter doctor 
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.3.3, on Microsoft Windows [Version 10.0.19044.2006], locale en-KE)
[√] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)       
[√] Chrome - develop for the web
[!] Visual Studio - develop for Windows (Visual Studio Build Tools 2019 16.11.12)
    X The current Visual Studio installation is incomplete. Please reinstall Visual Studio.
[√] Android Studio (version 2021.1)
[√] VS Code (version 1.71.2)
[√] Connected device (4 available)
[√] HTTP Host Availability
`

egoan82:
same problem

herrderkekse:
I had the same issue. It appears to be a problem with package versions 5.0.0 and above. For me, using version 4.6.1 works fine.

miguelpruivo:
Fixed with 5.2.1.

DetachHead:
this issue still happens to me on 5.2.2 @miguelpruivo

Issue #1124, opened 2022-09-15
 variable TIRAMISU not found in version code.


kjawadDeveloper2:
Getting same issue like this on this packege. 

https://github.com/react-native-image-picker/react-native-image-picker/issues/2018

alexmercerind:
@kjawadDeveloper2 Change `compileSdkVersion` in build.gradle to 33 OR pin dependency to `5.0.1` in `pubspec.yaml`.

deakjahn:
The problem with this approach is that we're not supposed to hardwire the SDK version any more but to rely on `flutter.compileSdkVersion` as set in the Flutter SDK. So, plugin developers shouldn't force us to advance it manually and to rely on versions the SDK itself doesn't dictate yet.

PR #1141, opened 2022-09-30
Android tiramisu fix

PatrickWehri:
Android:

Replaced the build variable `Build.VERSION_CODES.TIRAMISU` with its integer `33`.
This variable is not available if the app is targeting sdk 32 or lower, hence resulting in a compile error.

Fixes issues #1140 and #1124


alexmercerind:
Thanks for the change!

miguelpruivo (approved):
LGTM. Thanks.

PR #1169, opened 2022-11-10
Use flutter extension to avoid force the user update the android sdk …

jandrop:
Adding the flutter extension flutter.compileSdkVersion to avoid hardcode the SDK version here, so the user is not forced to update the compileSdkVersion to 33.


jandrop:
Closed because locally it was working but when I added the git library to the project it fails. I will investigate the problem.

PR #1172, opened 2022-11-13
make readStreamChunkSize property public

yanglinzhen:
I am picking file for a browser and wish to customize the chunk size of file stream.

### DIFF
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -9,8 +9,6 @@ class FilePickerWeb extends FilePicker {
   late Element _target;
   final String _kFilePickerInputsDomId = '__file_picker_web-file-input';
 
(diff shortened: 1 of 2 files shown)


lib/_internal/file_picker_web.dart
```diff
@@ -9,8 +9,6 @@ class FilePickerWeb extends FilePicker {
   late Element _target;
   final String _kFilePickerInputsDomId = '__file_picker_web-file-input';
 
-  final int _readStreamChunkSize = 1000 * 1000; // 1 MB
```
yanglinzhen, line 12: Should be 1024 instead of 1000

miguelpruivo (changes requested):
Got your point, but a setter should be used instead of a public non-final property.

PR #1180, opened 2022-11-28
Fix an issue when picking live photos where the key photo was changed

Ackar:
When picking multiple live photos where the key photo was changed, the same one is returned multiple times.

This is because the filename used when converting the photo from pvt to jpeg is the same for all pictures where the key photo was updated.

For example, here is the content of a regular live photo:
```
IMG_0042.pvt
|- IMG_0042.MOV
|- metadata.plist
|- IMG_0042.HEIC
```

After the key photo is changed, the content changes to:
```
IMG_0042.pvt
|- FullSizeRender.mov
|- metadata.plist
|- FullSizeRender.heic
```

This solves the issue by using the original filename (e.g. `IMG_0042.pvt`) instead of the renamed file (`FullSizeRender.heic`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.2.3
+### iOS
+- Fixes an issue when picking live photos where the key photo was changed.
+
 ## 5.2.2
 ### Android
 - Fix deprecation warning for `getParcelable(String key)` method.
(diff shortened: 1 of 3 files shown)


miguelpruivo (approved):
LGTM thanks.

PR #1188, opened 2022-12-08
Fix/call file dialog in isolate

SylphiaWindy:
GetOpenFileNameW and GetSaveFileNameW are sync methods, so if we call them in UI thread, parent windows will stop responding.
Call them in a separate isolate to match same behaviour to macOS.


miguelpruivo (approved):
LGTM. Thanks

SylphiaWindy:
Seems there're formatting errors, I'll fix them once I get home.

Issue #1208, opened 2023-01-08
Cannot start file_picker with the launch setting nor run command


LeonardoMussato:
**Describe the bug**
While I try to run the app on an IOS emulator a recive the following errors. I'm using the latest version 5.2.4 with flutter 3.3.8 with the Very Good Core framework. My PodFile is like this:
```
# Uncomment this line to define a global platform for your project
# platform :ios, '11.0'

# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
ENV['COCOAPODS_DISABLE_STATS'] = 'true'

project 'Runner', {
  'Debug' => :debug,
  'Profile' => :release,
  'Release' => :release,
}

def flutter_root
  generated_xcode_build_settings_path = File.expand_path(File.join('..', 'Flutter', 'Generated.xcconfig'), __FILE__)
  unless File.exist?(generated_xcode_build_settings_path)
    raise "#{generated_xcode_build_settings_path} must exist. If you're running pod install manually, make sure flutter pub get is executed first"
  end

  File.foreach(generated_xcode_build_settings_path) do |line|
    matches = line.match(/FLUTTER_ROOT\=(.*)/)
    return matches[1].strip if matches
  end
  raise "FLUTTER_ROOT not found in #{generated_xcode_build_settings_path}. Try deleting Generated.xcconfig, then run flutter pub get"
end

require File.expand_path(File.join('packages', 'flutter_tools', 'bin', 'podhelper'), flutter_root)

flutter_ios_podfile_setup

Pod::PICKER_AUDIO = false
Pod::PICKER_DOCUMENT = false

target 'Runner' do
  use_frameworks!
  use_modular_headers!

  flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
end

post_install do |installer|
  installer.pods_project.targets.each do |target|
    flutter_additional_ios_build_settings(target)
  end
end
```

**Platform**

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
MacOS 12.6 with Macbook pro M1 Max

**How are you picking?**
```
Nothing at the moment because the error occured while building
```

**Details to reproduce the issue**
The error appear everytime I try to run the app

**Error Log**Archiving com.uaffamily.mobile...
Automatically signing iOS for device deployment using specified development team in Xcode project: 8YS26433MD
Running pod install...                                           1,912ms
CocoaPods' output:
↳
      Preparing

[...]

PR #1209, opened 2023-01-08
Fix preprocessor definitions in podspec

tomk9:
https://github.com/miguelpruivo/flutter_file_picker/issues/1208


miguelpruivo:
Closes #1208.

PR #1230, opened 2023-02-23
fix: Android chooses files with the same name and the result is abnormal.

xuelongqy:
Selecting files with the same name from different directories, the cache always keeps the first selected file, so the results obtained from the subsequently selected files are incorrect.


xuelongqy:
This doesn't completely solve the problem, as selecting files with the same name from different directories can still result in overwriting. I think adding file hash for verification could be a solution.

xuelongqy:
I made some changes. As calculating the MD5 value of the file takes a long time, I used the hashCode of the URI as the directory name to prevent conflicts of files with the same name.

xuelongqy:
> Would it better to use hash as filename prefix rather subfolder? since in most cases, their hash would be different. Then each subfolder will contains only one file. It's somehow redundant.

There are two reasons for this:
- Calculating the file hash is a time-consuming operation 
- Using the hash as part of the file name, you cannot get the original file name.

xuelongqy:
> ```java
>         boolean fileExisted = file.exists();
>         if (fileExisted) {
>             try {
>                 fileExisted = file.length() == context.getContentResolver().openFileDescriptor(uri, "r").getStatSize();
>             } catch (final Exception ignored) {}
>         }
> ```
> 
> if exception raised, why keep `fileExisted` true? mostly, is that caused by filesystem or invalid `context`? I don't know android/java dev, but guess better to make fileExisted false in catch block?

You are right, but in general, I don’t think it will happen.

miguelpruivo:
This have been already fixed. Thanks.

PR #1241, opened 2023-03-14
Enhancements to stream API

karlprieb:
- Allows creating multiple streams from single picked file
- Allow specifying offset/length of stream
- Uses Uint8List instead of List<int> as stream datatype
- Support for `lastModified`

Same code from #1217 with fixes for `lastModified`. Now it works on  android, ios and all major browsers (including Firefox which had issues on the other PR)


miguelpruivo (commented):
If you think this is still worth it as of today, please rebase and update the changelog + bump minor version. Thanks!

vicajilau:
This branch is obsolete, and since it is an obsolete pull request, we have decided to close it. We apologize for the inconvenience.

Issue #989, opened 2022-03-06
Prompt to confirm saving when file already exists


merilec:
When selecting a location and file name to save a file to (via `saveFile`), a file with that file name might already exist. The picker should then prompt the user to confirm saving and replace it, similar to how it is done in many other programs.

![image](https://user-images.githubusercontent.com/99778788/156919044-1d5687d5-3189-4227-84da-6aa3b9501067.png)

bhlin790:
Is there any version of this feature planned to be implemented?

philenius:
@bhlin790, sorry for keeping you waiting so long. Contributing to open-source has been a low priority for me for the last 12 months.
The Linux implementation is ready. Now, I'm working on the macOS and Windows implementation.

philenius:
Will be implemented with v5.2.7. Now, we're waiting for approval from Miguel :wink:

PR #1248, opened 2023-03-30
#989: save file on Windows: prompt user for confirmation when selecting an already existing file

Yadav-Sunil:
~~Downgrade ffi version~~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: subosito/flutter-action@v1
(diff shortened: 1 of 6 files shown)


philenius:
@miguelpruivo 
@merilec
@bhlin790

Resolves ##989 for `saveFile()` on Windows:
* With our previous implementation, Windows did not show a confirmation dialog when showing the `saveFile()` dialog.
* This PR makes Windows show a confirmation dialog when the user selects an already existing file. Please see the following screen recording:  
   confirm-overwrite

Other, minor changes:
* I removed a duplicate step in our CICD pipeline which would setup Flutter twice.

* I changed the emojis in our README from previously  
  image
  to:
  image

Issue #1260, opened 2023-04-12
Flutter 3.0.5 does not compile on file_picker 5.2.8


micaelcid:
Currently, we're using Flutter 3.0.5 and as soon as we updated file_picker to 5.2.8, we started getting the following exception while compiling tests or compiling the app itself:
```
.pub-cache/hosted/pub.dartlang.org/file_picker-5.2.8/lib/src/platform_file.dart:53:9: Error: Type 'Uint8List' not found.
  final Uint8List? bytes;
        ^^^^^^^^^
.pub-cache/hosted/pub.dartlang.org/file_picker-5.2.8/lib/src/platform_file.dart:53:9: Error: 'Uint8List' isn't a type.
  final Uint8List? bytes;
```

After I manually downgraded this package back to 5.2.7, the compilation error disappeared.

By taking a glance over the change logs, I couldn't find what could cause this issue. Was this problem expected?

mateusvtt:
Having the same issue here

miguelslemos:
The issue is related to removed import here https://github.com/miguelpruivo/flutter_file_picker/commit/9d1d9efc3d70385b459d84adb431629f417a33f0#diff-38d8f870fecf77b1946a59ac5ed87e3eac24c8a86dc31ec3fd8970896cf8c016L2

philenius:
Thank you @miguelslemos for identifying the problematic commit. Turns out, that blindly following the Linter's recommendations, can be dangerous. Sorry for the inconvenience. Will be fixed with `file_picker` 5.2.9.

OutdatedGuy:
That's interesting, I checked `package:flutter/foundation.dart` exports and it does export `Uint8List` from the file `src/foundation/serialization.dart` as:

```dart
export 'dart:typed_data' show ByteData, Endian, Float32List, Float64List, Int32List, Int64List, Uint8List;
```

weird bug it seems.

PR #1178, opened 2022-11-16
Increase test coverage to utils.dart

victordmp:
### Initial Proposed

* Improve test coverage to utils.dart on macOs.

### My Proposed

* Improve the test suite coverage of `filePathsToPlatformFiles()`.
  * Create two more test cases.
* Create test suite for `runExecutableWithArguments()` function.
  * There are no tests for it in `file_picker_utils_test.dart`.
 
### Objective

* go from 52.2% coverage to 100% coverage for the utils file functions

### Setup

* macOs Ventura 13.0

#### Improve the test suite coverage of `filePathsToPlatformFiles()`.

* Test 1:
   * **Coverage Before:**
    Screenshot 2022-11-16 at 14 57 48
   * **Added test case**
```dart
test(
    'should tranform a list of file paths containing a path into a list of PlatformFiles',
    () async {
  final filePaths = <String>['test'];

  final platformFiles = await filePathsToPlatformFiles(
    filePaths,
    true,
    false,
  );

  expect(platformFiles.length, equals(filePaths.length));
});
```
   * **Coverage After:**
    Screenshot 2022-11-16 at 15 00 14

* Test 2: 
   * **Coverage Before:**
    Screenshot 2022-11-16 at 15 00 14

   * **Added test case**
```dart
test(
    'should transform a list of file paths containing a valid path into a list of PlatformFiles',
    () async {
  final filePaths = <String>['test/test_files/test.pdf'];

  final platformFiles = await filePathsToPlatformFiles(
    filePaths,
    false,
    true,
  );

  expect(platformFiles.length, equals(filePaths.length));
});
```
   * **Coverage After:**
      Screenshot 2022-11-16 at 15 02 19

#### Create test suite for `runExecutableWithArguments()` function.

* **Coverage Before:**
Screenshot 2022-11-16 at 15 02 19

* **Added test case**
```dart
group('runExecutableWithArguments()', () {
  test('should catch an exception when sending an empty filepath', () async {
    final filepath = '';

    expect(
      () async => await isExecutableOnPath(filepath),
      throwsA(isA<Exception>()),
    );
  });
});
```

* **Coverage After:**
Screenshot 2022-11-16 at 15 04 01

### Checking the changes

- [x] Update CHANGELOG.

PR #1261, opened 2023-04-12
fix: replace broken GIFs in README by videos and add missing import statement

philenius:
Contains two fixes:
* The GIFs in the README were broken: @miguelpruivo, I'm sorry for the trouble, when I committed the new GIF files, I was actually still playing around trying to find the best way to integrate them into the Git repo without committing too large GIF files. I shouldn't have committed my changes or at least mark the PR as _draft_.
With this PR, I uploaded mp4 videos instead of GIF files and I did this via the online editor of GitHub. GitHub provides drag-and-drop functionality so that one can insert images and videos directly into READMEs. The uploaded files are stored outside of the Git repository. For further info, see https://stackoverflow.com/a/4279746.
You can preview the new README here: https://github.com/miguelpruivo/flutter_file_picker/tree/patch/fix-broken-links-in-readme-and-Uint8List-type-not-found
I you prefer the GIF approach, I can revert to the GIF files, which we can store inside our Git repo, but then I'd have to shorten the GIFs (in time & size).

* #1260: This reverts commit 9d1d9efc3d70385b459d84adb431629f417a33f0 because it caused the errors `Type 'Uint8List' not found` and `'Uint8List' isn't a type`. Thank you @miguelslemos for identifying the problematic commit.

Contains one improvement by @victordmp who created PR #1178 to increase the code coverage. I cherry-picked his commits into this PR. With these additional test cases, the code coverage of `utils.dart` increases from 
image

to
image

Thank you @victordmp  :rocket:


philenius:
@miguelpruivo, this time my pull request is ready to be merged :wink:

miguelpruivo:
Thank you @philenius! In pub.dev however it seems that they don’t allow videos inline with the readme. We should probably stick with gifs instead. 😅

philenius:
Oh man, alright, I'll revert to GIF files in the next PR... :face_with_spiral_eyes: 

Thank you for your quick response.

Issue #1255, opened 2023-04-06
Update win32 please


kirill-21:
![image](https://user-images.githubusercontent.com/63228361/230415815-89e38310-6cec-438e-97de-f13c1f29940d.png)

philenius:
Good point!

Same issue: #1222

I'm planning to merge the bugfix branch that I'm currently working on and then release a new major version of `file_picker` with `win32` 4.1.1.

philenius:
I will fix https://github.com/miguelpruivo/flutter_file_picker/issues/1257 first, and then bump the win32 version. As of right now, I'm still undecided whether we should bump our major or minor version after updating `win32`. SemVer does not specify how to handle major dependency upgrades and there are several discussions, e.g.:
* https://softwareengineering.stackexchange.com/questions/168647/best-practices-for-versioning-project-after-dependency-upgrade?rq=1
* https://github.com/semver/semver/issues/148

Issue #1222, opened 2023-02-02
win32


kirill-21:
Because file_picker 5.2.5 depends on win32 ^3.0.0 and no versions of file_picker match >5.2.5 <6.0.0, file_picker ^5.2.5 requires win32 ^3.0.0.
So, because swiftsoft depends on both file_picker ^5.2.5 and win32 ^4.1.1, version solving failed.
pub get failed

PR #1266, opened 2023-04-19
#1255: update dependency `win32` to latest major v4

philenius:
Updates the dependency `win32`, which is used in our Windows `file_picker` implementation, to the latest major v4. See `win32`'s changelog: https://pub.dev/packages/win32/changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.2.11
+### Desktop (Windows)
+Updates dependencies, including win32 bump to 4.1.3 ([#1255](https://github.com/miguelpruivo/flutter_file_picker/issues/1255)).
+
 ## 5.2.10
 ### Desktop (Windows)
 Fixes the bug that the result of the save-file dialog was incorrect when it was invoked with a long default file name but the user selected a file with a much short file name ([#1257](https://github.com/miguelpruivo/flutter_file_picker/issues/1257)).
--- a/README.md
+++ b/README.md
@@ -156,7 +156,6 @@ For full usage details refer to the **[Wiki](https://github.com/miguelpruivo/flu
 #### Linux
 ![DemoLinux](https://github.com/miguelpruivo/flutter_file_picker/blob/master/example/screenshots/example_linux.gif)
 
-Android
 #### Windows
 ![DemoWindows](https://github.com/miguelpruivo/flutter_file_picker/blob/master/example/screenshots/example_windows.gif)
 
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   plugin_platform_interface: ^2.1.2
   ffi: ^2.0.1
   path: ^1.8.0
-  win32: ^3.0.0
+  win32: ^4.1.1
 
 dev_dependencies:
   lints: ^2.0.0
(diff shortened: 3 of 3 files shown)


philenius:
@miguelpruivo, there's a new major release of our dependency `win32` available (`3.0.0` -> `4.1.3`). I re-build the example app on Windows and tested all functionalities (pick file(s), pick folder, save file). Everything still works. The changes in `win32` did not affect the APIs which we're using.
My question is: how do we want to reflect this major dependency update in the version number of `file_picker`? We could bump our major version or just the patch version. There are ongoing discussions around this topic and there's no official or correct solution (e.g. on https://github.com/semver/semver/issues/148 or many questions on StackOverflow). In general, the consensus seems to be that if there were no changes to the public API of the consuming package (in this case `file_picker`), then the patch version should be incremented.

miguelpruivo:
@philenius since it won't introduce breaking changes, patch version should be enough.

miguelpruivo (approved):

zakblacki:
This doesn't work 

```
`Because path_provider_windows >=2.0.0 <2.1.1 depends on win32 ^2.0.0 and path_provider_windows >=2.1.1 <2.1.3 depends on win32 ^2.1.0, path_provider_windows >=2.0.0 <2.1.3 requires win32 ^2.0.0.
And because path_provider_windows >=2.1.3 depends on win32 >=2.1.0 <4.0.0, path_provider_windows >=2.0.0 requires win32 >=2.0.0 <4.0.0.
And because file_picker >=5.2.11 depends on win32 ^4.1.3 and shared_preferences_windows >=2.0.0 depends on path_provider_windows ^2.0.0, file_picker >=5.2.11 is incompatible with shared_preferences_windows >=2.0.0.
Because shared_preferences >=2.1.0 depends on shared_preferences_windows ^2.2.0 and shared_preferences >=2.0.10 <2.1.0 depends on shared_preferences_windows ^2.0.1, shared_preferences >=2.0.10 requires shared_preferences_windows ^2.0.1.
Thus, file_picker >=5.2.11 is incompatible with shared_preferences >=2.0.10.
And because shared_value 3.0.0 depends on shared_preferences ^2.0.15 and no versions of shared_value match >3.0.0 <4.0.0, file_picker >=5.2.11 is incompatible with shared_value ^3.0.0.
So, because com.dz.store depends on both shared_value ^3.0.0 and file_picker ^5.2.11, version solving failed.
pub get failed
command: "C:\flutter\bin\cache\dart-sdk\bin\dart __deprecated_pub --directory . get --example"
pub env: {
  "FLUTTER_ROOT": "C:\flutter",
  "PUB_ENVIRONMENT": "flutter_cli:get",
  "PUB_CACHE": "C:\Users\zakbl\AppData\Local\Pub\Cache",
}
exit code: 1`
```

philenius:
@zakblacki , thanks for posting this directly in this PR instead of opening a new issue :)

I'm sorry, we cannot make everyone happy. We had to upgrade win32 because other folks had their own dependency conflicts, too, e.g. https://github.com/miguelpruivo/flutter_file_picker/issues/1255 and https://github.com/miguelpruivo/flutter_file_picker/issues/1222.

The only solution for you would be to further restrict the version of `file_picker` in your `pubspec.yaml`, e.g.:
```yaml
dependencies:
  file_picker: '<5.2.11'
```

PR #1267, opened 2023-04-26
Adds Android namespace

asaarnak:
Adds a namespace attribute to the Android build.gradle, for compatibility with Android Gradle Plugin 8.0.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,6 +22,7 @@ rootProject.allprojects {
 apply plugin: 'com.android.library'
 
 android {
+    namespace 'com.mr.flutter.plugin.filepicker'
     compileSdkVersion 33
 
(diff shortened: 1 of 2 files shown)


Veteraanikoodari:
Revived an old project that has been dormant for almost two years... This is the last showstopper after all the other update problems. Waiting for merge!

davidmartos96:
I'm encountering with this as well. Thanks for the PR. 
It looks like the namespace can be conditionally added so that it works in older versions. This is how the official plugins do it:

```gradle
android {
    // Conditional for compatibility with AGP <4.2.
    if (project.android.hasProperty("namespace")) {
        namespace 'io.flutter.plugins.camera'
    }
    ....
}
```

Issue #1270, opened 2023-05-02
unable to pick files on Ubuntu 22.04.2 desktop


AbikMathew:
**The issue**
I tried to use the flutter_file_picker package to pick some files on my ubuntu laptop and it is not working. So I went onto try the example app from the github repo and the same issue occurs with that one also. When we call the method to pick files, dialog box won't appear. 

**Platform**
- [ ] Desktop

**Platform OS version**
Ubuntu 22.04.02

**How are you picking?**
``` I tired to pick it inside a project after importing the package and also tried to use the example app available on this github repo itself. 
  void _pickFiles() async {
    _resetState();
    try {
      _directoryPath = null;
      _paths = (await FilePicker.platform.pickFiles(
        type: _pickingType,
        allowMultiple: _multiPick,
        onFileLoading: (FilePickerStatus status) => print(status),
        allowedExtensions: (_extension?.isNotEmpty ?? false)
            ? _extension?.replaceAll(' ', '').split(',')
            : null,
        dialogTitle: _dialogTitleController.text,
        initialDirectory: _initialDirectoryController.text,
        lockParentWindow: _lockParentWindow,
      ))
          ?.files;
    } on PlatformException catch (e) {
      _logException('Unsupported operation' + e.toString());
    } catch (e) {
      _logException(e.toString());
    }
    if (!mounted) return;
    setState(() {
      _isLoading = false;
      _fileName =
          _paths != null ? _paths!.map((e) => e.name).toString() : '...';
      _userAborted = _paths == null;
    });
  }
```

**Details to reproduce the issue**
Just run the example app on an ubuntu 22.04 machine, and you won't be able to pick any files.

**Error Log**
There are no errors or exceptions caught

**Screenshots and/or video**
![image](https://user-images.githubusercontent.com/97751300/235653242-64a88195-761a-4fb5-b8cb-fd93aa834a79.png)

**Flutter Version details**
[!] Flutter (Channel unknown, 3.7.7, on Ubuntu 22.04.2 LTS 5.19.0-40-generic, locale en_US.UTF-8)
    ! Flutter version 3.7.7 on channel unknown 

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /home/abik/Android/Sdk
    • Platform android-33-ext5, build-tools 33.0.2
    • Java binary at: /snap/android-studio/125/android-studio/jre/bin/java
[...]

PR #1275, opened 2023-05-07
Refactor file_picker_linux to use the XDG FileChooser portal

MateusRodCosta:
This PR moves from using zenit, qarma and kdialog to use the [FileChooser portal](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.FileChooser) instead.

This issue is related to #1270.

Also, currently setting as a draft due to not implementing unit tests yet and due to having to check how to deal with the uris.


miguelpruivo:
@philenius could you review this?

MateusRodCosta:
I am moving this as ready for review.

A few notes:

Only the method for saving files on the FileChooser portal support setting an initial directory right now, they plan on enabling it for the opening files method in the future, see https://github.com/flatpak/xdg-desktop-portal/issues/796.
Also, while it could be enabled, I hit an issue where the file chooser actually refuses to access the directory specified so I decided to disable it. The code I was trying to use is commented out, and the issue related to that seems to be https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/29.

It seems the FileChooser API supports choosing whether dialogs modal, however since on my testing I couldn't see it working, I didn't implement support for that.

I added a small converter from the `file://` uri to a normal file path, since I believe that's how file_picker would like it, but I don't know whether it would work well on a sandbox such as the ones from flatpak or snap.

Also, tests are broken. Since I rewrote completely how the linux side works, I would have to re-do them from scratch, that is assuming I can figure out how.

philenius:
Thank you very much for your contribution, @MateusRodCosta. I'll look into your PR within the next few days.

lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,139 @@
...
+
+    final result = await client.fileChooser
+        .saveFile(
+          title: dialogTitle ?? defaultDialogTitle,
+          filters: [fileFilter],
+          currentName: fileName,
+          //currentFolder: directory,
+        )
+        .first;
+    return _uriToFilePath(result.uris.first);
```
philenius, line 137: And once more:

```suggestion
    try {
      final result = await client.fileChooser
          .saveFile(
            title: dialogTitle ?? defaultDialogTitle,
            filters: [fileFilter],
            currentName: fileName,
            //currentFolder: directory,
          )
          .first;
      return _uriToFilePath(result.uris.first);
    } on XdgPortalRequestFailedException {
      return null;
    }
```

lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,139 @@
...
+    final client = XdgDesktopPortalClient();
+
+    final result = await client.fileChooser
+        .openFile(
+          title: dialogTitle ?? defaultDialogTitle,
+          directory: true,
+        )
+        .first;
+
+    return _uriToFilePath(result.uris.first);
```
philenius, line 100: Same issue with cancelling the dialog:

```suggestion
    try {
      final result = await client.fileChooser
          .openFile(
            title: dialogTitle ?? defaultDialogTitle,
            directory: true,
          )
          .first;
      return _uriToFilePath(result.uris.first);
    } on XdgPortalRequestFailedException {
      return null;
    }
```

lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,139 @@
...
+
+    final filePaths = result.uris.map(_uriToFilePath).toList();
+
+    final List<PlatformFile> platformFiles = await filePathsToPlatformFiles(
+      filePaths,
+      withReadStream,
+      withData,
+    );
+
+    return FilePickerResult(platformFiles);
```
philenius, line 82: If the user cancels the dialog, this implementation throws an exception instead of returning `null`. The latter is how `file_picker` works across all platforms. We can easily solve this by catching exceptions of type `XdgPortalRequestFailedException`:

```suggestion
    try {
      final result = await client.fileChooser
          .openFile(
            title: dialogTitle ?? defaultDialogTitle,
            filters: [fileFilter],
            multiple: allowMultiple,
            directory: false,
            parentWindow: "example",
          )
          .first;
      final filePaths = result.uris.map(_uriToFilePath).toList();

      final List<PlatformFile> platformFiles = await filePathsToPlatformFiles(
        filePaths,
        withReadStream,
        withData,
      );
      return FilePickerResult(platformFiles);
    } on XdgPortalRequestFailedException {
      return null;
    }
```

philenius (changes requested):

philenius:
@MateusRodCosta thanks for your implementation.


> I added a small converter from the file:// uri to a normal file path, since I believe that's how file_picker would like it, but I don't know whether it would work well on a sandbox such as the ones from flatpak or snap.

The conversion works fine :+1: 

> Also, tests are broken. Since I rewrote completely how the linux side works, I would have to re-do them from scratch, that is assuming I can figure out how.

No problem, we can fix the test later. Let's rather focus on two issues here:

> Only the method for saving files on the FileChooser portal support setting an initial directory right now, they plan on enabling it for the opening files method in the future, see https://github.com/flatpak/xdg-desktop-portal/issues/796.
Also, while it could be enabled, I hit an issue where the file chooser actually refuses to access the directory specified so I decided to disable it. The code I was trying to use is commented out, and the issue related to that seems to be https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/29.

This would be a downgrade because our current solution does support setting the initial directory.

> It seems the FileChooser API supports choosing whether dialogs modal, however since on my testing I couldn't see it working, I didn't implement support for that.

I would consider this a blocker as long as we cannot figure our how to make it behave like a modal. Our current implementation provides already a limited usability since the file picker dialog **sometimes** opens in the background. However, I tested the FileChooser Portal implementation on Ubuntu 23 and  the file picker dialog is **always** spawned behind the actual Flutter app.

https://github.com/miguelpruivo/flutter_file_picker/assets/9049899/f8f2ae34-2c3a-430e-8dd5-419e13628077

In contrast, our current implementation:

https://github.com/miguelpruivo/flutter_file_picker/assets/9049899/2405413d-8fc6-461e-b296-77eff1356616


I'm sorry, I definitely see advantages in your implementation but I don't want to merge this PR until we can solve the modal issue.

philenius:
I posted a question in https://github.com/canonical/xdg_desktop_portal.dart/issues/87. Maybe we'll figure it out this way.

MateusRodCosta:
> > Only the method for saving files on the FileChooser portal support setting an initial directory right now, they plan on enabling it for the opening files method in the future, see [flatpak/xdg-desktop-portal#796](https://github.com/flatpak/xdg-desktop-portal/issues/796).
> > Also, while it could be enabled, I hit an issue where the file chooser actually refuses to access the directory specified so I decided to disable it. The code I was trying to use is commented out, and the issue related to that seems to be https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/29.
> 
> This would be a downgrade because our current solution does support setting the initial directory.

Yeah, makes sense, can't do much besides waiting for the upstream to actually do it =/.

From reading the related issue, it seems to be related to how the following three values are filled : current_directory, current_name and current_file.

I am using right now a combination of both current_name and current_directory. Maybe the bug could be fixed if when using both at the same time we used current_file instead? I will experiment with that.

> > It seems the FileChooser API supports choosing whether dialogs modal, however since on my testing I couldn't see it working, I didn't implement support for that.
> 
> I would consider this a blocker as long as we cannot figure our how to make it behave like a modal. Our current implementation provides already a limited usability since the file picker dialog **sometimes** opens in the background. However, I tested the FileChooser Portal implementation on Ubuntu 23 and the file picker dialog is **always** spawned behind the actual Flutter app.
> 2023-05-09-22-18-26_uyjUnumj_cropped.mp4
> 
> In contrast, our current implementation:
> 2023-05-09.22-34-07_trimmed_cropped.mp4
> 
> I'm sorry, I definitely see advantages in your implementation but I don't want to merge this PR until we can solve the modal issue.

Ok, so this I think I know why.
If you noticed, the portal version pops up a notification, that's because it's triggering GNOME's focus stealing prevention. Which means that basically GNOME thought it was some other application which didn't deserve to steal focus from whatever important thing you were doing.

First, the default for modals seem to be true:

> modal b
> 
> Whether the dialog should be modal. Default is yes. 

But it also can take the info of the parent window:

> IN s parent_window:
> 
> Identifier for the application window, see [Common Conventions](https://flatpak.github.io/xdg-desktop-portal/#parent_window)

On my system (Fedora + GNOME Wayland) it pops up in front, it might behave a bit diferently in X11.
Anyway, I am don't know how to get the window identifier, but using it would probably fix it.

lib/src/file_picker_linux.dart
MateusRodCosta: Weird, for me they always appear as `flutter: Request was cancelled`, but I will implement these changes.

MateusRodCosta:
I retried implementing the initialDirectory support to saveFile, this time by also including the "current_file" value.

It still didn't work =/.

I starting to think the issue might either be upstream on the xdg portal Flutter plugin or on the GNOME/GTK portal.

miguelpruivo:
Closing due to inactivity.

MateusRodCosta:
Did any updates happen on the Ubuntu-maintained xdg packages?

Also as a note we could do this via the native GTK integration by calling GTKFileChooserNative in C.

Issue #1300, opened 2023-06-02
Not picking up multiple images file on real device image gallery but on Google photos its working


joshua750:
Hello its not pickup multiple images on real devices but on android emulator its working fine my real device is real me c3 android version 10 i guess it only working on google photos not phone gallery

  Future<void> pickImage() async {
    try {
      // Request permission to access storage
      await Permission.storage.request();

      final result = await FilePicker.platform.pickFiles(
        allowMultiple: true,
        type: FileType.image,
      );

      if (result != null && result.files.isNotEmpty) {
        final appDir = await getApplicationDocumentsDirectory();
        final imagesDir = Directory('${appDir.path}/images');
        if (!await imagesDir.exists()) {
          await imagesDir.create(recursive: true);
          print('Created directory: $imagesDir');
        }

        for (var file in result.files) {
          final fileName = file.name;
          final filePath = file.path;
          final newPath = '${imagesDir.path}/$fileName';

          print('This filepath $filePath');
          // Move the image file to the images directory
          final File sourceFile = File(filePath!);
          final File destinationFile = await sourceFile.copy(newPath);

          print('Moved image: ${destinationFile.path}');

          await sourceFile.delete();
          print('Deleted image: $filePath');
        }

        refreshScreen();
      }
    } catch (e) {
      print('Error picking images: $e');
    }
  }

DarenF-20C:
Facing this issue too, happens on some OS built-in gallery app. 
Google's Photos won't have the issue.

So far tested, 
1. Vivo (Funtouch OS, Android 13) built-in gallery, Albums.
2. Oppo (ColorOS 11.1, Android 11) built-in gallery, Photos (same name with google one but different icon).
3. XiaoMi built-in FileManager

PR #1303, opened 2023-06-12
fix issue #1300: support pick multi-images on Android

weibili:


DarenF-20C:
@weibili Hi, is there's any problem with the fix? why it's deleted? cause when I use it solve my issue

PR #1342, opened 2023-08-07
fix(android): build.gradle namespace

DrBu7cher:
Make it conditional, because it could break the build on Gradle < 8.0

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -38,5 +38,7 @@ android {
         implementation "androidx.lifecycle:lifecycle-runtime:2.1.0"
     }
 
-    if (project.android.hasProperty("namespace")) { namespace 'com.mr.flutter.plugin.filepicker' }
+    if (project.android.hasProperty("namespace")) {
+        namespace 'com.mr.flutter.plugin.filepicker' 
+    }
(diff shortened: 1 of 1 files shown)


miguelpruivo (commented):
I have been using gradle < 8 without any issues without this. Would you mind to elaborate?

navaronbracke:
This is only an issue for the Android Gradle Plugin < 4.2 iirc

android/build.gradle
```diff
@@ -38,5 +38,5 @@ android {
         implementation "androidx.lifecycle:lifecycle-runtime:2.1.0"
     }
 
-    namespace 'com.mr.flutter.plugin.filepicker'
+    if (project.android.hasProperty("namespace")) { namespace 'com.mr.flutter.plugin.filepicker' }
```
navaronbracke, line 41:
```suggestion
    if (project.android.hasProperty("namespace")) {
        namespace 'com.mr.flutter.plugin.filepicker' 
    }
```

navaronbracke (changes requested):

navaronbracke (approved):

PR #1350, opened 2023-08-15
Add initialDirectory for iOS

toptalslr:
Hello @miguelpruivo,

I'm reaching out about a feature that has been requested for a while – the ability to pass an initialDirectory to the platform code. I've been working on a pull request for this feature for iOS, but I've hit a roadblock. Even though the initialDirectory is being passed correctly, the picker isn't opening at the specified location. I've tried a few different approaches without success.

I was wondering if you could take a look at my changes, give me your thoughts, and perhaps collaborate to get this feature working for iOS. Your expertise would be very valuable.

Thanks for your time.

Best,
Adrian


miguelpruivo:
Is this still working as of the latest iOS 17?

PR #1354, opened 2023-08-22
[iOS][Gallery] - Multi image selection fixes

GabrielAraujo:
This PR contains fixes for the following:

- When you selected multiple images and for some reason they had the same file name the plugin would not properly handle the situation and would cause the same file to be repeated on all selected files instead of the files itself.
- When you select multiple images the resulting list order would be different than the originally selected order by the user.


miguelpruivo:
Thanks. Please update the patch version to `.5.`

GabrielAraujo:
Done! It is 5.5.0 now

miguelpruivo:
Thanks!

PR #1356, opened 2023-08-23
fix getDirectoryPath() for window crash

sbrighth:
compare to https://github.com/dart-windows/win32/blob/main/example/dialogshow.dart
xxx.release() is useless and causes application crash


miguelpruivo:
Already fixed.

PR #1360, opened 2023-08-31
export utils.dart

yudaprama:
Need to use some function on utils.dart


miguelpruivo:
Utils won't be exposed as its intended for internal use only. You should use another lib or extract yourself to your project the function needed

Issue #1349, opened 2023-08-14
Google Drive pickers doesn't filter docs by type


JaseElder:
**Describe the bug**
Using
`file_picker: ^5.3.3`

When picking from Google Drive, using the `allowedExtensions:` doesn't filter.
Works fine for device-local file system.

**Platform**

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
Android 12

**How are you picking?**
```
                          onPressed: () async {
                            FilePickerResult? result = await FilePicker.platform.pickFiles(
                              type: FileType.custom,
                              allowedExtensions: ['pdf'],
                            );
                            if (result != null) {
                              var addedFile = await _addPdfToDir(result, ref);
                              print('added ${addedFile.path} ${addedFile.existsSync()}');
                            }
                          },
```

**Details to reproduce the issue**
Use the picker with snippet above to select from Google Drive. All files show, not just PDFs

**Flutter Version details**
```
 %  flutter doctor -v
[✓] Flutter (Channel stable, 3.10.6, on macOS 13.5 22G74 darwin-x64, locale en-GB)
    • Flutter version 3.10.6 on channel stable at /Users/jasonelder/Documents/Flutter/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f468f3366c (5 weeks ago), 2023-07-12 15:19:05 -0700
    • Engine revision cdbeda788a
    • Dart version 3.0.6
    • DevTools version 2.23.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    • Android SDK at /Users/jasonelder/Library/Android/sdk
    • Platform android-33, build-tools 33.0.1
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14E300c
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
[...]

PR #1377, opened 2023-10-04
Adds Standard CSV Mime Type

SoftWyer:
On Android, the CSV mime type from `getMimeTypeFromExtension()` returns "text/comma-separated-values" which is non-standard and doesn't filter CSV files in Google Drive.

This change adds the standard "text/csv" mime type when the csv extension is specified. 

We did think about adding the ability to specify explicit mime types from the API, but this would be a bigger change when this only seems to be an Android issue.

[see standards docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
[see Android code](https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java#439)


miguelpruivo:
Hi, thank you for your PR but the problem here is that iOS doesn't support it and may result in some devs assuming that is working on both and then finding that iOS doesn't support this kind of filter and then later report yet another issue.

Every filter should be supported from all OS to avoid unnecessary results.

SoftWyer:
Hi @miguelpruivo, perhaps you've misunderstood the change. Filtering the file selection by CSV is already supported by both Android and iOS. We can see files with these extensions when using file-picker on these devices.

The problem is that Android uses a **non-standard** CSV mime type extension. This means that it can filter on-device CSV files, but cannot filter CSV files using any cloud service (e.g. Google Drive) as these use the standard mime type.  This change just ensures that CSV filtering works on Android as expected regardless of file system. I tangentially mentioned the [CSV problem in this issue](https://github.com/miguelpruivo/flutter_file_picker/issues/1349#issuecomment-1706607602). 

If you'd rather not include this PR, that it OK with us. We needed Android CSV filtering to work consistently in our app so thought we'd give those changes back to the project 😃

SoftWyer:
Latest changes merged, should be good to go 👍

miguelpruivo (approved):

miguelpruivo (approved):

Issue #1379, opened 2023-10-06
Android: image picker and file picker apps with launchMode=singleInstance doesn't work (more specifically Samsung devices)


melWiss:
**Describe the bug**
the plugin was working just great but when we needed to make our app's launchMode=singleInstance, the picker doesn't work

**Platform**

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
Android 11 and Android 13

**Details to reproduce the issue**
create android app with launchMode=singleInstance and try to pick an image or custom file and it will always return null

**Error Log**
Please, post the _full_ console log of your issue, if applicable.

**Screenshots and/or video**
If applicable, add screenshots or video to help explain your problem.

**Flutter Version details**
Flutter 3.10.6

**Additional context**
None

rameshniyo:
I am also facing this in the samung and other devices

rameshniyo:
Please  include this in the latest version.Lots of users facing this issue

TheManuz:
I'm facing the same issue, I've had to change my `android:launchMode` to `singleTask` instead of `singleInstance` as a workaround.

melWiss:
but in some cases you need your app to be singleInstance, that's why i filed the bug.
I'll update the PR containing the fix.

TheManuz:
> but in some cases you need your app to be singleInstance, that's why i filed the bug. I'll update the PR containing the fix.

Yes, that's my case, I've preferred to change the launchMode just to have a working filepicker.

I was just sharing my workaround, but that's exactly that: a workaround.

petrnymsa:
Issue is not solved. Still occurss on Samsung devices, Android 14. 

In our app we need launchMode: sigleInstance.

TheManuz:
I still have this issue too

rivella50:
Same here, and the corresponding PR has just been closed 3 days ago...

PR #1336, opened 2023-07-30
fix #1335: fix intent flag from ACTION_GET_CONTENT to ACTION_OPEN_DOCUMENT

muttsu-623:
fix #1335: fix intent flag from ACTION_GET_CONTENT to ACTION_OPEN_DOCUMENT

PR #1380, opened 2023-10-06
Fix file & image picking with android apps that has launchMode=singleInstance

melWiss:
this pr resolves this issue: https://github.com/miguelpruivo/flutter_file_picker/issues/1379


miguelpruivo:
Bump the minor version instead of the patch version please.

melWiss:
done

melWiss:
@miguelpruivo can you take a look on this? we need it

rameshniyo (approved):
This looks fine

melWiss:
Thanks buddy

Musta-Pollo:
Legit! Thanks a lot :+1: !

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -232,8 +240,9 @@ private void startFileExplorer() {
         } else {
             if (type.equals("image/*")) {
                 intent = new Intent(Intent.ACTION_PICK, android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
+                intentChooser = Intent.createChooser(intent, this.imagePickerTitle != null ? this.imagePickerTitle : "Select a photo");
             } else {
-                intent = new Intent(Intent.ACTION_GET_CONTENT);
+                intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
```
dmitrio95, line 245: It seems this change breaks some older Android versions. For me with this change the file picker doesn't work on Android 9, but reverting this change or setting the intent type depending on Android version (like [here](https://github.com/dmitrio95/flutter_file_picker/commit/e7dcbff0e2240226fce69c468a6f23dacfe1fc92)) fixes this. I am not sure though which versions should use which intent type and why.

dmitrio95: It seems wrapping all intents with `createChooser` works with `launchMode="singleInstance"` regardless of the version of Android even with the old `ACTION_GET_CONTENT` intent type: [example](https://github.com/dmitrio95/flutter_file_picker/commit/cbdbf52a703082ad58b7ccd3e51c440b3eb05ab5). It will probably work with other launch modes too, but according to the description [here](https://developer.android.com/reference/android/content/Intent#ACTION_CHOOSER) it may force the system to show an extra activity chooser dialog if there are several possible activities for choosing a file. So perhaps this wrapping should better be somehow limited only to `singleInstance` applications.

melWiss:
Hello, sorry I've been away for some period, does this  issue still exist?

sneurlax:
#1336 works for us, so we have been using https://github.com/muttsu-623/flutter_file_picker

But it has issues on Windows, so we had to make https://github.com/cypherstack/flutter_file_picker which is 8.0.3 + muttsu-623's single change: https://github.com/muttsu-623/flutter_file_picker/commit/f0930d9fa79d347b2a0e25a7de4f5a4a88a9a907

syleishere:
I am having same problem on google TV streamer since it is android 14 permanently, can someone please merge
https://github.com/cypherstack/flutter_file_picker

PR #1381, opened 2023-10-07
#970 Support setting an initial directory in getDirectoryPath() on Windows 

andreaszapf:
Hi, thanks for developing and maintaining this plugin!

Here is a fix for issue #970 (getDirectoryPath() would ignore its initialDirectory parameter on Windows). I hope you'll find it helpful.


miguelpruivo (changes requested):
Please rebased and update the versions.

miguelpruivo:
Thanks.

Issue #1343, opened 2023-08-07
PlatformException on web


Mobiwoom:
I'm using  file_picker: ^5.3.3

**Describe the bug**
When trying to call await FilePicker.platform.pickFiles() even with type: FileType.any , I get no trouble on dev with flutter, but when going live on production server , now I get PlatformException error only on WEB
 
my code is simple : 
try {
      result = await FilePicker.platform.pickFiles(type: FileType.any);
    } on PlatformException catch (e) {
      print('Unsupported operation' + e.toString());
    } catch (e) {
      print(e.toString());
    }

**Error Log**
Platform._operatingSystem
Nothing happens when trying to pick files
works well on Android and IOS 

anyone found this isse too ? 
thanks

Ai-Kiwi:
I've had the same issue

Ai-Kiwi:
issue for me appears on final web build however if you run it as debug or release issue does not appear.

ElsimatAmir:
still have this issue

GiYeongUM:
Try `flutter clean` and rebuild project.
I hope this will solve the problem

navaronbracke:
There is a workaround using `kIsWeb` and `FilePickerWeb`. 

That will not be necessary once I am done with my refactor, though.

yogithesymbian:
> r relea

oh just use 
```
await FilePickerWeb.platform.pickFiles();
```
instead of
```
await FilePicker.platform.pickFiles();
```

ChaseGuru:
The documentation says "You are good to go as long as you are on Flutter 2.0 or above." for web.  The entire pub.dev page doesn't say anything about FilePickerWeb.  The repo example file also just uses a kIsWeb in one spot.  So clearly the documentation needs to be updated.

jan-siroky:
> > r relea
> 
> oh just use
> 
> ```
> await FilePickerWeb.platform.pickFiles();
> ```
> 
> instead of
> 
> ```
> await FilePicker.platform.pickFiles();
> ```

@yogithesymbian This works but then I can't build app on other platforms than Web, any thoughts?

@miguelpruivo Thank you for your hard work. Should we expect this plugin to be updated to work with newest Flutter 3.19 and consequent versions which are gonna drop use of dart:html because of WASM?

navaronbracke:
@jan-siroky We can indeed start on migrating to `package:web` (I have experience in this area, should not be too hard to do). 

[...]

PR #1383, opened 2023-10-08
Bump min iOS version to 11

### LINKED ISSUE #1382: [Tech Debt] Bump file_picker minimum iOS version to 11.0
Currently the `file_picker` plugin uses iOS 8 as the minimum supported iOS version.

On top of that it specifies the following SDK constraint:

```yaml
environment:
  sdk: ">=2.17.0 <4.0.0"
  flutter: ">=1.10.0"
```

There are a few issues with this current constraint:

- Dart 2.17.0 shipped with Flutter `3.0.0`, so the minimum Flutter version [is lower than what is actually possible](https://docs.flutter.dev/release/archive?tab=macos)
- The stable release of Flutter only supports [iOS 11.0 and higher](https://docs.flutter.dev/reference/supported-platforms)
- If we bump the minimum iOS version to 11.0, we get a warning in Swift:
  - `'documentPicker:didPickDocumentAtURL:'` which is a deprecated method, can be removed
  
Also, I see a warning about the use of a pointer type:

```
/ios/Classes/FilePickerPlugin.m:487:25 Incompatible pointer types sending 'NSNull * _Nonnull' to parameter of type 'NSURL * _Nonnull'
```

### Proposal

- We should fix the version constraint for Flutter & Dart. Per the policy in `flutter/packages`, they support the last 3 stable releases (so the current one + 2 releases before that)
- We should bump the minimum iOS version to 11.0, as Flutter supports at least iOS 11.0
- We should remove the deprecated method and fix the warning for the pointer type usage

navaronbracke:
This PR bumps the minimum iOS version to version 11.0, which is the minimum required by the current stable releases.
Flutter 3.0 was the last release to support iOS 10.

I also fixed two warnings in the plugin:
- a deprecated method for iOS 10 that could be removed
- a warning about passing `null` to an array of `NSURL`, which did not allow null

Fixes #1382


analysis_options.yaml
```diff
@@ -35,8 +35,7 @@ linter:
     - diagnostic_describe_all_properties
     - empty_statements
     - hash_and_equals
-    - iterable_contains_unrelated_type
-    - list_remove_unrelated_type
+    - collection_methods_unrelated_type
```
navaronbracke, line 38: The two lists are deprecated and are replaced by this new one

example/ios/Flutter/Flutter.podspec
```diff
@@ -1,17 +1,17 @@
 #
-# NOTE: This podspec is NOT to be published. It is only used as a local source!
-#       This is a generated file; do not edit or check into version control.
+# This podspec is NOT to be published. It is only used as a local source!
```
navaronbracke, line 2: These changes were generated by Flutter's tooling.

example/ios/Runner/Info.plist
```diff
@@ -50,5 +50,9 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
```
navaronbracke, line 53: This was generated by Flutter's tooling.

lib/src/file_picker_io.dart
```diff
@@ -73,7 +72,7 @@ class FilePickerIO extends FilePicker {
     bool? withData,
     bool? withReadStream,
   ) async {
-    final String type = describeEnum(fileType);
+    final String type = fileType.name;
```
navaronbracke, line 75: `decribeEnum()` has been deprecated for a while, in favor of the `name` property on enums themselves.

ios/Classes/FilePickerPlugin.m
```diff
@@ -484,7 +460,7 @@ -(void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPicke
     __block NSError * blockError;
     
     for (NSInteger index = 0; index < results.count; ++index) {
-        [urls addObject:[NSNull null]];
+        [urls addObject:[NSURL URLWithString:@""]];
```
navaronbracke, line 463: The warning came from the fact that the `urls` array did not accept null. Since the first element in the array is replaced, I just added an empty NSURL. (and rather than using `new` I went with `URLWithString`, as the former also generates a warning)

ios/Classes/FilePickerPlugin.m
```diff
@@ -173,12 +173,7 @@ - (void)resolvePickDocumentWithMultiPick:(BOOL)allowsMultipleSelection pickDirec
         return;
     }
     
-    if (@available(iOS 11.0, *)) {
-        self.documentPickerController.allowsMultipleSelection = allowsMultipleSelection;
-    } else if(allowsMultipleSelection) {
-        Log(@"Multiple file selection is only supported on iOS 11 and above. Single selection will be used.");
-    }
-    
+    self.documentPickerController.allowsMultipleSelection = allowsMultipleSelection;
```
navaronbracke, line 176: Removing availability checks for iOS 11.0, as that is now the minimum

example/ios/Flutter/Flutter.podspec
```diff
@@ -1,17 +1,17 @@
 #
```
navaronbracke, line 1: This is a generated file and should not be checked in, will fix this as well.

example/ios/Flutter/Flutter.podspec
navaronbracke, line 1: This file is autogenerated, so should not be checked in. It was the only one which is autogenerated, but checked in, so I removed it here.

miguelruivo:
Thank you, could you bump the changelog and also the pubspec version — I'd say that if this could _potentially_ introduce breaking changes (specially since it plays with version constraints) to bump the minor (or major?) version.

miguelpruivo (changes requested):
Thank you, could you bump the changelog and also the pubspec version — I'd say that if this could potentially introduce breaking changes (specially since it plays with version constraints) to bump the minor (or major?) version.

navaronbracke:
A breaking change would mean we bump the major version, per semantic versioning.

Will bump to version 6.0.0 then

miguelpruivo:
> A breaking change would mean we bump the major version, per semantic versioning.
> 
> Will bump to version 6.0.0 then

Correct. The question was more wether or not it _could_ actually break or could be just some constraints that wouldn't probably break anything. But, as per your reply, major seems correct to me. Thank you!

miguelpruivo:
@navaronbracke because of your recent contributions to this plugin I'd like to ask you if you're interested into become a contributor since lately I couldn't keep trace of a couple of issues that might appear (updates and related stuff) on the mobile/web side. Typically @philenius is responsible for the desktop implementations/maintenance.

Let me know if you're interested so I can assign you with the right permissions.

Thank you.

navaronbracke:
@miguelpruivo Yes I would be interested in helping out. I think I'll start on the plugin_platform_interface work that I had scheduled to do, which moves the selection of the file picker implementation to the pubspec. This is the fix for https://github.com/miguelpruivo/flutter_file_picker/issues/1343

After that is fixed, I'll try working my way through the backlog. I do not have a Linux machine available (nor do I have win32 or gtk experience), but I should be able to help out for other platforms.

miguelpruivo:
> @miguelpruivo Yes I would be interested in helping out. I think I'll start on the plugin_platform_interface work that I had scheduled to do, which moves the selection of the file picker implementation to the pubspec. This is the fix for #1343
> 
> After that is fixed, I'll try working my way through the backlog. I do not have a Linux machine available (nor do I have win32 or gtk experience), but I should be able to help out for other platforms.

Hi, thank you I really appreciate that, because due to this plugins popularity and being myself the only maintainer along with @philenius for the desktop side makes it harder to trace/handle all the PRs lately. A lot of them are just (as expected) doubts that are in wiki or stuff that is a no-op, but there are a couple of relevant issues that need attention.

I'd say to focus only in the the mobile/web issues and let alone the desktop (linux/mac/win) for @philenius and assign to him, as it is right now, unless is some immediate thing that we can take care of.

There is an action to automatically publish a new versions to the pub.dev upon `master` commit, so have that in mind.

I'll provide you with the right permissions now. Thank you once again.

PR #1392, opened 2023-10-23
Feature: keep picking order on web

test0terter0n:
Currently the order of the imported files in web is not respected. It seems like files are ordered by filesize. 
To have the feature to keep the picking order, the files are read synchronously.  
With `readSequential` order of files will be kept in web. The default value is false.


miguelpruivo:
Hi, update the minor version on the pubspec as well. Thanks.

test0terter0n:
@miguelpruivo  as you which ;)

PR #1396, opened 2023-10-26
saveFile method implemented for android platform

vinxxe:
I implemented a saveFile method for the android platform because I needed it.
I don't know if it is interesting, I think that you already know how to do it but maybe violates some android security rule and that's the reason why you didn't add it.
If this is the case just ignore it.

The only way I found to implement the method is by removing the file that the saveFile dialog creates because that file cannot be written by the calling thread. By removing the 0 length created file and returning the path, the calling thread is able to open the file, provided that the file is in an accessible folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 6.1.0
+### Android
+- saveFile() method implemented for the android platform
 ## 6.0.0
 Update minimum Flutter version to 3.7.0.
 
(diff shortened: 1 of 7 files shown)


Empitrix:
App crashes in android with:
Error:
```text
java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=34284, result=-1, data=Intent { dat=content://com.android.providers.downloads.documents/... flg=0x43 }} to activity {com.example.save_on_android/com.example.save_on_android.MainActivity}: java.lang.NullPointerException: Attempt to get length of null array
```

daniJimen:
@navaronbracke I propose closing this outdated PR; I think it's no longer relevant, since file saving in Android is already in master.

PR #1411, opened 2023-11-20
Support initialDirectory of getDirectoryPath in Android

nnnlog:
For example, given `/storage/emulated/0/DCIM` as initialDirectory, it converts path to `primary:/DCIM` internally. Using `EXTRA_INITIAL_URI` and `com.android.externalstorage.documents`, can set initialDirectory of file explorer. It works for android.

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
@@ -244,6 +244,21 @@ public class FileUtils {
     }
 
     @Nullable
+    @SuppressWarnings("deprecation")
+    public static String getTreeUriFromFullPath(@Nullable String fullPath, Context con) {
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
```diff
@@ -243,6 +243,21 @@ public static String getFullPathFromTreeUri(@Nullable final Uri treeUri, Context
         }
     }
 
+    @Nullable
+    @SuppressWarnings("deprecation")
```
miguelpruivo, line 247: Do we really need to use deprecated SDKs? I'd like to merge it but I'm afraid I don't want to introduce more deprecated SDKs.

miguelpruivo (changes requested):

miguelpruivo:
@nnnlog any news on this?

PR #1424, opened 2023-12-20
add namespace to build.gradle

pishguy:
i added `namespace` to `build.gradle` to support compatibility with AGP <4.2.


navaronbracke:
This is already set up a bit lower in the file.

PR #1430, opened 2024-01-08
Minor fixes

tvolkert:
* Allow m4v files in FileType.media filter
* throw if caller tries to use allowedExtensiosn with any file type other than custom

### DIFF
--- a/lib/src/file_picker_macos.dart
+++ b/lib/src/file_picker_macos.dart
@@ -108,8 +108,12 @@ class FilePickerMacOS extends FilePicker {
 
   String fileTypeToFileFilter(FileType type, List<String>? allowedExtensions) {
     if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
-      throw Exception(
-        'You are setting a type [$type]. Custom extension filters are only allowed with FileType.custom, please change it or remove filters.');
+      throw ArgumentError.value(
+         allowedExtensions,
+         'allowedExtensions',
+         'Custom extension filters are only allowed with FileType.custom. '
+         'Remove the extension filter or change the FileType to FileType.custom.'
+      );
     }
     switch (type) {
       case FileType.any:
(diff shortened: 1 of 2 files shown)


lib/src/file_picker_macos.dart
```diff
@@ -107,6 +107,10 @@ class FilePickerMacOS extends FilePicker {
   }
 
   String fileTypeToFileFilter(FileType type, List<String>? allowedExtensions) {
+    if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
+      throw Exception(
```
navaronbracke, line 111: Could you turn this into an `ArgumentError` ? That would make the intent a bit clearer IMO.

And we can tweak the message, to remove the "please" grammar.

```
throw ArgumentError.value(
   allowedExtensions,
   'allowedExtensions',
   'Custom extension filters are only allowed with FileType.custom. '
   'Remove the extension filter or change the FileType to FileType.custom.'
);
```

navaronbracke (changes requested):

lib/src/file_picker_macos.dart
tvolkert: Done - thanks for the review!

navaronbracke:
We have one test that is failing, because the expectation needs to be updated.

```
flutter_file_picker/flutter_file_picker/test/file_picker_macos_test.dart: fileTypeToFileFilter() should return the file filter (failed)
  Expected: '"avi", "flv", "mkv", "mov", "mp4", "mpeg", "webm", "wmv", "bmp", "gif", "jpeg", "jpg", "png"'
    Actual: '"avi", "flv", "m4v", "mkv", "mov", "mp4", "mpeg", "webm", "wmv", "bmp", "gif", "jpeg", "jpg", "png"'
     Which: is different.
            Expected: ...  "flv", "mkv", "mov" ...
              Actual: ...  "flv", "m4v", "mkv" ...
                                    ^
             Differ at offset [16](https://github.com/miguelpruivo/flutter_file_picker/actions/runs/7456317251/job/20290515807?pr=1430#step:8:17)
```

miguelpruivo (approved):
LGTM. Thanks.

PR #1443, opened 2024-02-06
fix exception message

gcb:
Maybe I am not too bright, but wasted a lot of time thinking a null check would correctly work around being on a web renderer. :sweat_smile:


miguelpruivo (approved):
Thanks.

PR #1452, opened 2024-03-03
Save file to mobile platforms.

Samoy:
User can save file to mobile platforms with bytes. Desktop platforms remain unchanged.

PR #1457, opened 2024-03-09
fix #1412: Picking a folder in iOS causes the original folder to be deleted.

### LINKED ISSUE #1412: BUG: Picking a folder in iOS causes the original folder to be deleted.
**Describe the bug**
This is a serious bug where users' files are deleted without them noticing. If use file_picker version 5.3.3 or higher, when an ios user picks a directory, the original directory will be deleted (actually, moved to the app's sandbox tmp directory, and the user cannot access them), and file_picker returns a path in tmp directory.
In 5.3.2 and earlier versions, file_picker will return the real absolute path of the original directory.
This BUG looks to be caused by this pull: https://github.com/miguelpruivo/flutter_file_picker/pull/1311 (I'm not an ios programmer so I don't know how to fix this bug...), which moves the file or directory pointed to by the original url into the tmp directory; please restore the previous behavior (5.3.2, getting the original absolute path, which is the best, I think ?), or copy the directory to tmp (rather than moving it).

**Platform**

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
iOS 16.5

**How are you picking?**
```dart
if (Platform.isIOS) {
  var selection = await FilePicker.platform.getDirectoryPath();
  result = selection;
}
```

1. open a directory picker.
2. select a directory.
3. the selected directory was deleted (moved to the `<application-sandbox-directory>/tmp` directory).
4. file_picker returnd the tmp path.

**Error Log**
None.

**Screenshots and/or video**
None.

**Flutter Version details**
Flutter 3.16, no probrlm.

**Additional context**
None.

twinkles-twinstar:
If use file_picker version 5.3.3 or higher, when an ios user picks a directory, the original directory will be deleted (actually, moved to the app's sandbox tmp directory, and the user cannot access them), and file_picker returns a path in tmp directory.

This PR fixes this BUG. file_picker will return the absolute path to the folder, just like before v5.3.3.

### DIFF
--- a/ios/Classes/FilePickerPlugin.m
+++ b/ios/Classes/FilePickerPlugin.m
@@ -349,26 +349,32 @@ didPickDocumentsAtURLs:(NSArray<NSURL *> *)urls{
     if(_result == nil) {
         return;
     }
-    NSMutableArray<NSURL *> *newUrls = [NSMutableArray new];
-    for (NSURL *url in urls) {
-        // Create file URL to temporary folder
-        NSURL *tempURL = [NSURL fileURLWithPath:NSTemporaryDirectory()];
-        // Append filename (name+extension) to URL
-        tempURL = [tempURL URLByAppendingPathComponent:url.lastPathComponent];
-        NSError *error;
-        // If file with same name exists remove it (replace file with new one)
-        if ([[NSFileManager defaultManager] fileExistsAtPath:tempURL.path]) {
-            [[NSFileManager defaultManager] removeItemAtPath:tempURL.path error:&error];
+    NSMutableArray<NSURL *> *newUrls;
+    if(controller.documentPickerMode == UIDocumentPickerModeOpen) {
+        newUrls = urls;
+    }
+    if(controller.documentPickerMode == UIDocumentPickerModeImport) {
+        newUrls = [NSMutableArray new];
+        for (NSURL *url in urls) {
+            // Create file URL to temporary folder
+            NSURL *tempURL = [NSURL fileURLWithPath:NSTemporaryDirectory()];
+            // Append filename (name+extension) to URL
+            tempURL = [tempURL URLByAppendingPathComponent:url.lastPathComponent];
+            NSError *error;
+            // If file with same name exists remove it (replace file with new one)
+            if ([[NSFileManager defaultManager] fileExistsAtPath:tempURL.path]) {
+                [[NSFileManager defaultManager] removeItemAtPath:tempURL.path error:&error];
+                if (error) {
+                    NSLog(@"%@", error.localizedDescription);
+                }
+            }
+            // Move file from app_id-Inbox to tmp/filename
+            [[NSFileManager defaultManager] moveItemAtPath:url.path toPath:tempURL.path error:&error];
             if (error) {
                 NSLog(@"%@", error.localizedDescription);
+            } else {
+                [newUrls addObject:tempURL];
             }
-        }
-        // Move file from app_id-Inbox to tmp/filename
-        [[NSFileManager defaultManager] moveItemAtPath:url.path toPath:tempURL.path error:&error];
-        if (error) {
-            NSLog(@"%@", error.localizedDescription);
-        } else {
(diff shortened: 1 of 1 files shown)


ios/Classes/FilePickerPlugin.m
```diff
@@ -349,27 +349,33 @@ - (void)documentPicker:(UIDocumentPickerViewController *)controller
...
-        // Create file URL to temporary folder
-        NSURL *tempURL = [NSURL fileURLWithPath:NSTemporaryDirectory()];
-        // Append filename (name+extension) to URL
-        tempURL = [tempURL URLByAppendingPathComponent:url.lastPathComponent];
-        NSError *error;
-        // If file with same name exists remove it (replace file with new one)
-        if ([[NSFileManager defaultManager] fileExistsAtPath:tempURL.path]) {
-            [[NSFileManager defaultManager] removeItemAtPath:tempURL.path error:&error];
+    NSMutableArray<NSURL *> *newUrls;
+    if(controller.documentPickerMode == UIDocumentPickerModeOpen) {
```
navaronbracke, line 353: The `documentPickerMode` is deprecated. Could you rewrite this to only use the deprecated API up to iOS 14 ?

See https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller#1658475

navaronbracke (changes requested):

ios/Classes/FilePickerPlugin.m
twinkles-twinstar: Sorry, I don't have a mac device now, so I can't rewrite and test it. 
In addition, I noticed that the deprecated API was also used in the [previous code](https://github.com/miguelpruivo/flutter_file_picker/blob/dff1dde7277b03ac286a3957503731c23a2e4d53/ios/Classes/FilePickerPlugin.m#L169). Maybe this also needs to be migrated to the new API, but I am not familiar with objc.
BTW, on my own ios device, when using the `init(forOpeningContentTypes:)` swift api, I can directly get the absolute path of file/directory. Why doesn't file_picker use this api, but makes a cache file?

navaronbracke: I do not now why file_picker uses a cache for that, maybe @miguelpruivo knows?

miguelpruivo: > I do not now why file_picker uses a cache for that, maybe @miguelpruivo knows?

There are 2 approaches when opening descriptors for files within app's container
1. Opening a resource directly to the file through URI/UTI or;
2. Caching a copy of the file so it can be manipulated individually within the app and without any constraint regarding the original file;

2 is typically the preferred option since most of the apps use the files for either manipulating (editing) or, even if it's just for sending somewhere, it doesn't have any consequence. If I recall, this was the original approach from the Flutter's image_picker so I followed the same path.

twinkles-twinstar: @navaronbracke @miguelpruivo 
Please pay attention to #1412 . This bug is very serious on iOS and will endanger user data security.
It has been half a year since I raised the issue, but this dangerous bug still exists and has not been fixed.
This PR is made based on the current implementation, I know that `UIDocumentPickerMode` is a deprecated api, but the code in this library always uses this deprecated api when creating `UIDocumentPickerViewController`.
In this case, why is this PR banned because I use the same deprecated API? (In fact, another PR  #1452 that was approved also used the deprecated `UIDocumentPickerMode` api).
I think a more reasonable solution is to first pass this PR to solve the BUG instead of banning this PR because of an deprecated api that already exists in the library.
Please allow me to repeat it again: this bug is very serious, it will delete files without the user knowing! Please don't keep this bug alive for longer.

navaronbracke:
@twinkles-twinstar Can you a) fix the merge conflict and b) update the version and changelog? The version can bump the patch version.

twinkles-twinstar:
> @twinkles-twinstar Can you a) fix the merge conflict and b) update the version and changelog? The version can bump the patch version.

Completed, please review.

navaronbracke (approved):

Issue #1464, opened 2024-03-14
2 files same name, different folder, returns same path for both files


rsanjuan87:
**Platform**

-  Android

**Platform OS version**
Android 13 and others

**How are you picking?**
pdf

**Details to reproduce the issue**
Set 2 files with the same file name in different folders, will replace first one with the second
When return the file path  both times the same path

miguelpruivo:
Fixed by you and already published. Thanks.

PR #1466, opened 2024-03-14
add store in folder by millis

rsanjuan87:
prevent be replaced by same file name
fix to https://github.com/miguelpruivo/flutter_file_picker/issues/1464


miguelpruivo:
At first I believe I exactly wanted to replace to prevent multiple repeated picks, but now that I think and since we're caching, it may not be bad idea to always cache a new file for the app's content storage. Thank you.

miguelpruivo (approved):

PR #1369, opened 2023-09-12
Fix issue 1367

GabrielAraujo:
[Issue 1367](https://github.com/miguelpruivo/flutter_file_picker/issues/1367)

On iOS 17 the UTI type public.item is not properly working for live photos that are syncing with iCloud (download or upload).
Using media specific UTI types seemed to have fixed the problem but still any thoughts on the approach are welcome so we can adjust on how we handle the issue.

Issue #1367, opened 2023-09-07
[iOS] Picking live photos not properly working on iOS 17.


GabrielAraujo:
Recently taken or older from iCloud `live photos`, are failing to load the file representation using `public.item` UTI type on iOS 17 Beta.

**Platform**

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
iOS 17

**How are you picking?**
```
final pickerResult = await FilePicker.platform.pickFiles(
          type: MediaType.image,
          allowMultiple: true,
          allowCompression: true,
        );
```

**Details to reproduce the issue**
1. Take live photos
2. Try to select those from your app
3. Process fails and exception is returned

**Error Log**
```
***** Could not load the picked given file: Error Domain=NSItemProviderErrorDomain Code=-1000 "Cannot load representation of type public.item" UserInfo={NSLocalizedDescription=Cannot load representation of type public.item, NSUnderlyingError=0x28282de30 {Error Domain=NSCocoaErrorDomain Code=4097 "Couldn’t communicate with a helper application." UserInfo={NSUnderlyingError=0x2829c7540 {Error Domain=NSCocoaErrorDomain Code=4097 "connection from pid 1220 on anonymousListener or serviceListener" UserInfo={NSDebugDescription=connection from pid 1220 on anonymousListener or serviceListener}}}}}
```

**Flutter Version details**
```
[!] Flutter (Channel unknown, 3.7.8, on macOS 13.5.1 22G90 darwin-arm64, locale en-BR)
    ! Flutter version 3.7.8 on channel unknown at /Users/gabrielaraujo/fvm/versions/3.7.8
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
    ! Unknown upstream repository.
      Reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
    • Framework revision 90c64ed42b (6 months ago), 2023-03-21 11:27:08 -0500
    • Engine revision 9aa7816315
    • Dart version 2.19.5
    • DevTools version 2.20.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[!] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    • Android SDK at /Users/gabrielaraujo/Library/Android/sdk
[...]

PR #1475, opened 2024-03-25
Fix Issue 1367

GastonTestay:
We rebased the PR #1369 fixing the issue [1367](https://github.com/miguelpruivo/flutter_file_picker/issues/1367)

All credits should go to @GabrielAraujo

> On iOS 17 the UTI type public.item is not properly working for live photos that are syncing with iCloud (download or upload).
> Using media specific UTI types seemed to have fixed the problem but still any thoughts on the approach are welcome so we can adjust on how we handle the issue.


miguelpruivo:
Dup of #1369

Issue #1480, opened 2024-04-06
Migrate to `package:web` to support WASM


IchordeDionysos:
The short version:
  - Migrate usage of dart:html to package:web
  - Migrate usage of dart:js, dart:js_util, and package:js to dart:js_interop

This will also compiling to WASM engine instead of just JavaScript on web.

Full migration guide: https://dart.dev/interop/js-interop/package-web

amrgetment:
My PR to support WASM

https://github.com/miguelpruivo/flutter_file_picker/pull/1481

IchordeDionysos:
Still relevant

navaronbracke:
This will be included in the next release.

PR #1481, opened 2024-04-08
WASM support for Flutter web

amrgetment:
Adding WASM support to close
https://github.com/miguelpruivo/flutter_file_picker/issues/1480

Test by Flutter Beta channel or master channel and build by the following command
`flutter build web --wasm`


pubspec.yaml
```diff
@@ -17,6 +17,8 @@ dependencies:
   path: ^1.8.2
   win32: ^5.1.1
   cross_file: ^0.3.3+7
+  js: ^0.7.1
```
IchordeDionysos, line 20: Is that intended?
It's not used anymore and doesn't work with WASM anyways :)

IchordeDionysos (commented):
Apart from the import it looks good!

Thanks for the PR 😍

pubspec.yaml
amrgetment: it is a mistake, I removed it I don’t use it anyway
thanks

IchordeDionysos (approved):

amrgetment:
@IchordeDionysos You have to be a team member of this project to pass CI approval requirements 😁
Thanks for approval 😍🎯

IchordeDionysos:
Yeah true :D But given I've looked over it, I figured it can't hurt :P

lib/_internal/file_picker_web.dart
```diff
@@ -58,7 +60,9 @@ class FilePickerWeb extends FilePicker {
         Completer<List<PlatformFile>?>();
 
     String accept = _fileType(type, allowedExtensions);
-    InputElement uploadInput = FileUploadInputElement() as InputElement;
+    HTMLInputElement uploadInput =
```
navaronbracke, line 63:
```suggestion
    HTMLInputElement uploadInput = HTMLInputElement();
```

The constructor calls document.createElement under the hood.

lib/_internal/file_picker_web.dart
```diff
@@ -101,7 +105,12 @@ class FilePickerWeb extends FilePicker {
         }
       }
 
-      for (File file in files) {
+      for (var i = 0; i < files.length; i++) {
```
navaronbracke, line 108:
```suggestion
      for (int i = 0; i < files.length; i++) {
```

lib/_internal/file_picker_web.dart
```diff
@@ -144,14 +155,18 @@ class FilePickerWeb extends FilePicker {
...
+    uploadInput.addEventListener('change', changeEventListener.toJS);
+    uploadInput.addEventListener('cancel', cancelledEventListener.toJS);
 
     // Listen focus event for cancelled
-    window.addEventListener('focus', cancelledEventListener);
+    window.addEventListener('focus', cancelledEventListener.toJS);
 
     //Add input element to the page body
-    _target.children.clear();
+    var firstChild = _target.firstChild;
```
navaronbracke, line 165: I'd prefer to use the real type instead of `var` here.

navaronbracke (changes requested):
Some minor remarks

lib/_internal/file_picker_web.dart
amrgetment: refactored

lib/_internal/file_picker_web.dart
amrgetment: refactored

amrgetment:
I added the change requests

navaronbracke (approved):

Issue #1486, opened 2024-04-09
Compilation fails with Pod::PICKER_DOCUMENT = false 


Aulig:
**Describe the bug**
Set Pod::PICKER_DOCUMENT = false in the Podfile as documented in the Wiki. Compilation fails on version 7.0.0 and above (it was working in 6.x.x)

**Platform**

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Desktop

**Error Log**
Semantic Issue (Xcode): Property 'documentPickerController' not found on object
of type 'FilePickerPlugin *'; did you mean 'audioPickerController'?
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:189:9

Semantic Issue (Xcode): Property 'documentPickerController' not found on object
of type 'FilePickerPlugin *'; did you mean 'audioPickerController'?
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:190:9

Semantic Issue (Xcode): Property 'documentPickerController' not found on object
of type 'FilePickerPlugin *'; did you mean 'audioPickerController'?
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:191:9

Semantic Issue (Xcode): Property 'documentPickerController' not found on object
of type 'FilePickerPlugin *'; did you mean 'audioPickerController'?
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:194:17

Semantic Issue (Xcode): Property 'directoryURL' not found on object of type
'MPMediaPickerController *'
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:194:42

Semantic Issue (Xcode): Property 'documentPickerController' not found on object
of type 'FilePickerPlugin *'; did you mean 'audioPickerController'?
/Users/administrator/.pub-cache/hosted/pub.dev/file_picker-8.0.0+1/ios/Classes/F
ilePickerPlugin.m:197:68

**Additional context**
The problem is that documentPickerController doesn't get defined due to the flag in the Podfile, but not all references are properly removed too.

aggeloskoutanis:
+1

[...]

PR #1494, opened 2024-04-16
Compile out broken references if PICKER_DOCUMENT is set to false

Aulig:
Fixes the issue where compilation fails with Pod::PICKER_DOCUMENT = false 
See issue https://github.com/miguelpruivo/flutter_file_picker/issues/1486

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.0.1
+### iOS
+Fixes an issue preventing compilation on iOS when using Pod::PICKER_DOCUMENT = false.
+
 ## 8.0.0+1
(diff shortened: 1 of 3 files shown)


navaronbracke (approved):

navaronbracke:
Thanks for catching this! We indeed missed an ifdef.

Aulig:
> Thanks for catching this! We indeed missed an ifdef.

No problem - thanks a lot for reviewing!

navaronbracke:
Sadly, I'm not part of the publisher on pub, so I cannot give a guarantee on when we will be able to release this fix.

Aulig:
grafik

It seems to already be out :)

PR #1496, opened 2024-04-20
Fix: TypeError: Instance of 'NativeByteBuffer': type 'NativeByteBuffer' is not a subtype of type 'List'.

vito-go:
When built in the html web-renderer, an error may occur:
 TypeError: Instance of 'NativeByteBuffer': type 'NativeByteBuffer' is not a subtype of type 'List<int>'.

 Therefore, we need to make a type discrimination of the reader.result.

### DIFF
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -208,6 +208,15 @@ class FilePickerWeb extends FilePicker {
       final blob = file.slice(start, end);
       reader.readAsArrayBuffer(blob);
       await EventStreamProviders.loadEvent.forTarget(reader).first;
+      if (reader.result is ByteBuffer){
+        // When built in the html web-renderer, an error may occur:
+        // TypeError: Instance of 'NativeByteBuffer': type 'NativeByteBuffer' is not a subtype of type 'List<int>'.
+        // Therefore, we need to make a type discrimination of the reader.result.
+        yield  (reader.result as ByteBuffer).asUint8List();
+        start += _readStreamChunkSize;
+        continue;
+      }
+      
       yield reader.result as List<int>;
       start += _readStreamChunkSize;
     }
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -212,29 +212,18 @@ class FilePickerWeb extends FilePicker {
       if (readerResult == null) {
         continue;
       }
-      if (readerResult is ByteBuffer) {
-        // An error may occur:
-        // TypeError: Instance of 'NativeByteBuffer': type 'NativeByteBuffer' is not a subtype of type 'List<int>'.
-        // Therefore, we need to make a type discrimination of the reader.result.
-        yield (readerResult as ByteBuffer).asUint8List();
+      // TODO: use `isA<JSArrayBuffer>()` when switching to Dart 3.4
+      // Handle the ArrayBuffer type. This maps to a `ByteBuffer` in Dart.
+      if (readerResult.instanceOfString('ArrayBuffer')) {
+        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
         start += _readStreamChunkSize;
         continue;
       }
-      // Defensive programming: maybe the code can not reach here. ByteBuffer can cover most cases.
-      // Handle byte buffers.
-      if (readerResult.instanceOfString('JSArrayBuffer')) {
-        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
-        start += _readStreamChunkSize;
-      }
-      // Handle JS Array type.
+      // TODO: use `isA<JSArray>()` when switching to Dart 3.4
+      // Handle the Array type.
       if (readerResult.instanceOfString('Array')) {
         // Assume this is a List<int>.
         yield (readerResult as JSArray).toDart.cast<int>();
-        start += _readStreamChunkSize;
-      }
-      // Handle JS ArrayBuffer type.
-      if (readerResult.instanceOfString('ArrayBuffer')) {
-        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
         start += _readStreamChunkSize;
       }
     }
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -208,16 +208,33 @@ class FilePickerWeb extends FilePicker {
       final blob = file.slice(start, end);
       reader.readAsArrayBuffer(blob);
       await EventStreamProviders.loadEvent.forTarget(reader).first;
-      if (reader.result is ByteBuffer){
-        // When built in the html web-renderer, an error may occur:
+      final JSAny? readerResult = reader.result;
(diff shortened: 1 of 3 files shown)


lib/_internal/file_picker_web.dart
```diff
@@ -208,6 +208,15 @@ class FilePickerWeb extends FilePicker {
       final blob = file.slice(start, end);
       reader.readAsArrayBuffer(blob);
       await EventStreamProviders.loadEvent.forTarget(reader).first;
+      if (reader.result is ByteBuffer){
```
navaronbracke, line 211: We should use a local variable, so that type promotion works.

According to the package:web documentation, the `FileReader` result is a `JSAny?` type.
https://pub.dev/documentation/web/latest/web/FileReader/result.html

You'll have to handle `null`, `JSArray` and `JSArrayBuffer`.
You can check using `instanceOfString` (and add a TODO to switch to isA<T> once we move to Dart 3.4) before casting the JS value.
See https://dart.dev/interop/js-interop/js-types#compatibility-type-checks-and-casts

```suggestion
     final JSAny? readerResult = reader.result;

      if (readerResult == null) {
        continue;
      }
      
      // Handle byte buffers.
      if (readerResult.instanceOfString('JSArrayBuffer')) {
        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
        start += _readStreamChunkSize;
        
        continue;
      }
      
      // Handle JS Array type.
      if (readerResult.instanceOfString('Array')) {
        // Assume this is a List<int>.
        yield (readerResult as JSArray).toDart.cast<int>();
        start += _readStreamChunkSize;
      }
```

navaronbracke (changes requested):

lib/_internal/file_picker_web.dart
```diff
@@ -208,7 +208,33 @@ class FilePickerWeb extends FilePicker {
       final blob = file.slice(start, end);
       reader.readAsArrayBuffer(blob);
       await EventStreamProviders.loadEvent.forTarget(reader).first;
-      yield reader.result as List<int>;
+      final JSAny? readerResult = reader.result;
+      if (readerResult == null) {
+        continue;
+      }
+      if (readerResult is ByteBuffer) {
```
navaronbracke, line 215: This if-block can be removed, as you now handle it below, with `if (readerResult.instanceOfString('JSArrayBuffer')) {`. This is because the `JSArrayBuffer` maps to a ByteBuffer in Dart.

lib/_internal/file_picker_web.dart
```diff
@@ -208,7 +208,33 @@ class FilePickerWeb extends FilePicker {
...
+        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
+        start += _readStreamChunkSize;
+        continue;
+      }
+      // Handle JS Array type.
+      if (readerResult.instanceOfString('Array')) {
+        // Assume this is a List<int>.
+        yield (readerResult as JSArray).toDart.cast<int>();
+        start += _readStreamChunkSize;
+        continue;
```
navaronbracke, line 234: This `continue;` is redundant.

lib/_internal/file_picker_web.dart
```diff
@@ -208,7 +208,33 @@ class FilePickerWeb extends FilePicker {
...
+      }
+      // Default this is a List<int>
```
navaronbracke, line 236: You already handled the `List<int>` case with `if (readerResult.instanceOfString('Array')) {` on line 230, so this block can be removed.

navaronbracke (changes requested):
Some code is now redundant with the new approach.

lib/_internal/file_picker_web.dart
vito-go: Well, when I debug on my machine, it shows that readerResult.runtimeType is "ByteBuffer" and 
readerResult.instanceOfString('JSArrayBuffer')-> false, readerResult.instanceOfString('Array')-> false.

readerResult

vito-go: Yes, you are right, `readerResult.instanceOfString('ArrayBuffer')` is true.

If without a cast, although the type system can help in promoting the type, the IDE can not. That's to say, `readerResult` does not have a function named `asUint8List`, but `(readerResult as ByteBuffer)` does have.

I don't know use `instanceOfString` can cover all aspects of the readerResult type.
I guess handling the ByteBuffer case directly  is better.

lib/_internal/file_picker_web.dart
vito-go: done

lib/_internal/file_picker_web.dart
navaronbracke: This is still not quite there yet.

The following should do it, though.

A short explanation:
Since we are running on the web, we are always dealing with the types in 
https://github.com/dart-lang/sdk/blob/cf564396d263434c29efe984f1f2220814e5969a/sdk/lib/_internal/js_shared/lib/js_types.dart
as that is a guarantee by the DDC / dart2js compiler.

Since the `FileReader.result` is `JSAny?` we handle null first.
Now we still have a `JSAny` which can be `ArrayBuffer` or `Array` in JS.
Since the `isA<T extends JSAny>()` helper is not yet available, we use the JavaScript `instanceof`, by calling out to `instanceOfString()`

1) `instanceOfString('ArrayBuffer')`
-> (which maps to `NativeByteBuffer` in the types above, which is why you see the `NativeByteBuffer`)
2) `instanceOfString('Array')`
-> here we assume we got a JS Array with integers, instead of an ArrayBuffer

In the future, the `isA<T extends JSAny>()` helper _will help_ with type promotion on the variable, but since we don't have that yet, we have to use `instanceof` to be safe.


```suggestion
  Stream<List<int>> _openFileReadStream(File file) async* {
    final reader = FileReader();

    int start = 0;
    while (start < file.size) {
      final end = start + _readStreamChunkSize > file.size
          ? file.size
          : start + _readStreamChunkSize;
      final blob = file.slice(start, end);
      reader.readAsArrayBuffer(blob);
      await EventStreamProviders.loadEvent.forTarget(reader).first;
      final JSAny? readerResult = reader.result;

      if (readerResult == null) {
        continue;
      }

      // TODO: use `isA<JSArrayBuffer>()` when switching to Dart 3.4
      // Handle the ArrayBuffer type. This maps to a `ByteBuffer` in Dart.
      if (readerResult.instanceOfString('ArrayBuffer')) {
        yield (readerResult as JSArrayBuffer).toDart.asUint8List();
        start += _readStreamChunkSize;

        continue;
      }

      // Handle the Array type.
      if (readerResult.instanceOfString('Array')) {
        // Assume this is a List<int>.
        yield (readerResult as JSArray).toDart.cast<int>();
        start += _readStreamChunkSize;
      }
    }
  }
```

vito-go: Your explanation is very clear. I have applied your suggestions. Thank you! :)

navaronbracke (approved):
LGTM

If you can update the patch version and add a changelog entry, then we can publish this fix.

PR #1510, opened 2024-05-17
web split cancel and focus event, increase focus time

jiajiawei:
[c5b01e072ffa24a7500a38c0e0b927f2e371b4a5](https://github.com/miguelpruivo/flutter_file_picker/commit/c5b01e072ffa24a7500a38c0e0b927f2e371b4a5)

I have two computer tests here I found that one second is still not enough time. So changeEventTriggered is always true. Is there any other way to determine whether to cancel?

One is Win7+Chrome109, and the other is Win10+Chrome122

### DIFF
--- a/lib/_internal/file_picker_web.dart
+++ b/lib/_internal/file_picker_web.dart
@@ -140,12 +140,17 @@ class FilePickerWeb extends FilePicker {
     }
 
     void cancelledEventListener(Event _) {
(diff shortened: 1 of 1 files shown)


miguelpruivo:
Not sure this is the best approach — I guess that's up to each implementation to try ago around that if it happens, instead of delegating it to the plugin.

Issue #1357, opened 2023-08-25
[iOS] File Picker sometimes crashes when you swipe the modal down and call it again


chapsan2001:
**Describe the bug**
When user closes the native bottom sheet by swiping it down, `FilePicker.platform.pickFiles` sometimes does not complete. It leads to PlatformException and crash when you try to call it again. You have to close the sheet quickly and quickly tap the button again. Package version is 5.3.3 which is the latest

Similar issue seems to be fixed earlier (in version 4.1.1) - [Link](https://github.com/miguelpruivo/flutter_file_picker/issues/813#issuecomment-918463077). Seems to be a regression.

**Platform**

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
iOS 16.5

**How are you picking?**
```
import 'package:file_picker/file_picker.dart';
import 'package:flutter/material.dart';

void main() => runApp(const MyApp());

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: ElevatedButton(
            onPressed: FilePicker.platform.pickFiles,
            child: const Text('Pick File'),
          ),
        ),
      ),
    );
  }
}
```

**Error Log**
```
flutter: [MethodChannelFilePicker] Platform exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
[VERBOSE-2:dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:652:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:310:18)
<asynchronous suspension>
#2      MethodChannel.invokeListMethod (package:flutter/src/services/platform_channel.dart:496:35)
<asynchronous suspension>
#3      FilePickerIO._getPath (package:file_picker/src/file_picker_io.dart:92:33)
<asynchronous suspension>



***** FilePicker canceled
* -[FilePickerPlugin documentPickerWasCancelled:] [Line 635]
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
[...]

PR #1529, opened 2024-06-08
Fix 1357

kreativityapps:
Fix https://github.com/miguelpruivo/flutter_file_picker/issues/1357
Added `CustomDocumentPickerViewController`, a subclass of `UIDocumentPickerViewController` that overrides the `viewDidDisappear` method, making sure that `documentPickerWasCancelled` is called when no file is selected, even when the bottom sheet is suddenly swiped down.
In `documentPickerWasCancelled` added additional checks, but the root cause is that `documentPickerWasCancelled` was not called before.


miguelpruivo (changes requested):
Thanks. Please update the changelog accordingly and bump patch version.

miguelpruivo:
Already fixed. Thanks.

Issue #1484, opened 2024-04-09
 Web:Uncaught Unsupported operation: Platform._operatingSystem


mokanus:
**Describe the bug**
The local operation is perfect, and when I use Flutter run -- release, it is also possible. However, when deploying and accessing Flutter build web, errors may occur

**Platform**
- [x] Web

**Platform OS version**
MacOS：14.2.1 (23C71)
Chrome: 123.0.6312.87（正式版本） (x86_64)

**How are you picking?**
I just trigger by one button.Just like:
![image](https://github.com/miguelpruivo/flutter_file_picker/assets/19709109/53ef9f7f-96b5-4d4e-9b73-141f6d1b624c)
then:
![image](https://github.com/miguelpruivo/flutter_file_picker/assets/19709109/24edba55-b87b-4aa9-9e3f-3dd04d96922b)

After testing this logic, I found an error occurring in the part where the file retrieval is opened, which is the location shown in my screenshot. The issue is that when I click the button, there is no response, and the console displays this error log.

Error:
main.dart.js:5168 Uncaught Unsupported operation: Platform._operatingSystem
    at Object.c (http://38.47.108.163/main.dart.js:5168:19)
    at Object.b00 (http://38.47.108.163/main.dart.js:9314:15)
    at Object.b01 (http://38.47.108.163/main.dart.js:9346:16)
    at http://38.47.108.163/main.dart.js:106263:22
    at a.<computed> [as yG] (http://38.47.108.163/main.dart.js:30:36)
    at http://38.47.108.163/main.dart.js:106270:24
    at a.<computed> [as aQc] (http://38.47.108.163/main.dart.js:30:36)
    at Object.aVD (http://38.47.108.163/main.dart.js:10949:6)
    at http://38.47.108.163/main.dart.js:106298:23
    at a.<computed> [as aPV] (http://38.47.108.163/main.dart.js:27:35)

my code is very simple.
![image](https://github.com/miguelpruivo/flutter_file_picker/assets/19709109/3562cdcf-b0df-4ec1-b2bd-2e96eb3dcca7)
```

**Details to reproduce the issue**
Provide all the details to reproduce the issue.

**Error Log**
```
main.dart.js:5168 Uncaught Unsupported operation: Platform._operatingSystem
    at Object.c (http://38.47.108.163/main.dart.js:5168:19)
    at Object.b00 (http://38.47.108.163/main.dart.js:9314:15)
[...]

Issue #1483, opened 2024-04-09
Problem for Web  - having to use FilePickerWeb is not documented


graemep-nz:
This should be mentioned in the docs, at least until the underlying issue is fixed.

_Originally posted by @worldgamesdk in https://github.com/miguelpruivo/flutter_file_picker/issues/1343#issuecomment-2016232699_

This issue made me stop using flutter_file_picker and switch to file_selector package.        
The fact that you have to import a supposedly "internal" file is also not documented
import 'package:file_picker/_internal/file_picker_web.dart';

Issue #1555, opened 2024-07-18
Default plugins for linux/macos/windows need to register implementation; warning on latest Flutter


gnprice:
**Describe the bug**

Building an app using `file_picker` and Flutter from main, with a Flutter version since https://github.com/flutter/flutter/pull/137040 last week, produces these warnings (taken from full output below):
```
Package file_picker:linux references file_picker:linux as the default plugin, but it does not provide an inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: linux: default_package: file_picker` or add an inline implementation to file_picker via `platforms: linux:` `pluginClass` or `dartPluginClass`.

Package file_picker:macos references file_picker:macos as the default plugin, but it does not provide an inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: macos: default_package: file_picker` or add an inline implementation to file_picker via `platforms: macos:` `pluginClass` or `dartPluginClass`.

Package file_picker:windows references file_picker:windows as the default plugin, but it does not provide an inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: windows: default_package: file_picker` or add an inline implementation to file_picker via `platforms: windows:` `pluginClass` or `dartPluginClass`.
```

In fact just `flutter pub get` is enough to produce these warnings.

So here I am doing as the warnings say, by asking you to do as the warnings say :slightly_smiling_face: 

There's a bit of a lack of documentation of this new warning. Perhaps you already know what it's saying, but in case it's helpful here's how another plugin fixed a similar warning:
https://github.com/MaikuB/flutter_local_notifications/pull/2371
https://github.com/MaikuB/flutter_local_notifications/issues/2368

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop

**Platform OS version**
n/a (happens at `flutter pub get` time)

**How are you picking?**
n/a

**Details to reproduce the issue**

Run `flutter pub get` after cloning this repo: https://github.com/zulip/zulip-flutter
(while using a recent Flutter from the main/master channel, commit d55a5f96c117725c134099b78deb7af431a93902 or later)

Or see the logs of this CI run: https://github.com/zulip/zulip-flutter/actions/runs/10000051702/job/27641881195

**Error Log**
<details>

```
$ flutter pub get
Resolving dependencies... 
Downloading packages... 
[...]

PR #1567, opened 2024-08-02
fix default_plugin warnings with recent dart sdk

Sameri11:
This PR addresses #1555 and removes warning. Implementation is based on [this comment](https://github.com/flutter/flutter/issues/152037#issuecomment-2239909463) from https://github.com/flutter/flutter/issues/152037

Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1484
Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1483
Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1343

**Details:**  

Old implementation with `default_plugin` never did anything over than providing correct list of supported platforms for package page on pub.dev. Since some version of dart sdk, this behaviour has become warning.
New implementation provides (arguably) correct usage of `flutter:plugins:` section to avoid warning. Since all logic for choosing current platform and appropriate plugin's implementation is encapsulated inside current code, this PR just provides empty `registerWith` static methods needed for compilation. Also, it is mandatory for plugin implementations to be public, so we export them.

**Testing:**

Current tests are not affected by this PR. Example app is also functioning (tested on `android` and `ios`). Since tests on macOS (my dev machine) and example on both emulators works – I believe that functionality is not broken.

#1555 provides reproduction with https://github.com/zulip/zulip-flutter repo. This fix can be tested in that repo by adding `dependency_overrides` to `pubspec.yaml`. This override will point to my fork with fix:

```yaml
dependency_overrides:
  file_picker:
    git:
      url: git@github.com:Sameri11/flutter_file_picker.git
      ref: fix-issue-1555
```


Sameri11:
Hey @miguelpruivo, @philenius can you please review this fix and/or approve workflow?

lib/src/file_picker_io.dart
```diff
@@ -16,6 +16,8 @@ const EventChannel _eventChannel =
 
 /// An implementation of [FilePicker] that uses method channels.
 class FilePickerIO extends FilePicker {
+  static void registerWith() {}
```
navaronbracke, line 19: Don't you have to set `FilePicker.platform` to an instance of `FilePickerIO()` in `registerWith()` ?
(and the other implementations in the other registerWith() functions)

You might also need to double check the one for web.

Then you could get rid of the `filePickerWithFFI()` stub and fix a bunch of issues relating to using the `Platform` class on the web.

The following issues are caused by that bug: 
https://github.com/miguelpruivo/flutter_file_picker/issues/1484
https://github.com/miguelpruivo/flutter_file_picker/issues/1483
https://github.com/miguelpruivo/flutter_file_picker/issues/1343

If you could get that fixed in this PR also, that would be great, since it has been a bit of a pain to deal with (and I didn't have time to look into fixing it myself yet)

navaronbracke (changes requested):
Mainly a comment to see if we can fix a bunch of long standing issues with mis-use of `Platform._operatingSystem` in the plugin.

OlaiyaO:
I see you're all actively working on it.
I'm still experiencing the issue here. 

"Package file_picker:windows references file_picker:windows as the default plugin, but it does not provide an inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: windows: default_package: file_picker` or add an inline implementation to file_picker via `platforms: windows:` `pluginClass` or `dartPluginClass`."

navaronbracke:
> I see you're all actively working on it. I'm still experiencing the issue here.
> 
> "Package file_picker:windows references file_picker:windows as the default plugin, but it does not provide an inline implementation. Ask the maintainers of file_picker to either avoid referencing a default implementation via `platforms: windows: default_package: file_picker` or add an inline implementation to file_picker via `platforms: windows:` `pluginClass` or `dartPluginClass`."

That is indeed what this PR would aim to solve yes. Once we get the requires changes in this PR, I'll review it again and then we can release a fix version.

lib/src/file_picker_io.dart
Sameri11: Thanks for your review and additional clarification! My goal here was to provide minimal effort fix for the warning, hence this dummy implementation. 
I would try to make changes to address your comment soon, thanks.

Sameri11: > Don't you have to set FilePicker.platform to an instance of FilePickerIO() in registerWith() ?

About this – as far as I understand, it is the best practice :) So yes, I guess we can move logic there and it is preferred option. I would try that.
But current variant works because all logic for platform implementation choosing is already in _setPlatform(), so we needed only dummy empty method to satisfy compiler.

Sameri11: Well, I dived in this issue and I don't think that I'm happy with what I found :)

When platform implementation is added as `dartPluginClass`, this implementation must be exported from main library file, as per [dart_plugin_registrant template](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/flutter_plugins.dart#L621-L673)

In our case, it means that `lib/file_picker.dart` will look something like this for registrant to be able to resolve implementation classes:

```dart
library file_picker;

export './src/file_picker.dart';
export './src/platform_file.dart';
export './src/file_picker_result.dart';

// Added lines start from here

export './src/file_picker_macos.dart';
export './src/linux/file_picker_linux.dart';
export './src/file_picker_io.dart';
export './src/windows/file_picker_windows.dart';
```

Of course it will break in web build on last line when exporting windows implementation for very same reason (ffi types) which `filePickerWithFFI()` needed for. Basically, this abstraction just goes to another level, where, instead of stub implementation, stub export is needed.

I will investigate further, but at this point I'm not sure that some appropriate solution, without creation of federation plugins, is possible. 

Would love to hear another opinions.

navaronbracke: According to https://docs.flutter.dev/packages-and-plugins/developing-packages#dart-only-platform-implementations

you don't specifically have to export these? Exporting an implementation would only be needed if there is extra functionality exposed by the specific implementation (i.e. an extra method that only exists for that implementation)

Currently the plugin does not expose such "platform specific extra functionality". If that is something we want to provide, we should switch to providing sub-packages for each platform.

And if we absolutely need a conditional export, you could do the following:

1) In https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/windows/file_picker_windows.dart change the existing stub to return an instance of `FilePickerWindows`

```
FilePickerWindows filePickerWithFFI() => FilePickerWindows();
```

2) Use the existing stub from https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/windows/stub.dart
as the stub implementation for the conditional export.

Sameri11: > you don't specifically have to export these? 

I've also read this part of docs and thought like that too. But there is implementation detail which is not throughly described in documentation: framework generates `dart_plugin_registrant.dart` file where it calls `registerWith()` methods of concrete implementation. After `pubspec.yaml` modifications file looks like this:

```dart
//
// Generated file. Do not edit.
// This file is generated from template in file `flutter_tools/lib/src/flutter_plugins.dart`.
//

// @dart = 2.19

import 'dart:io'; // flutter_ignore: dart_io_import.

@pragma('vm:entry-point')
class _PluginRegistrant {

  @pragma('vm:entry-point')
  static void register() {
    if (Platform.isAndroid) {
      try {
        FilePickerIO.registerWith();
      } catch (err) {
        print(
          '`file_picker` threw an error: $err. '
          'The app may not function as expected until you remove this plugin from pubspec.yaml'
        );
      }

    } else if (Platform.isIOS) {
      try {
        FilePickerIO.registerWith();
      } catch (err) {
        print(
          '`file_picker` threw an error: $err. '
          'The app may not function as expected until you remove this plugin from pubspec.yaml'
        );
      }

    } else if (Platform.isLinux) {
      try {
        FilePickerLinux.registerWith();
      } catch (err) {
        print(
          '`file_picker` threw an error: $err. '
          'The app may not function as expected until you remove this plugin from pubspec.yaml'
        );
      }

    } else if (Platform.isMacOS) {
      try {
        FilePickerMacOS.registerWith();
      } catch (err) {
        print(
          '`file_picker` threw an error: $err. '
          'The app may not function as expected until you remove this plugin from pubspec.yaml'
        );
      }

    } else if (Platform.isWindows) {
      try {
        FilePickerWindows.registerWith();
      } catch (err) {
        print(
          '`file_picker` threw an error: $err. '
          'The app may not function as expected until you remove this plugin from pubspec.yaml'
        );
      }

    }
  }
}
```

As you can see, it only imports `package:file_picker/file_picker.dart`. And if I'm not wrong, there is no possibility to change this behaviour.


As for conditional import – I considered these options too. But I believe that some kind of better implementation is possible :) Thing is, while breaking web build, those exported members are never really used there. That's why I think that stub implementation is a little bit overkill and, maybe, just export a completely empty stub file is more appropriate.

But, again, I will investigate further and come back with some solution shortly.

And, of course, thanks for your guidance and feedback!

stuartmorgan: In the short term, you will indeed need to either do a conditional import layer (`path_provider_windows` and `path_provider_linux` have two different variants of that approach) to make the implementation class web-safe, or use the stub `defaultClass` approach.

Sameri11: @stuartmorgan thanks!

Sameri11:
@navaronbracke can you please take another look at this? I think I implemented what you talked about [here](https://github.com/miguelpruivo/flutter_file_picker/pull/1567#discussion_r1702126247).

Btw, I had hard times reproducing #1343 both based on this PR and master – no success. Either I do something wrong, or, maybe, these issues were fixed earlier somehow.

lib/file_picker.dart
```diff
@@ -3,3 +3,9 @@ library file_picker;
 export './src/file_picker.dart';
 export './src/platform_file.dart';
 export './src/file_picker_result.dart';
+export './src/file_picker_macos.dart';
+export './src/linux/file_picker_linux.dart';
+export './src/file_picker_io.dart';
+// Conditional export needed for web to successfully compile.
+export './src/windows/file_picker_windows_stub.dart'
+    if (dart.library.ffi) './src/windows/file_picker_windows_real.dart';
```
navaronbracke, line 11: Can we keep the `file_picker_windows.dart` filename for the real implementation? (rather than `file_picker_windows_real.dart`)
The fact that the fake one is called `file_picker_windows_stub.dart` clarifies its use in my opinion.

lib/file_picker.dart
```diff
@@ -3,3 +3,9 @@ library file_picker;
 export './src/file_picker.dart';
 export './src/platform_file.dart';
 export './src/file_picker_result.dart';
+export './src/file_picker_macos.dart';
+export './src/linux/file_picker_linux.dart';
+export './src/file_picker_io.dart';
+// Conditional export needed for web to successfully compile.
```
navaronbracke, line 9:
```suggestion
// Conditional export needed for web to successfully compile,
// as `dart:ffi` is not available on the web.
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 8.0.7
+### General
```
navaronbracke, line 2: The last bullet point isn't needed, since that is an implementation detail.
Also, perhaps we should reword it to something like this?
I also added a specific note for people who were using a workaround for the Platform class issue like:

```
if (kIsWeb) {
  FilePickerWeb()
} else {
  FilePicker()
}
```

```suggestion
### General
- Fixes an issue relating to incorrect registration of platform-specific implementations. [#1555](https://github.com/miguelpruivo/flutter_file_picker/issues/1555).
- Fixes an issue that required users to use `kIsWeb` and `FilePickerWeb` when on the web.
```

navaronbracke:
I also edited the original PR comment to list fixed issues, so these get closed when the PR is merged.

As for not being able to reproduce https://github.com/miguelpruivo/flutter_file_picker/issues/1343 I'm not sure either.

lib/file_picker.dart
Sameri11: Have no strong opinion about this. So yes, done :)

CHANGELOG.md
Sameri11: Fully agree here, thanks!

navaronbracke (approved):
LGTM!

I also edited out the redundant bullet points (one was an implementation detail, which users don't care about and the other was now a duplicate point since we already mention fixing #1555 in the first point)

navaronbracke:
It looks one of the test is failing, because https://github.com/miguelpruivo/flutter_file_picker/pull/1561/files did not update a test expectation. I'll address this with a commit and then this can land!

navaronbracke:
Checks are green, so ship it!

PR #1578, opened 2024-08-13
Fix web example on Flutter 3.22

navaronbracke-endare:
This PR fixes the example for the web, that now needs a different method of bootstrapping in Flutter 3.22

I had forgotten to include this commit in my last PR, after testing the web example.
 
We still hit the CanvasKit undefined module error when running the example a second time, but this was reported upstream in flutter/flutter.

### DIFF
--- a/example/web/index.html
+++ b/example/web/index.html
@@ -16,16 +16,8 @@
 
   <title>example</title>
   <link rel="manifest" href="manifest.json">
-  <script src="flutter.js"></script>
 </head>
 <body>
-  <!-- This script installs service_worker.js to provide PWA functionality to
-       application. For more information, see:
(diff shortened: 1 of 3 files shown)


example/web/index.html
```diff
@@ -16,16 +16,8 @@
 
   <title>example</title>
   <link rel="manifest" href="manifest.json">
-  <script src="flutter.js"></script>
```
navaronbracke-endare, line 19: See https://docs.flutter.dev/platform-integration/web/initialization

navaronbracke:
The undefined module error is tracked for a cherry pick to stable, see https://github.com/flutter/flutter/issues/152844

navaronbracke (approved):

PR #1587, opened 2024-08-27
adding support for the “m4a” format

siyangstar:
adding support for the “m4a” format

### DIFF
--- a/lib/src/linux/kdialog_handler.dart
+++ b/lib/src/linux/kdialog_handler.dart
@@ -56,7 +56,7 @@ class KDialogHandler implements DialogHandler {
       case FileType.any:
         return '';
       case FileType.audio:
-        return 'Audio File (${DialogHandler.toCaseInsensitive("*.aac *.midi *.mp3 *.ogg *.wav")})';
+        return 'Audio File (${DialogHandler.toCaseInsensitive("*.aac *.midi *.mp3 *.ogg *.wav *.m4a")})';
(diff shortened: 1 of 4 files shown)


navaronbracke:
What about support on other platforms?

siyangstar:
> What about support on other platforms?

Thank you for your feedback. Regarding the .m4a support on other platforms:

Android: .m4a files are natively supported on Android (4.1 and above), and should work well with file_picker. However, I will perform additional testing to ensure seamless integration.

iOS: Since .m4a is part of Apple’s AAC format, it’s natively supported on iOS devices. This format is commonly used and works reliably with iOS apps.

Windows: .m4a files are generally supported on Windows 10 and above due to built-in media capabilities. I will verify if file_picker handles .m4a correctly on Windows.

macOS: Native support for .m4a is excellent on macOS, with Finder and Quick Look providing easy access and playback. I will confirm compatibility with the plugin.

Web: Modern browsers like Chrome, Firefox, Safari, and Edge all support .m4a playback. I will ensure file_picker correctly handles these files in web applications.

siyangstar:
Additionally, for Linux, .m4a files are generally supported by most modern distributions, especially when the necessary codecs (like those from GStreamer or FFmpeg) are installed.

stephane-archer (approved):

miguelpruivo:
Will be added in the next version. Thanks.

PR #1615, opened 2024-10-17
Fix null safety in filepickerdelegate

Lilian-C:


android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -86,25 +86,26 @@ public boolean onActivityResult(final int requestCode, final int resultCode, fin
...
-                if (uri != null) {
-                  String  path = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS)
-                            .getAbsolutePath() + File.separator + FileUtils.getFileName(uri, this.activity);
-                    try {
-                        OutputStream outputStream = this.activity.getContentResolver().openOutputStream(uri);
-                        if(outputStream != null){
-                            outputStream.write(bytes);
-                            outputStream.flush();
-                            outputStream.close();
+                if (data != null) {
```
navaronbracke, line 89: Can you use an early return instead?

Lilian-C: Sure, I fixed it :)

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -85,6 +85,8 @@ public boolean onActivityResult(final int requestCode, final int resultCode, fin
         // Save file
         if (requestCode == SAVE_FILE_CODE) {
             if (resultCode == Activity.RESULT_OK) {
+                if (data == null)
```
navaronbracke, line 88: Since the other if's also use curly braces, can you do that here as well?

Also, if you update the patch version and add an entry in the changelog, then this can be published.

Lilian-C: Sure, I added all of this

navaronbracke (approved):

PR #1618, opened 2024-10-23
Fixed an issue where null was returned when picking up files

nespjin:
Fixed an issue that always returned null when selecting files on Redmi NOTE 9


nespjin (commented):
Using Intent directly does not return null

miguelpruivo (approved):

miguelpruivo:
Thanks.

PR #1649, opened 2024-12-16
[fix] Fix the issue where the last modified time of a file is lost after selecting the file.

ChenTian-plaud:
Fix the issue where the last modified time of a file is lost after selecting the file.


vicajilau:
Hi @ChenTian-plaud,

We are closing this PR as it is outdated and targets the old Java implementation. 
 
Furthermore, with the recent release of version 12, we have introduced full support for Android SAF. The `PlatformFile` now returns an `identifier` (the original Android URI), which allows developers to query all original metadata (including the last modified time) directly from the source, rather than relying on the cached copy.

We believe this is a more robust way to handle file metadata moving forward. Thank you for your contribution!

Issue #1647, opened 2024-12-14
undefined registrar variable in Java code of file_picker plugin


jacklee1995:
**Describe the bug**

-----------------
AppData\Local\Pub\Cache\hosted\pub.dev\file_picker-8.1.5\android\src\main\java\com\mr\flutter\plugin\filepicker\FilePickerPlugin.java:257: ����: �Ҳ�������
        if (registrar != null) {
            ^
  ����:   ���� registrar
  λ��: �� FilePickerPlugin
C:\Users\a2911\AppData\Local\Pub\Cache\hosted\pub.dev\file_picker-8.1.5\android\src\main\java\com\mr\flutter\plugin\filepicker\FilePickerPlugin.java:260: ����: �Ҳ�������
            registrar.addActivityResultListener(this.delegate);
            ^
  ����:   ���� registrar
  λ��: �� FilePickerPlugin
2 ������

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':file_picker:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.

* Try:
> Run with --info option to get more log output.
> Run with --scan to get full insights.

BUILD FAILED in 9s
---------------------------

It seems that this error is caused by the use of undefined registrar variable in Java code of file_picker plugin, which leads to compilation failure.

According to the error message, the problem lies in lines 257 and 260 of the' FilePickerPlugin.java' file. The code tries to access a variable named' registrar', but this variable is not defined.

----------------
Android 14,   
Flutter 3.24.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision dec2ee5c1f (4 weeks ago) • 2024-11-13 11:13:06 -0800
Engine • revision a18df97ca5
Tools • Dart 3.5.4 • DevTools 2.37.3

Similar errors occurred in flutter_file_picker 8.1.5 and 8.1.4.

TheAnh188:
> **Describe the bug**
> 
[...]

PR #1644, opened 2024-12-11
removing v1 embedding to fix compile on flutter 3.27

abdelaziz-mahdy:
fixing this https://github.com/miguelpruivo/flutter_file_picker/issues/1643

PR #1650, opened 2024-12-16
Fix [#1647]

ksmohamed:


android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
```diff
@@ -23,6 +23,7 @@
 import io.flutter.plugin.common.EventChannel;
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
+import io.flutter.plugin.common.PluginRegistry.Registrar;
```
MrCsabaToth, line 26: Why would you bring V1 back even more? It's deprecated for good.
Correct solution: https://github.com/miguelpruivo/flutter_file_picker/pull/1644

MrCsabaToth (changes requested):

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerPlugin.java
ksmohamed: This package are missing so its showing error now its clear

ksmohamed:
bug fixed

JakobLichterfeld:
Please drop as #1648 is real solution

JakobLichterfeld (changes requested):
Please drop, real solution is #1648 
btw. the PR title must be self-explanatory

Issue #751, opened 2021-06-20
[web] path is always null


maxzod:
**Describe the bug**
file picker always reutrn `null` as the path of the picked file but returns the file name right !

**Platform**
 Web

**Platform OS version**
Version 91.0.4472.106 (Official Build) (64-bit)

**How are you picking?**
https://gist.github.com/maxzod/060e33849a3f6729e9d83165bee3d9c5

**Details to reproduce the issue**
flutter 2.2.0
dart 2.12  
file_picker: ^3.0.2+2

maxzod:
sorry my bad 🙏

maxzod:
#754

PR #1665, opened 2024-12-31
Add Blob Creation as Path in flutter_file_picker

vicajilau:
# Pull Request: Add Blob Creation as Path in flutter_file_picker

## Description

This PR introduces the capability to create a blob from the picked file and set it as the path in the `flutter_file_picker` plugin. This change allows for better handling of file data directly as URLs, enhancing the flexibility of file management in web applications.

## Changes Made

- **Blob Creation**: Implemented the creation of a blob from the selected file and set it as the path in the `PlatformFile` object.
  
### Code Snippet

```dart
void addPickedFile(
  File file,
  Uint8List? bytes,
  String? path,
  Stream<List<int>>? readStream,
) {
  final blob = Blob([bytes ?? Uint8List(0)]);
  final blobUrl = Url.createObjectUrl(blob); // Create a blob URL

  pickedFiles.add(PlatformFile(
    name: file.name,
    path: blobUrl, // Set the blob URL as the path
    size: bytes != null ? bytes.length : file.size,
    bytes: bytes,
    readStream: readStream,
  ));

  if (pickedFiles.length >= files.length) {
    if (onFileLoading != null) {
      onFileLoading(FilePickerStatus.done);
    }
    filesCompleter.complete(pickedFiles);
  }
}
```

## Benefits
Enhanced File Handling: Allows for seamless integration of file data as URLs, making it easier to work with files in web applications.
Checklist
- [x] I have tested the changes locally.
## Related Issues
- #751 
- It currently requires different flows in universal applications. This minimises the gap between platforms.

### Additional Notes
Please review the changes and let me know if further modifications are needed. Thank you! 🤗


vicajilau:
Well done @daniJimen and many thanks!!

lib/_internal/file_picker_web.dart
```diff
@@ -1,10 +1,11 @@
 import 'dart:async';
+import 'dart:html' as html;
```
navaronbracke, line 2: `dart:html` is not WASM compatible & is deprecated in Dart 3.7

lib/src/platform_file.dart
```diff
@@ -33,19 +33,7 @@ class PlatformFile {
...
-      /// https://github.com/miguelpruivo/flutter_file_picker/issues/751
-      throw '''
-      On web `path` is unavailable and accessing it causes this exception.
-      You should access `bytes` property instead,
-      Read more about it [here](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ)
-      ''';
-    }
-    return _path;
-  }
+  final String? path;
```
navaronbracke, line 36: The DartDoc comment for this getter will need to be updated.

navaronbracke (changes requested):

lib/_internal/file_picker_web.dart
vicajilau: Fixed!!

lib/src/platform_file.dart
vicajilau: Right!! But I don't have write permissions there

navaronbracke: What do you mean? You can just edit the doc comment on line 35?

navaronbracke: Do you mean the Wiki? I can edit the wiki right before merging.

vicajilau: Yes 🤗

vicajilau: Got it! Changed DartDoc on 35 line :rocket:

lib/_internal/file_picker_web.dart
```diff
@@ -86,9 +86,15 @@ class FilePickerWeb extends FilePicker {
         String? path,
         Stream<List<int>>? readStream,
       ) {
+        final blobParts = bytes ?? Uint8List(0);
```
navaronbracke, line 89: Can we avoid allocating a Blob if the bytes is null / empty?

I also worry about leaking the underlying Blob, as the url from `URL.createObjectURL()` is not getting freed with 
https://pub.dev/documentation/web/latest/web/URL/revokeObjectURL.html

lib/src/platform_file.dart
```diff
@@ -31,21 +31,9 @@ class PlatformFile {
   /// ```
   /// final File myFile = File(platformFile.path);
   /// ```
-  /// On web this is always `null`. You should access `bytes` property instead.
+  /// On web, this will return a blob URL instead of a file path.
```
navaronbracke, line 34:
```suggestion
  /// On web the path points to a Blob URL, if present.
```

navaronbracke (changes requested):

lib/_internal/file_picker_web.dart
vicajilau: Removed! I am aware that it is not released, but I have not had any problems associated with this. Do you suggest another approach? I think it is worth unifying the flows 🚀

lib/_internal/file_picker_web.dart
```diff
@@ -86,9 +86,16 @@ class FilePickerWeb extends FilePicker {
         String? path,
         Stream<List<int>>? readStream,
       ) {
+        String? blobUrl;
+        if (bytes != null) {
```
navaronbracke, line 90: We should probably also check that the length of the bytes isn't zero? That is just a sanity check, but you never know what users will do.

lib/src/platform_file.dart
```diff
@@ -31,21 +31,9 @@ class PlatformFile {
   /// ```
   /// final File myFile = File(platformFile.path);
   /// ```
-  /// On web this is always `null`. You should access `bytes` property instead.
+  /// On web the path points to a Blob URL, if present.
```
navaronbracke, line 34: Should we perhaps tell users to call `URL.revokeObjectURL` on the path when they are done with the file (on web)?

```suggestion
/// On web the path points to a Blob URL, if present, which can be cleaned up using [URL.revokeObjectURL].
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.3.8
+### Web
+- **BREAKING CHANGE:** `pickFiles()` now loads files as blobs. See the note in the updated [wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki/api#-pickfiles)
```
navaronbracke, line 3: Breaking changes should bump the major version, though.

navaronbracke (changes requested):

lib/_internal/file_picker_web.dart
vicajilau: Done!!

CHANGELOG.md
vicajilau: Makes sense!! Done 🚀

navaronbracke (approved):

navaronbracke:
@vicajilau I updated the note on the Wiki as well

Issue #1669, opened 2025-01-03
Auto naming incrémentation when duplicate filename after extension


baptistesx:
Hi,

i am using the last version of the package 8.1.7.

When I use saveFile() and a file with the same filename already exists, it increments but after the actual filename (containing its extension).

Example: "filename.pdf" already exists 

The new file will be saved under "filename.pdf (1)". So it breaks the extension and the filesystem will not recognize the file as a pdf.

The solution would be to put the increment before the extension.

Temporary solution: add a uuid between the name and the extension (I used a 5 chars length uuid, it should be fine)

```
 Future<void> _saveFileToDownloads() async {
    if (localPath == null) return;

    try {
      // Generate a 5-character UUID
      final String uuid = DateTime.now().microsecondsSinceEpoch.toString();
      final String shortUuid = uuid.substring(uuid.length - 5);

      // Separate the file name and the extension
      final int lastDot = widget.filename.lastIndexOf('.');
      final String nameWithoutExt = widget.filename.substring(0, lastDot);
      final String extension = widget.filename.substring(lastDot);

      // Create the new file name with the UUID
      final String newFilename = '${nameWithoutExt}_$uuid$extension';

      final HearingAttachmentType fileType =
          HearingAttachment.getTypeFromFilename(widget.filename);
      FileType pickerFileType;
      List<String>? allowedExtensions;

      // Configure the type and extensions based on the file type
      switch (fileType) {
        case HearingAttachmentType.image:
          pickerFileType = FileType.image;
          allowedExtensions = <String>['jpg', 'jpeg', 'png'];
        case HearingAttachmentType.pdf:
          pickerFileType = FileType.custom;
          allowedExtensions = <String>['pdf'];
        case HearingAttachmentType.document:
        case HearingAttachmentType.other:
          pickerFileType = FileType.any;
          allowedExtensions = null;
      }

      final String? outputFile = await FilePicker.platform.saveFile(
        dialogTitle: 'Save File',
        fileName: newFilename,
        allowedExtensions: allowedExtensions,
        type: pickerFileType,
        bytes: await File(localPath!).readAsBytes(),
        lockParentWindow: true,
      );

      if (outputFile != null) {
        setState(() {
          fileSaved = true;
        });

        if (context.mounted) {
          showSuccessToast(
            text: 'File saved successfully',
            title: 'File Saved',
          );
        }
      }
    } catch (e) {
      print('Save error: $e');
      if (context.mounted) {
[...]

PR #1682, opened 2025-01-14
feat-add-mime-type

EArminjon:
Light temporary solution until better one for https://github.com/miguelpruivo/flutter_file_picker/issues/1669

The `savefile` method can be more simple, as a dev, we want to save a file by showing the native prompt. In theory, we don't have to provide `allowedExtensions` and `type` properties but we need a `mimeType` which is missing.

### DIFF
--- a/lib/src/file_picker.dart
+++ b/lib/src/file_picker.dart
@@ -162,6 +162,9 @@ abstract class FilePicker extends PlatformInterface {
   /// stay in front of the Flutter window until it is closed (like a modal
   /// window). This parameter works only on Windows desktop.
   ///
+  /// [mimeType] can be optionally set to provide a mime type for the file.
+  /// It only work for iOS and Android.
+  ///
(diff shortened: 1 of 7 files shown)


Xsantosh:
@miguelpruivo can you please spare some free time check this feature and merge it.

lib/src/file_picker.dart
```diff
@@ -162,6 +162,9 @@ abstract class FilePicker extends PlatformInterface {
   /// stay in front of the Flutter window until it is closed (like a modal
   /// window). This parameter works only on Windows desktop.
   ///
+  /// [mimeType] can be optionally set to provide a mime type for the file.
```
navaronbracke, line 165: This argument somewhat conflicts with `FileType` when the FileType is any or custom.
Can we instead get the mime type from the file bytes or name?

This also only does something for Android in your PR, despite claiming that it works on iOS also.

navaronbracke:
Mime type detection was implemented as part of the Kotlin migration, so this can be closed.

PR #1684, opened 2025-01-20
fix(android): prevent OOM-crashes when selecting large files

nomo-app:
Hi,
This is about the issue that our app crashes with an out-of-memory-crash when selecting large files (>100 Megabytes or so).
Currently, Nomo is running on an older branch, but if you are interested in merging this fix then I would be willing to rebase this fix onto a newer branch and to potentially also make it configurable via Dart-arguments.


vicajilau:
This has been resolved by using an approach that does not involve limiting the file size to a fixed value, but rather by handling exceptions appropriately.

Issue #1680, opened 2025-01-14
[macOS] Picked file with name containing forward slash can't be found 


kaciula:
On macOS, the GUI allows for file names to contain the forward slash character, like: `Cha/Cha 1 Start.mp4` In terminal, this same file looks like: `Cha:Cha 1 Start.mp4` This seems to be "normal" behavior on macOS. Forward slash inside filenames is allowed in GUI but colon is not. The opposite is true in terminal: colon is allowed in terminal but forward slash is not. Accordingly. macOS shows a different filename for the same file in the GUI vs terminal. If you create file with a colon in terminal, its name will contain a forward slash in Finder or any other GUI.

My issue is that my macOS app cannot find a picked file that contains forward slashes in the name. Trying to copy it does not work as the path has a wrong forward slash due to the file name. The forward slash needs to be replaced with a colon in order for the operation to work.

This issue does not appear on iOS as the picked file has the correct filename with the colon replacing the forward slash.

Issue #1105, opened 2022-08-31
How to pick files or directory?


BirjuVachhani:
I know we can use `pickFiles()` to open file picker and `getDirectoryPath()` to choose a directory but my use case allows user to select files or directory or both at the same time. How can I achieve this with this plugin?

For example, I also implemented drag and drop files support using [desktop_drop](https://pub.dev/packages/desktop_drop) package which supports both files and directory dropping. So when I drop things, it gives me `List<XFile>` where can contain directories if I have selected any. I can't do this with this plugin. A lot of native apps does this. 

I have tested this on MacOS. `pickFiles()` doesn't allow selecting directories too. Only files. 

I'd like to allow both at the same time. Can you please help?

BirjuVachhani:
Any updates?

miguelpruivo:
Which platform are you trying this on?

au2001:
@miguelpruivo I have the same use-case as OP and can confirm that exactly what they described happens when using macOS (Monterey 12.6).

With `pickFiles`, when selecting a directory and clicking `Choose`, the directory is simply opened as if I had double-clicked it, rather than being selected.
With `getDirectoryPath`, all files are grayed out and thus cannot be selected. Also, it is not possible to select multiple folders at once.

BirjuVachhani:
Commenting to keep this issue open.

au2001:
up

Aunali321:
@miguelpruivo can you reopen this issue?

bobmoff:
This was immediately the first thing i realized was missing. Need a way to select a mixture of multiple files and folders in the same picker window. The native picker on macOS has support for this at least.

Without this feature, opening av picker becomes a much worse experience, compared to drag and drop. But some people are not tech savvy enough to perform a drop and drop, especially not on a small laptop where cmd+tab needs to be involved at the same time when dragging.

🤞🏻

stephane-archer:
up, please @miguelpruivo reopen this issue!

stephane-archer:
https://github.com/miguelpruivo/flutter_file_picker/issues/1628

stephane-archer:
<img width="1440" alt="Screenshot 2024-11-13 at 16 31 52" src="https://github.com/user-attachments/assets/dc7e68e6-e0e9-4453-bbea-ff19a8f2502f">

On MacOS with this plugin, it's impossible to select files and folders at the same time @miguelpruivo

BirjuVachhani:
[...]

Issue #1469, opened 2024-03-20
Package is very hard to use, nullability semantics in particular


matanlurey:
_I want to mention I think this package is awesome, and saves me tons of work on hobby projects. Kudos._

That being said, the API state of the project leaves a lot to be desired. Some suggestions:

- Consider adding a `.pickFile` with semantics that are less confusing. The default for `.pickFiles` is ironically `allowMultiple: false`, but that also means I need to write overly defensive code just to see if a file was actually picked:

    ```dart
    final result = await FilePicker.platform.pickFiles(...);
    if (result == null || result.files.isEmpty) {
      return;
    }
    final file = result.files.single;
    // ... I finally have a file.
    ```

    I could imagine a much better API:
    
    ```dart
    final file = await FilePicker.platform.pickFile(...):
    if (file == null) {
      return;
    }
    // ... I have a file
    ```

    In general, I think named arguments are over-used in Dart. They are great for things where users might need to customize (at runtime) arguments, but are less great for different code branches (it's rare that you will want to, dynamically, pick a single or multiple files - usually that's static in business logic).

- Document how `PlatformFile` is intended to be used most of the time. It has three properties that are all nullable (`path`, `bytes`, `readStream`) and 0 documentation explaining why any of these would be null, or what is even intended to be used. Hitting null assertions at runtime sucks.

    In addition, the presence of `bytes` and `readStream` is really confusing - the latter makes it seem like I can read from the file asynchronously, but the former synchronously (and perhaps it's even pre fetched, because it's stored as `final Uint8List? bytes`. Which one is it? Why do I need both?

I have more suggestions, but these are the two that hit me in every project I start with this package.

miguelpruivo:
Hi, thank you for the suggestion but you are actually saving just 1 loc with that change vs. the effort that has to be put in order to refactor the API at this point, specially having support for all the platforms. The `pickFiles` supports multiple properties that are quite verbose and somewhat easy to understand, specially because they are very well documented and with examples on the Wiki.

The way the API was made was to make it as much as seamless experience between the different platforms and not only on mobile, so the devs don't have to call different API's for each like:
```
if(kIsWeb){
// Pick in a way
} else if(Platform.isIOS){
// Use another
}
```
Specially when multiple devs support multiple platforms at once.

[...]

PR #1690, opened 2025-01-29
Reimplement file picker on macOS with method channels

### LINKED ISSUE #1445: pickFiles dialog does not receive focus on macOS
**Describe the bug**

I try to open an "open" dialog as follows, in reaction to a click on a material button:

```dart
Future<PlatformFile?> openFile(
  BuildContext context, {
  required String title,
  required String extension,
}) async {
  Navigator.push(context, _createModalBackdrop());
  try {
    FilePickerResult? result = await FilePicker.platform.pickFiles(
      dialogTitle: title,
      type: FileType.custom,
      allowedExtensions: [extension],
      withReadStream: true,
      lockParentWindow: true,
    );
    return result?.files.single;
  } finally {
    // ignore: use_build_context_synchronously
    Navigator.pop(context);
  }
}
```

...and the dialog pops up above my macOS application, but the open files dialog is not focused; the app retains focus.

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop

**Platform OS version**

macOS 14.2.1

**Flutter Version details**

```
ianh@coupari app % flutter doctor -v
[✓] Flutter (Channel main, 3.20.0-7.0.pre.31, on macOS 14.2.1 23C71 darwin-arm64, locale en-US)
    • Flutter version 3.20.0-7.0.pre.31 on channel main at /Users/ianh/dev/flutter
    • Upstream repository git@github.com:flutter/flutter.git
    • Framework revision 5025681129 (55 minutes ago), 2024-02-13 21:48:41 +0000
    • Engine revision 0091a4914d
    • Dart version 3.4.0 (build 3.4.0-138.0.dev)
    • DevTools version 2.33.0-dev.6

[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.


[!] Xcode - develop for iOS and macOS (Xcode 15.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15C500b
    ✗ Unable to get list of installed Simulator runtimes.
    • CocoaPods version 1.14.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions).

[✓] VS Code (version 1.85.2)
    • VS Code at /Users/ianh/Downloads/Visual Studio Code.app/Contents
    • Flutter extension version 3.82.0

[✓] Connected device (3 available)
    • macOS (desktop)                 • macos                 • darwin-arm64   • macOS 14.2.1 23C71 darwin-arm64
    • Mac Designed for iPad (desktop) • mac-designed-for-ipad • darwin         • macOS 14.2.1 23C71 darwin-arm64
    • Chrome (web)                    • chrome                • web-javascript • Google Chrome 121.0.6167.184

[✓] Network resources

! Doctor found issues in 3 categories.
ianh@coupari app % 
```

### LINKED ISSUE #1492: Reimplement macos file picker using native APIs in the app rather than calling out to osascript
The current approach results in this weird UI experience where the file picker isn't focused correctly, and where there's a new system process for the dialog that is separate from the actual app. It also runs into problems if osascript is disabled for some reason (e.g. security policy).

There doesn't seem to be any reason why we could implement this using regular message channels to code implemented natively in the app, like a normal plugin.

### LINKED ISSUE #1582: Add Swift Package Manager support
Hello,

Flutter is migrating to [Swift Package Manager][]. Please [add Swift Package Manager support][] to your plugin.

Flutter will eventually deprecate and then remove support for CocoaPods. Adding Swift Package Manager support future-proofs your plugins.

[Swift Package Manager]: https://www.swift.org/documentation/package-manager/
[add Swift Package Manager support]: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin

### How to add Swift Package Manager support to your plugins

You can find the migration guide here: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin

If you run into issues or have questions, please reach out to the Flutter team. You can ping me at @loic-sharma, send a message to the #hackers-ios channel on [Flutter’s discord][], or [open a GitHub issue][].

Thank you for your wonderful contributions to the Flutter ecosystem!

[Flutter’s discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[open a GitHub issue]: https://github.com/flutter/flutter/issues/new/choose

### Why is Flutter migrating to Swift Package Manager?

[CocoaPods is now in maintenance mode][].

Swift Package Manager support is one of Flutter’s most requested features: [flutter#33850][].

Flutter's Swift Package Manager integration has several benefits:

1. **Access to the Swift package ecosystem**. Flutter plugins can use the growing ecosystem of [Swift packages][].
2. **Simplifies Flutter installation**. Swift Package Manager is bundled with Xcode. In the future, you won’t need to install Ruby and CocoaPods to target iOS or macOS.

[CocoaPods is now in maintenance mode]: https://blog.cocoapods.org/CocoaPods-Support-Plans/
[flutter#33850]: https://github.com/flutter/flutter/issues/33850
[Swift packages]: https://swiftpackageindex.com/

### LINKED ISSUE #1674: [suggestion] make lockParentWindow work on MacOS
```` Dart
  /// If [lockParentWindow] is set, the child window (file picker window) will
  /// stay in front of the Flutter window until it is closed (like a modal
  /// window). This parameter works only on Windows desktop.
  Future<String?> saveFile({
    String? dialogTitle,
    String? fileName,
    String? initialDirectory,
    FileType type = FileType.any,
    List<String>? allowedExtensions,
    Uint8List? bytes,
  }) async =>
```

### LINKED ISSUE #1685: `FilePicker.platform.getDirectoryPath()` returns on MacOS path containing symbolic links
`FilePicker.platform.getDirectoryPath()` returns on MacOS path containing symbolic links

``` Dart
    String? selectedDirectory = await FilePicker.platform.getDirectoryPath();
    if (selectedDirectory == null) {
      // User canceled the picker
      return;
    }
    String selectedDirectoryCleaned =
        await Directory(selectedDirectory).resolveSymbolicLinks();
```
here:
`selectedDirectory == "/Volumes/Macintosh HD/Users/fractale/Desktop/LutTests"`
`selectedDirectoryCleaned == "/Users/fractale/Desktop/LutTests"`

orestesgaolin:
This PR:

- closes #1492 
- closes #1445 
- closes #1674
- closes #1685

It is still somewhat prone to #1680 - if the picked file contains `/` then it's replaced by `:`

I was wondering if I should address #1582 but figured it would be better to migrate both iOS and macOS later. I'll try if I find some time.

Now we can also address #1105 but that would require public API change.

Can someone confirm if this addresses #1685?

Here's a demo of this plugin on macOS 15.1.1 (24B91)

https://github.com/user-attachments/assets/9f2ee1f9-34fc-4c62-8268-79a7c28b12d1

Feel free to add the note below to https://github.com/miguelpruivo/flutter_file_picker/wiki/Setup

> On macOS it's necessary to enable "Read" User Selected File entitlement for selecting files and "Read/Write" for saving.

screenshot_20250129_103407


lib/src/file_picker_macos.dart
```diff
@@ -90,31 +79,27 @@ class FilePickerMacOS extends FilePicker {
     Uint8List? bytes,
     bool lockParentWindow = false,
   }) async {
-    final String executable = await isExecutableOnPath('osascript');
-    final String fileFilter = fileTypeToFileFilter(
+    assert(bytes == null, "Passing bytes is not supported on macOS.");
```
navaronbracke, line 82: Can we use an `UnsupportedError` instead? I wonder how we do this for other platforms (i.e. web)

macos/file_picker.podspec
```diff
@@ -0,0 +1,25 @@
...
+  s.license          = { :file => '../LICENSE' }
+  s.author           = 'Miguel Ruivo, Dominik Roszkowski'
+  s.source           = { :path => '.' }
+  s.source_files = 'Classes/**/*'
+
+  s.resource_bundles = {'file_picker_privacy' => ['Resources/PrivacyInfo.xcprivacy']}
+
+  s.dependency 'FlutterMacOS'
+
+  s.platform = :osx, '10.11'
```
navaronbracke, line 22: Just to double-check but does this align with Flutter's minimum for MacOS for the current minimum Flutter version that we support?

lib/src/file_picker.dart
```diff
@@ -50,34 +50,42 @@ abstract class FilePicker extends PlatformInterface {
...
   /// that will give you the current status of picking.
+  /// Not supported on macOS.
   ///
   /// If [allowCompression] is set, it will allow media to apply the default OS compression.
   /// Defaults to `true`.
   ///
   /// If [lockParentWindow] is set, the child window (file picker window) will
   /// stay in front of the Flutter window until it is closed (like a modal
   /// window). This parameter works only on Windows desktop.
+  /// On macOS parent window will be locked and this parameter is ignored.
```
navaronbracke, line 70:
```suggestion
  /// On macOS the parent window will be locked and this parameter is ignored.
```

lib/src/file_picker.dart
```diff
@@ -50,34 +50,42 @@ abstract class FilePicker extends PlatformInterface {
...
   ///
-  /// [dialogTitle] can be optionally set on desktop platforms to set the modal window title. It will be ignored on
-  /// other platforms.
+  /// [dialogTitle] can be optionally set on desktop platforms to set the modal window title.
+  /// Not supported on macOS. It will be ignored on other platforms.
   ///
   /// [initialDirectory] can be optionally set to an absolute path to specify
   /// where the dialog should open. Only supported on Linux, macOS, and Windows.
+  /// On macOS home directory shortcut (~/) is not necessary and passing it will be ignored.
```
navaronbracke, line 77:
```suggestion
  /// On macOS the home directory shortcut (~/) is not necessary and passing it will be ignored.
```

lib/src/file_picker.dart
```diff
@@ -50,34 +50,42 @@ abstract class FilePicker extends PlatformInterface {
...
   ///
   /// [readSequential] can be optionally set on web to keep the import file order during import.
+  /// Not supported on macOS.
   ///
   /// The result is wrapped in a [FilePickerResult] which contains helper getters
   /// with useful information regarding the picked [List<PlatformFile>].
   ///
   /// For more information, check the [API documentation](https://github.com/miguelpruivo/flutter_file_picker/wiki/api).
   ///
+  /// Note: on macOS User Selected File Read entitlement is required.
```
navaronbracke, line 87:
```suggestion
  /// Note: This requires the User Selected File Read entitlement on macOS.
```

lib/src/file_picker.dart
```diff
@@ -112,14 +120,17 @@ abstract class FilePicker extends PlatformInterface {
...
   /// Note: Some Android paths are protected, hence can't be accessed and will return `/` instead.
   ///
-  /// [dialogTitle] can be set to display a custom title on desktop platforms. It will be ignored on Web & IO.
+  /// [dialogTitle] can be set to display a custom title on desktop platforms.
+  /// Not supported on macOS. It will be ignored on other platforms.
   ///
   /// If [lockParentWindow] is set, the child window (file picker window) will
   /// stay in front of the Flutter window until it is closed (like a modal
   /// window). This parameter works only on Windows desktop.
+  /// On macOS parent window will be locked and this parameter is ignored.
```
navaronbracke, line 129:
```suggestion
  /// On macOS the parent window will be locked and this parameter is ignored.
```

lib/src/file_picker.dart
```diff
@@ -112,14 +120,17 @@ abstract class FilePicker extends PlatformInterface {
...
   ///
   /// [initialDirectory] can be optionally set to an absolute path to specify
   /// where the dialog should open. Only supported on Linux, macOS, and Windows.
+  /// On macOS home directory shortcut (~/) is not necessary and passing it will be ignored.
```
navaronbracke, line 133:
```suggestion
  /// On macOS the home directory shortcut (~/) is not necessary and passing it will be ignored.
```

lib/src/file_picker.dart
```diff
@@ -128,6 +139,7 @@ abstract class FilePicker extends PlatformInterface {
   /// Note: on Windows, throws a `WindowsException` with a detailed error message, if the dialog
   /// could not be instantiated or the dialog result could not be interpreted.
   /// Note: Some Android paths are protected, hence can't be accessed and will return `/` instead.
+  /// Note: on macOS User Selected File Read entitlement is required.
```
navaronbracke, line 142:
```suggestion
  /// Note: The User Selected File Read entitlement is required on macOS.
```

lib/src/file_picker.dart
```diff
@@ -140,10 +152,12 @@ abstract class FilePicker extends PlatformInterface {
   ///
   /// For mobile platforms, this function will save file with [bytes] to return a path.
   ///
-  /// For desktop platforms (Linux, macOS & Windows),This function does not actually
+  /// For desktop platforms (Linux, macOS & Windows) this function does not actually
   /// save a file. It only opens the dialog to let the user choose a location and
   /// file name. This function only returns the **path** to this (non-existing) file.
   ///
+  /// On macOS a User Selected File Read/Write entitlement is required.
```
navaronbracke, line 159:
```suggestion
  /// The User Selected File Read/Write entitlement is required on macOS.
```

lib/src/file_picker.dart
```diff
@@ -152,6 +166,7 @@ abstract class FilePicker extends PlatformInterface {
   ///
   /// [initialDirectory] can be optionally set to an absolute path to specify
   /// where the dialog should open. Only supported on Linux, macOS, and Windows.
+  /// On macOS home directory shortcut (~/) is not necessary and passing it will be ignored.
```
navaronbracke, line 169:
```suggestion
  /// On macOS the home directory shortcut (~/) is not necessary and passing it will be ignored.
```

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private func handleFileSelection(
+    _ call: FlutterMethodCall, result: @escaping FlutterResult
+  ) {
+    let dialog = NSOpenPanel()
+    let args = call.arguments as! [String: Any]
+
+    dialog.showsResizeIndicator = true
+    dialog.directoryURL = URL(
```
navaronbracke, line 37: What if the directory url is invalid? Do we need to handle this with a `result(FlutterError())` ? Or do we not set the url?

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+      fileURLWithPath: args["initialDirectory"] as? String ?? ""
+    )
+    dialog.showsHiddenFiles = false
+    dialog.allowsMultipleSelection = args["allowMultiple"] as? Bool ?? false
+    dialog.canChooseDirectories = false
```
navaronbracke, line 42: Do we need to set `canChooseFiles` to true explicitly?

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  private func handleFileSelection(
+    _ call: FlutterMethodCall, result: @escaping FlutterResult
+  ) {
+    let dialog = NSOpenPanel()
+    let args = call.arguments as! [String: Any]
+
+    dialog.showsResizeIndicator = true
```
navaronbracke, line 36: This property is deprecated https://developer.apple.com/documentation/appkit/nswindow/showsresizeindicator

Can you use an availability check?

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+    dialog.directoryURL = URL(
+      fileURLWithPath: args["initialDirectory"] as? String ?? ""
+    )
+    dialog.showsHiddenFiles = false
+    dialog.allowsMultipleSelection = args["allowMultiple"] as? Bool ?? false
+    dialog.canChooseDirectories = false
+    dialog.allowedFileTypes = args["allowedExtensions"] as? [String] ?? []
```
navaronbracke, line 43: This property is deprecated: https://developer.apple.com/documentation/appkit/nssavepanel/allowedfiletypes

Can you switch to the replacement on newer MacOS versions, using an availability check?

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+
+    if dialog.runModal() == NSApplication.ModalResponse.OK {
+      let pathResult = dialog.url
```
navaronbracke, line 46: This doesn't look like it is correct? The `url` is probably nil when in multiple selection mode.
Can you check `dialog.urls` directly instead? That is an array so you'd need the first item when in single selection mode (or return nil if it's empty)

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+    result(nil)
+  }
+
+  private func handleDirectorySelection(
+    _ call: FlutterMethodCall, result: @escaping FlutterResult
+  ) {
+    let dialog = NSOpenPanel()
+    let args = call.arguments as! [String: Any]
+
+    dialog.showsResizeIndicator = true
```
navaronbracke, line 72: Same remark for the deprecated `showsResizeIndicator`

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+    dialog.directoryURL = URL(
```
navaronbracke, line 73: Same remark for invalid URLs.

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+    let dialog = NSSavePanel()
+    let args = call.arguments as! [String: Any]
+
+    dialog.title = args["dialogTitle"] as? String ?? ""
+    dialog.showsTagField = false
+    dialog.showsHiddenFiles = false
+    dialog.canCreateDirectories = true
+    dialog.nameFieldStringValue = args["fileName"] as? String ?? ""
+
+    if let initialDirectory = args["initialDirectory"] as? String,
```
navaronbracke, line 103: Same remark for invalid URLs.

macos/Classes/FilePickerPlugin.swift
```diff
@@ -0,0 +1,123 @@
...
+      !initialDirectory.isEmpty
+    {
+      dialog.directoryURL = URL(fileURLWithPath: initialDirectory)
+    }
+
+    let allowedExtensions = args["allowedExtensions"] as? [String] ?? []
+    if !allowedExtensions.isEmpty {
+      dialog.allowedFileTypes = allowedExtensions
```
navaronbracke, line 111: Same remark for the deprecated `allowedFileTypes`

navaronbracke (changes requested):
Overall this looks great! Just some minor feedback around deprecations & docs.

example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+@testable import file_picker
+
+// This demonstrates a simple unit test of the Swift portion of this plugin's implementation.
+//
+// See https://developer.apple.com/documentation/xctest for more information about using XCTest.
+
+class RunnerTests: XCTestCase {
+
+  func testGetPlatformVersion() {
```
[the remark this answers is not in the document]
orestesgaolin: good catch, sorry -- left the default

example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -0,0 +1,12 @@
+//
```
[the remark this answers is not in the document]
orestesgaolin: thanks, will check what's the convention around these files

macos/file_picker.podspec
orestesgaolin: Created with `flutter create -t plugin` on current stable, so assumed it should follow the current requirements. Does that sound right?

lib/src/file_picker_macos.dart
orestesgaolin: done

navaronbracke:
@orestesgaolin We can add SPM support for MacOS/iOS as a follow-up PR.

navaronbracke:
I will check if this fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1685 with symbolic links.
If not, then we should probably do the resolution internally.

macos/file_picker.podspec
navaronbracke: I think this should be fine. We support Flutter 3.22 or higher, and that Flutter version requires a higher MacOS minimum, so it's Flutter that will dictate the minimum MacOS version.

navaronbracke:
@orestesgaolin For https://github.com/miguelpruivo/flutter_file_picker/issues/1105 we could add a new method, that allows picking both files and directories?

What if we add a method that allows both? But we also need to pay attention to the feedback in https://github.com/miguelpruivo/flutter_file_picker/issues/1469

example/macos/RunnerTests/RunnerTests.swift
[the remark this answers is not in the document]
orestesgaolin: I've created a basic suite of tests but frankly I don't find them useful, rather noisy. They require Xcode and app to launch to test so imho we can leave this empty as you suggested.

```swift
import Cocoa
import FlutterMacOS
import XCTest

@testable import file_picker

// This demonstrates a simple unit test of the Swift portion of this plugin's implementation.
//
// See https://developer.apple.com/documentation/xctest for more information about using XCTest.

class RunnerTests: XCTestCase {
  var plugin: FilePickerPlugin!

  override func setUp() {
    super.setUp()
    plugin = FilePickerPlugin()
  }

  override func tearDown() {
    plugin = nil
    super.tearDown()
  }

  func testUnimplementedMethod() {
    let call = FlutterMethodCall(methodName: "unknownMethod", arguments: nil)

    let resultExpectation = expectation(description: "result block must be called")
    plugin.handle(call) { result in
      XCTAssertTrue(result as? NSObject == FlutterMethodNotImplemented)
      resultExpectation.fulfill()
    }
    waitForExpectations(timeout: 1)
  }

  func testPickFilesArgumentParsing() {
    let arguments: [String: Any] = [
      "initialDirectory": "/test/path",
      "allowMultiple": true,
      "allowedExtensions": ["pdf", "doc"],
    ]

    let call = FlutterMethodCall(methodName: "pickFiles", arguments: arguments)

    let resultExpectation = expectation(description: "result block must be called")
    plugin.handle(call) { result in
      // Since we can't actually show the dialog in tests, we expect nil
      XCTAssertNil(result)
      resultExpectation.fulfill()
    }
    waitForExpectations(timeout: 1)
  }

  func testDirectorySelectionArgumentParsing() {
    let arguments: [String: Any] = [
      "initialDirectory": "/test/path"
    ]

    let call = FlutterMethodCall(methodName: "getDirectoryPath", arguments: arguments)

    let resultExpectation = expectation(description: "result block must be called")
    plugin.handle(call) { result in
      // Since we can't actually show the dialog in tests, we expect nil
      XCTAssertNil(result)
      resultExpectation.fulfill()
    }
    waitForExpectations(timeout: 1)
  }

  func testSaveFileArgumentParsing() {
    let arguments: [String: Any] = [
      "dialogTitle": "Test Dialog",
      "fileName": "test.txt",
      "initialDirectory": "/test/path",
      "allowedExtensions": ["txt", "doc"],
    ]

    let call = FlutterMethodCall(methodName: "saveFile", arguments: arguments)

    let resultExpectation = expectation(description: "result block must be called")
    plugin.handle(call) { result in
      // Since we can't actually show the dialog in tests, we expect nil
      XCTAssertNil(result)
      resultExpectation.fulfill()
    }
    waitForExpectations(timeout: 1)
  }

  func testEmptyArguments() {
    let emptyArguments: [String: Any] = [:]

    // Test pickFiles with empty arguments
    let pickFilesCall = FlutterMethodCall(methodName: "pickFiles", arguments: emptyArguments)
    let pickFilesExpectation = expectation(description: "pickFiles result block must be called")
    plugin.handle(pickFilesCall) { result in
      XCTAssertNil(result)
      pickFilesExpectation.fulfill()
    }

    // Test directory selection with empty arguments
    let dirCall = FlutterMethodCall(methodName: "getDirectoryPath", arguments: emptyArguments)
    let dirExpectation = expectation(description: "getDirectoryPath result block must be called")
    plugin.handle(dirCall) { result in
      XCTAssertNil(result)
      dirExpectation.fulfill()
    }

    // Test save file with empty arguments
    let saveCall = FlutterMethodCall(methodName: "saveFile", arguments: emptyArguments)
    let saveExpectation = expectation(description: "saveFile result block must be called")
    plugin.handle(saveCall) { result in
      XCTAssertNil(result)
      saveExpectation.fulfill()
    }

    waitForExpectations(timeout: 1)
  }

  func testInvalidArguments() {
    let invalidArguments: [String: Any] = [
      "allowMultiple": "invalid_bool_value",  // Should be boolean
      "allowedExtensions": "invalid_array",  // Should be array
    ]

    let call = FlutterMethodCall(methodName: "pickFiles", arguments: invalidArguments)

    let resultExpectation = expectation(description: "result block must be called")
    plugin.handle(call) { result in
      XCTAssertNil(result)
      resultExpectation.fulfill()
    }
    waitForExpectations(timeout: 1)
  }

}

```

macos/Classes/FilePickerPlugin.swift
orestesgaolin: macOS falls back to the default or previous correct folder. I can add this to doc

macos/Classes/FilePickerPlugin.swift
orestesgaolin: I can add this for clarity but I think it's true by default.

macos/Classes/FilePickerPlugin.swift
orestesgaolin: For multiple selections the `.url` includes the first of the files, so it's safe to access it for check, but I think you are right that to reduce confusion it may be better to check `.urls`

screenshot_20250129_134502

macos/Classes/FilePickerPlugin.swift
orestesgaolin: updated and unified for pick and save

macos/Classes/FilePickerPlugin.swift
orestesgaolin: deprecated entirely, no need to use

macos/Classes/FilePickerPlugin.swift
orestesgaolin: updated docs

orestesgaolin:
@navaronbracke I applied all suggestions and responded to comments, marked them as resolved -- hope you don't mind. Feel free to unresolve any pending issues or questions.

orestesgaolin:
Btw I can bump the version and update the changelog here too, should I bump it to 8.2.0 or 8.1.8?

navaronbracke (approved):
LGTM. Can you update the version to 8.2.0 and update the changelog? Then this can be released when the PR is merged :)

navaronbracke:
I still need to check if this handles symlinks, so I'll probably do that before we merge this.

navaronbracke:
The analysis failure is unrelated to this PR. I'll look into that as well.

navaronbracke:
> Btw I can bump the version and update the changelog here too, should I bump it to 8.2.0 or 8.1.8?

Version 8.2.0 is fine.

> The analysis failure is unrelated to this PR. I'll look into that as well.

The analysis reports 3 switch cases where the `default` clause can be removed, since it is unreachable.
Maybe you can address this, too?

```
warning • lib/src/linux/kdialog_handler.dart:68:7 • This default clause is covered by the previous cases. Try removing the default clause,
          or restructuring the preceding patterns. • unreachable_switch_default
warning • lib/src/linux/qarma_and_zenity_handler.dart:60:7 • This default clause is covered by the previous cases. Try removing the default
          clause, or restructuring the preceding patterns. • unreachable_switch_default
warning • lib/src/windows/file_picker_windows.dart:230:7 • This default clause is covered by the previous cases. Try removing the default
          clause, or restructuring the preceding patterns. • unreachable_switch_default

3 issues found.
```

navaronbracke:
@orestesgaolin I was wrong about the minimum MacOS version. When I run the example app with your branch, I get

```
/Users/navaronbracke/Desktop/flutter_file_picker-implement-on-macos-method-channel/example/macos/Pods/Pods.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.11, but the range of supported deployment target versions is 10.13 to 15.2.99. (in target 'file_picker-file_picker_privacy' from project 'Pods')
```

So the minimum should be `10.13` and not 10.11.

navaronbracke:
@orestesgaolin This does seem to fix the symlink bug!

I did the following to test:
1) `cd ~/Desktop && mkdir test_folder && cd test_folder && mkdir test_folder_2 && cd test_folder_2`
2) I then created some file in `nested_folder`, i.e. `Untitled.rtf`
3) Make a symlink: `ln -s "/Users/navaronbracke/Desktop/test_folder/nested_folder" /Users/navaronbracke/Desktop/test_folder_2`, so that the `nested_folder` is hidden by the symlink
4) run the example app
5) In the file picker dialog, go into the `test_folder_2` (has an arrow icon, so it's the symlink folder)
6) Select the `Untitled.rtf` file
7) The result path is `/Users/navaronbracke/Desktop/test_folder/nested_folder/Untitled.rtf`, so without the symlink (test_folder_2)

orestesgaolin:
Updated min version and proposed a changelog. Also tweaked list items in few recent releases to look nicer ;)

navaronbracke:
@orestesgaolin Did you manage to look into the analysis failures on the Dart side?

You can remove the following `default` switch cases, as these are now reported as unreachable. Sadly, this isn't yet supported by `dart fix`, so you'll have to do it manually.

https://github.com/orestesgaolin/flutter_file_picker/blob/implement-on-macos-method-channel/lib/src/linux/kdialog_handler.dart#L68-L69

https://github.com/orestesgaolin/flutter_file_picker/blob/implement-on-macos-method-channel/lib/src/linux/qarma_and_zenity_handler.dart#L60-L61

https://github.com/orestesgaolin/flutter_file_picker/blob/implement-on-macos-method-channel/lib/src/windows/file_picker_windows.dart#L230-L231

These prevent the updated tests to run in CI. I can do this later, but in case you want to address this still.

navaronbracke:
@orestesgaolin I went ahead and put out a patch for the lint fixes (on my corp account, but hey)
Once that is done I'll rebase your PR and then this can land!

orestesgaolin:
Sorry, was doing some of my corp stuff too ;) you were quicker!

navaronbracke:
No worries! I'll get it sorted real quick!

navaronbracke:
Tree is green! I'll check in with pub in a moment. Thanks!

stephane-archer:
@navaronbracke @orestesgaolin thank you for your amazing work! I can't wait to ship with it!

navaronbracke:
This should be available on Pub now.

orestesgaolin:
🕺

potatosalad775:
It would be appreciated if someone could update the wiki to add what's mentioned above.

Not enabling permissions under 'Signing & Capabilities' - 'App Sandbox' - 'File Access' causes the .pickFile function to return null without any explanation, and I was just wondering who keeps aborting my file picker function :)

navaronbracke:
@potatosalad775 Thanks for the reminder. Yes that was mentioned in the OP. I'll look into updating the Wiki.

navaronbracke:
Done!

Issue #1700, opened 2025-02-06
returns null when having allowedExtensions: ['jpg', 'png', 'pdf']


jtavio:
**Describe the bug**
when I configure my filepicker with what I want it to filter and either when selecting image or pdf the return is null, but when I define in the allowedExtensions only images or only pdf it brings me the correct answer, so I don't know why it doesn't work when I put images and pdf to select?

**Platform**

- [x ] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
Android 15 (SDK 35) running physical phone.

**How are you picking?**
```
                    FilePickerResult? getBill;
                    getBill = await FilePicker.platform.pickFiles(
                      type: FileType.custom,
                      allowMultiple: true,
                      allowedExtensions: ['jpg', 'png', 'pdf'],
                    );
```

**Details to reproduce the issue**
test in this way 
```
                    FilePickerResult? getBill;
                    getBill = await FilePicker.platform.pickFiles(
                      type: FileType.custom,
                      allowMultiple: true,
                      allowedExtensions: ['jpg', 'png', 'pdf'],
                    );
```
and it will return null, whatever you select, either image or pdf.

and thus returns the selected information

```
                    FilePickerResult? getBill;
                    getBill = await FilePicker.platform.pickFiles(
                      type: FileType.custom,
                      allowMultiple: true,
                      allowedExtensions: ['jpg', 'png'] // or allowedExtensions: ['pdf'],
                    );
```

**Error Log**
I/FilePickerDelegate(21258): User cancelled the picker request

**Flutter Version details**
✓] Flutter (Channel stable, 3.27.1, on macOS 13.6 22G120 darwin-arm64, locale en-VE)
    • Flutter version 3.27.1 on channel stable at /Users/gregspinetti/Development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 17025dd882 (hace 7 semanas), 2024-12-17 03:23:09 +0900
    • Engine revision cb4b5fff73
    • Dart version 3.6.0
    • DevTools version 2.40.2

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/gregspinetti/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
[...]

Issue #1404, opened 2023-11-08
Unable to pick files with `custom` or `any` file type on Android 14


VladStarikovEk:
**Describe the bug**
When picking files with `custom` or `any` file type on Android 14, when one or multiple files are picked it just closes assuming that the user closed it. I also got `I/FilePickerDelegate(21624): User cancelled the picker request` message in logs.

When picking `image`, `audio`, `video` or `media` everything works fine, but the picker has a `gallery` UI, rather then `file explorer` UI. I also checked this on Android 13 emulator - it works file there.

**Platform**

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
Pixel 5 - Android 14

**How are you picking?**
```
final pickerResult = await _filePicker.pickFiles(
      type: FileType.custom,
      allowMultiple: true,
      allowedExtensions: ['jpg', 'jpeg', 'png', 'pdf'],
 );
```
or
```
final pickerResult = await _filePicker.pickFiles(
      type: FileType.any,
);
```

**Details to reproduce the issue**

**Error Log**
```
D/FilePickerDelegate(21624): Selected type */*
I/FilePickerDelegate(21624): User cancelled the picker request
```

**Screenshots and/or video**

https://github.com/miguelpruivo/flutter_file_picker/assets/99865323/ade3be44-b8f9-48b8-a509-d7aff62de0ad
<img width="932" alt="Screenshot 2023-11-08 at 15 35 34" src="https://github.com/miguelpruivo/flutter_file_picker/assets/99865323/0d2d2d0f-86ba-45d2-8e76-e77b5c8c4c47">

**Flutter Version details**
```
❯ flutter doctor -v
[✓] Flutter (Channel stable, 3.13.9, on macOS 14.1 23B74 darwin-arm64, locale en-UA)
    • Flutter version 3.13.9 on channel stable at /Users/vstarikov/fvm/versions/3.13.9
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision d211f42860 (2 weeks ago), 2023-10-25 13:42:25 -0700
    • Engine revision 0545f8705d
    • Dart version 3.1.5
    • DevTools version 2.25.0

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/vstarikov/Library/Android/sdk
    • Platform android-34, build-tools 33.0.0
[...]

PR #1701, opened 2025-02-07
fix (Android) pickFiles with launchMode=singleInstance

jay-k98:
This pull request targets the issue https://github.com/miguelpruivo/flutter_file_picker/issues/1700 and uses the solution from the discussion in https://github.com/miguelpruivo/flutter_file_picker/issues/1404 to fix pickFiles when launchMode="singleInstance" is configured in AndroidManifest.xml

### DIFF
--- a/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
+++ b/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
@@ -246,7 +246,11 @@ public class FilePickerDelegate implements PluginRegistry.ActivityResultListener
             if (type.equals("image/*")) {
                 intent = new Intent(Intent.ACTION_PICK, android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
             } else {
-                intent = new Intent(Intent.ACTION_GET_CONTENT);
+                if(Build.VERSION.SDK_INT >= 19) {
+                    intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
+                } else {
+                    intent = new Intent(Intent.ACTION_GET_CONTENT);
+                }
                 intent.addCategory(Intent.CATEGORY_OPENABLE);
             }
             final Uri uri = Uri.parse(Environment.getExternalStorageDirectory().getPath() + File.separator);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.3.2
+### Android
+- Fix pickFiles for Android with launchMode=singleInstance [#1700](https://github.com/miguelpruivo/flutter_file_picker/issues/1700), [#1404](https://github.com/miguelpruivo/flutter_file_picker/issues/1404)
+
 ## 8.3.1
 ### iOS
 - Fix [1367](https://github.com/miguelpruivo/flutter_file_picker/issues/1367)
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -247,6 +247,11 @@ private void startFileExplorer() {
                 intent = new Intent(Intent.ACTION_PICK, android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
             } else {
                 intent = new Intent(Intent.ACTION_GET_CONTENT);
+	        if(Build.VERSION.SDK_INT >= 19) {
```
theDeniZ, line 250: correct indentation missing on this and next line

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -247,6 +247,11 @@ private void startFileExplorer() {
                 intent = new Intent(Intent.ACTION_PICK, android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
             } else {
                 intent = new Intent(Intent.ACTION_GET_CONTENT);
```
theDeniZ, line 249: is this still needed?

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -246,7 +246,11 @@ private void startFileExplorer() {
             if (type.equals("image/*")) {
                 intent = new Intent(Intent.ACTION_PICK, android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI);
             } else {
-                intent = new Intent(Intent.ACTION_GET_CONTENT);
+                if(Build.VERSION.SDK_INT >= 19) {
```
navaronbracke, line 249: Can you update the changelog & version for this fix?

navaronbracke (changes requested):

android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
jay-k98: Done. Thanks for the advice :)

navaronbracke (approved):

mbfakourii:
@navaronbracke 

Next merge this PR and next version 8.3.1
we can not see this gallery picker in android 14

image

and we see like this

image

I had to downgrade to version 8.3.1!

navaronbracke:
@mbfakourii The screenshot you refer to is only when you use the https://developer.android.com/training/data-storage/shared/photopicker API? That is currently not yet implemented as such, though.

navaronbracke:
See also https://github.com/miguelpruivo/flutter_file_picker/issues/1705

mbfakourii:
> @mbfakourii The screenshot you refer to is only when you use the https://developer.android.com/training/data-storage/shared/photopicker API? That is currently not yet implemented as such, though.

So why is it available in version 8.3.1?

navaronbracke:
Well, https://developer.android.com/reference/android/content/Intent#ACTION_OPEN_DOCUMENT is the newer Intent type, so I would not expect this to be a bug. That said, the old `Intent.ACTION_GET_CONTENT` is not deprecated. @jay-k98 Was there anything specific that needed the other intent type, other than the launch mode? Because the original change seemed fairly benign?

jay-k98:
Yes, besides the launch Mode there was this weird behaviour described in https://github.com/miguelpruivo/flutter_file_picker/issues/1700. Images or pdf could be picked in isolate but not when combined in allowedExtensions. The fix seemingly solved both problems.

Issue #1706, opened 2025-02-12
[macOS] pickFiles always returns null after updating to 8.3.0


vicajilau:
## **Description**  
After updating to `flutter_file_picker` version **8.3.0**, calling `pickFiles()` on macOS always returns `null` without opening the file picker dialog. This issue occurs even if the implementation was working fine in previous versions.  

## **Steps to Reproduce**  
1. Upgrade `flutter_file_picker` to `8.3.0`.  
2. Run the following code on macOS:  
   ```dart
   final result = await FilePicker.platform.pickFiles();
   print(result); // Always prints null

3. Observe that the file picker does not open and the result is null.

## **Expected Behavior** 
The file picker should open as expected, allowing the user to select files.

## **Actual Behavior**  
No file picker window appears, and `pickFiles()` immediately returns `null`.  

## **Cause**  
This is due to a **silent requirement introduced in version 8.3.0**:  
- The **User Selected File Read entitlement (`com.apple.security.files.user-selected.read-only`)** is now **mandatory** on macOS.  
- Previously, this entitlement was not required.  

## **Solution**  
To fix this issue, developers need to manually **add the entitlement** to both `DebugProfile.entitlements` and `Release.entitlements`:  

```xml
<key>com.apple.security.files.user-selected.read-only</key>
<true/>
````

## **Proposed Fix**  
- This change should be documented as a **breaking change** in the changelog.  
- The package should provide a **clear error message** when the entitlement is missing, instead of silently returning `null`.

acastanos:
Great info. Thx!!!!

navaronbracke:
@vicajilau I'm not sure if we can detect a missing entitlement at runtime?

vicajilau:
I don't know if it makes sense to make such an effort. For me, the fact that it can be read both in the changelog and in the documentation is enough. This issue can be closed as completed 🚀

PR #1707, opened 2025-02-12
Add Breaking Change for macOS Entitlement Requirement

vicajilau:
### Description  
This pull request introduces a breaking change for macOS users of the `flutter_file_picker` package.  

### Breaking Change  
Starting from version **8.3.0**, the `pickFiles()` method now requires the entitlement `com.apple.security.files.user-selected.read-only` on macOS. If this entitlement is not present, the method will return `null` and the file picker will not open, potentially leading to confusion among developers upgrading from previous versions where this entitlement was not required.  

### Justification  
- This change was introduced silently in version **8.3.0** without clear documentation, which can cause issues for developers who may not be aware of the requirement.
- To ensure a smooth transition and proper functionality of the file picker on macOS, it is crucial to document this change as breaking in the changelog and update the API documentation accordingly.

### Changes Made  
- Updated the `CHANGELOG.md` to include the breaking change notice.
- Recommended that users add the entitlement to both `DebugProfile.entitlements` and `Release.entitlements` to restore the expected functionality.

### Related Issue  
- Reference to issue: #1706

### Additional Notes  
- Future updates should include a warning or error message when the entitlement is missing to improve user experience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,6 @@
 - Reimplement macOS file picker using method channels (fixes [#1492](https://github.com/miguelpruivo/flutter_file_picker/issues/1492), [#1445](https://github.com/miguelpruivo/flutter_file_picker/issues/1445), [#1674](https://github.com/miguelpruivo/flutter_file_picker/issues/1674), [#1685](https://github.com/miguelpruivo/flutter_file_picker/issues/1685))
 - **BREAKING CHANGE:** `pickFiles()` now requires the `com.apple.security.files.user-selected.read-only` entitlement on macOS. Without this entitlement, the file picker will not open, and the method will return `null`. See the note in the updated [wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki/Setup#macos)
 
-
 ## 8.1.7
 ### iOS 
 - Fix Image Picker to Handle Partial Failures Gracefully [#1554](https://github.com/miguelpruivo/flutter_file_picker/issues/1554)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 ## 8.3.0
 ### Desktop (macOS) && iOS
 - Adds support for Swift Package Manager for compatibility with new projects [#1582](https://github.com/miguelpruivo/flutter_file_picker/issues/1582)
+### macOS
```
navaronbracke, line 25: This was added on the Wiki: https://github.com/miguelpruivo/flutter_file_picker/wiki/Setup#macos

Maybe add a `See also: https://github.com/miguelpruivo/flutter_file_picker/wiki/Setup#macos` instead of pointing to the linked PR ?

navaronbracke (changes requested):

CHANGELOG.md
vicajilau: Makes sense! I didn't see it. On its way

vicajilau:
Resolved @navaronbracke  🚀

CHANGELOG.md
navaronbracke: @vicajilau Apologies, but due to the other change that just landed, this is now in the wrong place.

Can you put it at the bottom of 

```
## 8.2.0
### Desktop (macOS)
- Reimplement macOS file picker using method channels ...
```

navaronbracke (changes requested):

CHANGELOG.md
vicajilau: Fixed 💪

navaronbracke (approved):

navaronbracke:
@vicajilau I adjusted the changelog note a little, so that it is in bold in the markdown

PR #1708, opened 2025-02-13
Windows fix: already initialized COM theading

alnitak:
Hi,

I am the author of [flutter_soloud](https://github.com/alnitak/flutter_soloud) plugin. Has been reported that when the audio player has already been initialized, an error occurs when trying to use `getDirectoryPath` in this [issue](https://github.com/alnitak/flutter_soloud/issues/180) with this error:
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Error 0x80010106: Cannot change thread mode after it is set.
```

This should fix the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer window doesn't pop up.

Digging around and searching on the web, I found that it's safe to use an already created COM and this PR checks this case, but I'm not familiar with Windows, so it might be worth doing a bit more research.

I can provide a simple project using *flutter_soloud* together *flutteR_file_picker* if you want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,9 @@
-## 8.3.7
-### Desktop (Windows)
-- Fixes the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer windows doesn't popup. [#1708](https://github.com/miguelpruivo/flutter_file_picker/pull/1708)
-
 ## 8.3.6
 ### General
 - Added compatibility with Flutter 3.29. [@vicajilau](https://github.com/vicajilau).
+
+### Desktop (Windows)
+- Fixes the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer windows doesn't popup. [#1708](https://github.com/miguelpruivo/flutter_file_picker/pull/1708)
 
 ## 8.3.5
 ### Android
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.3.6
+### Desktop (Windows)
+-- Fixes the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer windows doesn't popup. [#1708](https://github.com/miguelpruivo/flutter_file_picker/pull/1708)
+
 ## 8.3.5
 ### Android
 - Fixes allowCompression not working on Android. [1633](https://github.com/miguelpruivo/flutter_file_picker/issues/1633)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.3.6
+### Desktop (Windows)
+-- Fixes the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer windows doesn't popup. [#1708](https://github.com/miguelpruivo/flutter_file_picker/pull/1708)
```
navaronbracke, line 3: Nit: small typo in the changelog

```suggestion
- Fixes the issue under Windows that could be raised when another plugin uses `CoInitializeEx` and the File Explorer windows doesn't popup. [#1708](https://github.com/miguelpruivo/flutter_file_picker/pull/1708)
```

lib/src/windows/file_picker_windows.dart
```diff
@@ -100,7 +100,12 @@ class FilePickerWindows extends FilePicker {
       COINIT.COINIT_APARTMENTTHREADED | COINIT.COINIT_DISABLE_OLE1DDE,
     );
 
-    if (!SUCCEEDED(hr)) throw WindowsException(hr);
+    // Ignore the error if COM is already initialized.
+    // If this is the case, CoInitializeEx will return
+    // RPC_E_CHANGED_MODE (0x80010106 = -2147417850).
```
navaronbracke, line 105: Can we use the `RPC_E_CHANGED_MODE` constant from win32 instead?

See https://github.com/halildurmus/win32/blob/main/packages/win32/lib/src/constants_nodoc.dart#L2241

navaronbracke (changes requested):
Just some minor comments, but otherwise LGTM.

lib/src/windows/file_picker_windows.dart
alnitak: Sure, I didn't pay attention to constants!

Fixed

alnitak:
Thanks for your fast review!

navaronbracke (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.3.7
```
navaronbracke, line 1: https://pub.dev/packages/file_picker/changelog doesn't list 8.3.6 yet, so that is possibly still awaiting publish

navaronbracke: Edit: it seems the previous change didn't get a pubspec change, so it was not yet published. Give me a moment to address that

navaronbracke (approved):

navaronbracke:
The Flutter 3.29.0 compatibility change was scheduled for 8.3.6, but was not published.
Instead, I'm going to release this bugfix alongside that change.

navaronbracke:
This should be published with version 8.3.6 in a few moments

alnitak:
> This should be published with version 8.3.6 in a few moments

That's great! Thanks for your incredibly fast support!

Issue #1636, opened 2024-11-23
saveFile always return the download path on Android


hioshih:
**Describe the bug**
Reopening report [1590](https://github.com/miguelpruivo/flutter_file_picker/issues/1590).
As stated, saveFileDialog returns a file path pointing to the download folder regardless of the directory picked by the user.

**Workaround**
A workaround is to send the file data directly through bytes, but if you intended to check on the saved file afterwards you will fail to do so.
But then, this will become incompatible with Desktop, because on windows it does return the correct folder but doesn't save the bytes contents automatically.

**Platform**

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**How are you picking?**
```
var fileName;
    String data = 'this is the\n text file content';
    final bytes = Uint8List.fromList(data.codeUnits);
    //picks a folder and filename to save
    fileName = await FilePicker.platform.saveFile(
      bytes: bytes,
    );
    //saves in the correct folder but enforces returning string as the download folder
    print('saveFile returned file: $fileName');
```
file saved in "/storage/emulated/0/myfolder/saved.txt"
output:  "saveFile returned file:: /storage/emulated/0/Download/saved.txt"

**Details to reproduce the issue**
Brand new blank flutter project.

hioshih:
up

hioshih:
> This issue is stale because it has been open for 7 days with no activity.

This bot is just closing all the issues aggressively (7 days) without solving anything and neither giving a chance for people to deal with them. 😢

alexaungmyooo:
May I know how do you solve this issue?

hioshih:
> May I know how do you solve this issue?

I didn't. 

It would either give an exception, or I could send the file contents using the bytes which is a shady undocumented parameter. But even so, it wouldn't return the correct folder where the file was saved.

Tried to look into the code but it seemed very complex with almost no comments, so I couldn't figure out yet why it stopped working, even though I'm almost sure it wasn't like this in the past. My guess is that some recent updates broke this feature on android.

PR #1712, opened 2025-02-17
[Android:Fix] Fix an issue where `saveFile` would return an incorrect path 

Dogbonee:
Currently `saveFile` always returns a path to downloads with the filename appended on the end. This doesn't seem like intended behavior, but the path seemed to be hardcoded to the downloads directory environment variable. This PR just changes the path to the activity's uri real path. Also fixed a minor issue in the example where the `bytes` argument was required to save, but was not included in the example.

Related to issue [#1636](https://github.com/miguelpruivo/flutter_file_picker/issues/1636)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,26 @@
-## 9.0.1
+## 9.2.1
 ### Android
 - Fixes an issue where `saveFiles()` would return an incorrect path. [#1636](https://github.com/miguelpruivo/flutter_file_picker/issues/1636)
+
+## 9.2.0
+### Desktop (macOS, Windows, Linux)
+- Fixes an inconsistency for saveFile that did not save the file, when bytes are provided on desktop platforms. [@vicajilau](https://github.com/vicajilau).
+
+## 9.1.0
+### Web
+- Added an implementation for `saveFile()` on the web. [@vicajilau](https://github.com/vicajilau).
+
+## 9.0.3
+### Android
+- Added ProGuard consumer rules to automatically exclude `androidx.lifecycle.DefaultLifecycleObserver` from obfuscation.
+
+## 9.0.2
+### Android
+- Fixes: File streams not getting closed.
+
+## 9.0.1
+### Windows
+- Move `getDirectoryPath()` to its own isolate to avoid COM initialization conflicts with other plugins [#1713](https://github.com/miguelpruivo/flutter_file_picker/pull/1713) [@tguerin](https://github.com/tguerin). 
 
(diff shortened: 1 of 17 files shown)


navaronbracke:
@Dogbonee Do you have some before/after samples to verify your fix?

navaronbracke:
> As a side note, the merge that you committed changed the bytes argument in the example from Uint8List(0) to _paths?.first.bytes, which caused a Invalid Argument exception when I tested it with the example.

Hmm, that error probably has to do with a recent change for `saveFile`. I'll take a look at it.

navaronbracke:
@vicajilau With the recent Kotlin migration, can you look into porting this fix into the new Kotlin sources?

vicajilau:
@navaronbracke this has been resolved as you comment. I attach evidence:
image

navaronbracke:
@vicajilau The returned path is probably not valid for the dart:io File API, and only works with Android's ContentResolver API?

vicajilau:
@navaronbracke Android does not expose the actual file path behind a `content://`, for security and storage encapsulation reasons (especially from Android 10 onwards with Scoped Storage). So our Android `Uri` won't work with a simple `File` of `dart:io` unless we move it to a temporary folder, which I'm not particularly happy about... I'm still analyzing alternatives

vicajilau:
@navaronbracke After analysis, the conclusion is that there are 3 options to solve this. 
1. Return the path to the temporary file instead of the written file even though it will not be the actual path to the file.
2. Return the filename without the full path.
3. Return the status of the operation without returning a path.

The disadvantage of working with or returning the temporary path is that we do not control or decide how long Android will decide that the file exists. For that reason, I think the API should aim to ensure that the `saveFile` does not return the path.

navaronbracke:
@daniJimen Is this still relevant with the recent fixes to saveFile() ?

daniJimen:
@navaronbracke No, this PR is no longer relevant, as this has already been handled; the location of the temporary file is currently being provided:
image
i think that,s the best form to express it, because you inform the real ubication of file

navaronbracke:
cc @vicajilau for a second opinion, since you listed 3 options in https://github.com/miguelpruivo/flutter_file_picker/pull/1712#issuecomment-2801530666

(Although I think we should close this)

vicajilau:
@navaronbracke  This no longer makes sense since we're implementing the first option I suggested.

> 1. Return the path to the temporary file instead of the written file even though it will not be the actual path to the file.

I'm agree, closing the PR

PR #1713, opened 2025-02-19
fix: Run `getDirectoryPath` in its own isolate to avoid COM initialization issues

tguerin:
I’m the original reporter of https://github.com/alnitak/flutter_soloud/issues/180. Even with the previous fix, we continued seeing the following error in production: 

> comdlg32.dll exception code 0xc000005

After further investigation, I discovered that the issue was related to COM initialization on the main thread. If COM is initialized on the main thread, we should **not** call `CoUninitialize`, as it may affect other components relying on COM.

To prevent this issue, the best approach is to run `getDirectoryPath` in its own isolate. This allows us to safely initialize and release COM within the isolate without impacting the main thread or other parts of the application.

Additionally, I added `finally` blocks to properly release allocated memory and ensure cleanup happens reliably.

This should resolve the issue while maintaining stability for other COM-dependent components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 9.0.1
+### Windows
+- Move `getDirectoryPath()` to its own isolate to avoid COM initialization conflicts with other plugins [#1713](https://github.com/miguelpruivo/flutter_file_picker/pull/1713) [@tguerin](https://github.com/tguerin). 
+
 ## 9.0.0
 ### Web
 - **BREAKING CHANGE:** `pickFiles()` now loads files as blobs. See the note in the updated [wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki/api#-pickfiles)
--- a/lib/src/windows/file_picker_windows.dart
+++ b/lib/src/windows/file_picker_windows.dart
@@ -5,9 +5,10 @@ import 'dart:typed_data';
 
 import 'package:ffi/ffi.dart';
 import 'package:file_picker/file_picker.dart';
-import 'package:file_picker/src/utils.dart';
 import 'package:file_picker/src/exceptions.dart';
+import 'package:file_picker/src/utils.dart';
 import 'package:file_picker/src/windows/file_picker_windows_ffi_types.dart';
+import 'package:flutter/foundation.dart';
 import 'package:path/path.dart';
 import 'package:win32/win32.dart';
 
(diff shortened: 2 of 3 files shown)


navaronbracke:
Thanks for the follow up! I'll put this on my review queue for tomorrow!

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
     String? initialDirectory,
-  }) {
+  }) async {
+    return compute(_getDirectoryPathIsolate, {
+      'dialogTitle': dialogTitle,
+      'initialDirectory': initialDirectory,
+    });
+  }
+
+  String? _getDirectoryPathIsolate(Map<String, dynamic> args) {
```
navaronbracke, line 105:
```suggestion
  String? _getDirectoryPathIsolate(Map<String,Object?> args) {
    String? dialogTitle = args['dialogTitle'] as String?;
    String? initialDirectory = args['initialDirectory'] as String?;
```

Super minor nit: I'd like to avoid `dynamic`, as it disables type-checking. Prefer `Object?` instead.

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
+
     int hr = CoInitializeEx(
       nullptr,
       COINIT.COINIT_APARTMENTTHREADED | COINIT.COINIT_DISABLE_OLE1DDE,
     );
 
-    // Ignore the error if COM is already initialized.
-    // If this is the case, CoInitializeEx will return
-    // RPC_E_CHANGED_MODE.
-    if (!SUCCEEDED(hr) && hr != RPC_E_CHANGED_MODE) {
```
navaronbracke, line 106: Just to confirm, the `hr != RPC_E_CHANGED_MODE` check is no longer needed, since we init COM in its own Isolate now?

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
+          throw WindowsException(hr);
+        }
 
-      if (hr == HRESULT_FROM_WIN32(WIN32_ERROR.ERROR_CANCELLED)) {
-        return Future.value(null);
+        final item = IShellItem(ppv.cast());
+        hr = fileDialog.setFolder(item.ptr.cast<Pointer<COMObject>>().value);
+        if (!SUCCEEDED(hr)) throw WindowsException(hr);
+      } finally {
+        free(riid);
```
navaronbracke, line 158: Nit: can we sort these `free()` calls  in the order that the pointers are declared?

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
+        free(folder);
+        free(ppv);
       }
-      throw WindowsException(hr);
     }
 
-    final ppsi = calloc<COMObject>();
-    hr = fileDialog.getResult(ppsi.cast());
-    if (!SUCCEEDED(hr)) throw WindowsException(hr);
+    hr = fileDialog.show(NULL);
```
navaronbracke, line 164: The `hwndOwner` part is gone? Can that be restored?

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
-      CoUninitialize();
+      try {
+        hr = SHCreateItemFromParsingName(folder, nullptr, riid, ppv);
+        if (!SUCCEEDED(hr) || ppv.value == nullptr) {
+          throw WindowsException(hr);
+        }
 
-      if (hr == HRESULT_FROM_WIN32(WIN32_ERROR.ERROR_CANCELLED)) {
-        return Future.value(null);
+        final item = IShellItem(ppv.cast());
```
navaronbracke, line 154: Do we need to release the shell item?

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
-    final path = pathPtr.value.toDartString();
+    final ppv = calloc<Pointer<COMObject>>();
+    String? selectedPath;
+    try {
+      hr = fileDialog.getResult(ppv);
+      if (!SUCCEEDED(hr)) throw WindowsException(hr);
 
-    hr = item.release();
-    if (!SUCCEEDED(hr)) throw WindowsException(hr);
+      final item = IShellItem(ppv.cast());
```
navaronbracke, line 173: Do we need to release the shell item?

navaronbracke (changes requested):
I don't know that much about the win32 API or COM, but I tried to review as best I could.

lib/src/windows/file_picker_windows.dart
tguerin: Yes the conflict was coming it was already initialized on the calling thread, with a new thread we don’t need this anymore.

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
+      if (!SUCCEEDED(hr)) throw WindowsException(hr);
+    } finally {
+      free(optionsPointer);
+    }
 
-    final title = TEXT(dialogTitle ?? defaultDialogTitle);
-    hr = fileDialog.setTitle(title);
-    if (!SUCCEEDED(hr)) throw WindowsException(hr);
-    free(title);
+    final title = TEXT(dialogTitle ?? 'Select a Folder');
```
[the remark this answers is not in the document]
tguerin: my bad will use the defaultTitle

lib/src/windows/file_picker_windows.dart
tguerin: i can constraint to String? for now as it’s the only type of value passed as parameters

lib/src/windows/file_picker_windows.dart
```diff
@@ -94,78 +95,94 @@ class FilePickerWindows extends FilePicker {
...
+      hr = item.getDisplayName(SIGDN.SIGDN_FILESYSPATH, pathPtr);
+      if (SUCCEEDED(hr)) {
+        selectedPath = pathPtr.value.toDartString();
+      }
+      free(pathPtr);
+    } finally {
+      free(ppv);
+    }
 
     CoUninitialize();
```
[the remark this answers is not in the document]
tguerin: You're right the first implementation was only releasing in the best case scenario i added a try finally after the successful initialization to properly clean the com.

lib/src/windows/file_picker_windows.dart
tguerin: That's a good question and honestly I'm not expert enough in this kind of code to give an accurated answer. I guess it’s not necessary as it's a wrapper around ppv which is the allocated pointer.

tguerin:
Once the PR is ready, please let me know so we can run the tests again on our Windows machines to ensure nothing else was unintentionally broken. It seems the issue was affecting only a few specific machines.

lib/src/windows/file_picker_windows.dart
navaronbracke: @tguerin This is not restored yet?

tguerin: ah yes my bad

navaronbracke (changes requested):
Just a minor typo. We will also need a changelog entry for version 9.0.1, then this can be automatically published upon merging.

tguerin:
Ok will test this version on our machines and add the entry if everything goes well.

navaronbracke:
Let me know how the test went, then we can continue with landing this :)

tguerin:
Everything is ok, no crash on our machines. I've added the entry in the changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 9.0.1
```
navaronbracke, line 1: This will also need a version bump in the pubspec :)

navaronbracke (changes requested):

navaronbracke (approved):

PR #1728, opened 2025-03-15
Added an illustrative code in README.md

Lapyl:
Added an illustrative code in README.md


Lapyl:
Added an illustrative code in README.md

navaronbracke:
We have an interactive example app that demonstrates usage of the plugin.
Putting stuff in the readme on "how to use the plugin" is therefore irrelevant.

PR #1729, opened 2025-03-16
Fix: Ensure `saveFile` writes bytes on desktop platforms

### LINKED ISSUE #1725: `saveFile` should have consistent behavior across platforms when `bytes` is provided
### Current Behavior: 
- On mobile, saveFile writes bytes to a file and returns the path.
- On desktop, it only shows a save dialog and ignores bytes.
- On macOS, passing bytes throws UnsupportedError.

### Issue:

This inconsistency forces platform-specific code. If saveFile wrote bytes to the selected path on desktop (as it does on mobile), developers wouldn't need to handle platforms differently.

### Proposal:

When `bytes` is provided on desktop:

- Open the save dialog.
- Write bytes to the selected file.
- Return the file path (matching mobile behavior).

Would it be possible to implement this for consistency? Are there technical limitations preventing it?

vicajilau:
## Changes Made
On **desktop platforms**, when `bytes` is provided:
  1. The save dialog opens as usual.
  2. If a path is selected, the provided `bytes` are written to the file.
  3. The file path is returned (matching mobile behavior).

## Why This Change?
This modification improves cross-platform consistency, allowing developers to use `saveFile` without needing platform-specific logic.

## Testing
- Verified the new behavior on **Windows, macOS, and Linux**.
- Ensured existing functionality on mobile remains unchanged.

### Related Issue
Fixes #1725

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
+## 9.2.0
+### Desktop (macOS, Windows, Linux)
+- `saveFile` now writes the provided `bytes` to the selected file, ensuring consistency with mobile behavior. [@vicajilau](https://github.com/vicajilau).
+
 ## 9.1.0
 ### Web
-- Added an implementation for `saveFile()` on the web.
+- Added an implementation for `saveFile()` on the web. [@vicajilau](https://github.com/vicajilau).
 
 ## 9.0.3
 ### Android
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 9.2.0
 ### Desktop (macOS, Windows, Linux)
-- `saveFile` now writes the provided `bytes` to the selected file, ensuring consistency with mobile behavior. [@vicajilau](https://github.com/vicajilau).
+- Fixes an inconsistency for saveFile that did not save the file, when bytes are provided on desktop platforms. [@vicajilau](https://github.com/vicajilau).
 
 ## 9.1.0
 ### Web
--- a/lib/src/file_picker.dart
+++ b/lib/src/file_picker.dart
@@ -154,7 +154,10 @@ abstract class FilePicker extends PlatformInterface {
   /// Opens a save file dialog which lets the user select a file path and a file
   /// name to save a file.
   ///
-  /// For mobile and desktop platforms, this function will save file with [bytes] to return a path.
+  /// For mobile, this function will save a file with the given [fileName] and [bytes] and return the path where the file was saved.
+  ///
+  /// For desktop platforms, this function opens a dialog to let the user choose a location for the file and returns the selected path.
+  /// If the bytes are provided, then the bytes are written to a file at the chosen path.
   ///
   /// On the web, this function will start a download for the file with [bytes] and [fileName].
   /// If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].
--- a/lib/src/file_picker.dart
+++ b/lib/src/file_picker.dart
@@ -154,15 +154,10 @@ abstract class FilePicker extends PlatformInterface {
   /// Opens a save file dialog which lets the user select a file path and a file
   /// name to save a file.
   ///
-  /// For mobile platforms, this function will save file with [bytes] to return a path.
-  /// Throws UnsupportedError on macOS if bytes are provided.
+  /// For mobile and desktop platforms, this function will save file with [bytes] to return a path.
   ///
   /// On the web, this function will start a download for the file with [bytes] and [fileName].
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
+## 9.2.0
+### Desktop (macOS, Windows, Linux)
+- `saveFile` now writes the provided `bytes` to the selected file, ensuring consistency with mobile behavior. [@vicajilau](https://github.com/vicajilau).
```
navaronbracke, line 3:
```suggestion
- Fixes an inconsistency for saveFile that did not save the file on desktop platforms. [@vicajilau](https://github.com/vicajilau).
```

lib/src/file_picker.dart
```diff
@@ -154,16 +154,11 @@ abstract class FilePicker extends PlatformInterface {
   /// Opens a save file dialog which lets the user select a file path and a file
   /// name to save a file.
   ///
-  /// For mobile platforms, this function will save file with [bytes] to return a path.
-  /// Throws UnsupportedError on macOS if bytes are provided.
+  /// For mobile and desktop platforms, this function will save file with [bytes] to return a path.
```
navaronbracke, line 157:
```suggestion
  /// On mobile platforms, this function will save a file with the given [fileName] and [bytes] and return the path where the file was saved.
```

lib/src/utils.dart
```diff
@@ -64,6 +64,13 @@ Future<String> isExecutableOnPath(String executable) async {
   return path;
 }
 
+void saveBytesSyncToFile(Uint8List? bytes, String? path) {
```
navaronbracke, line 67: There are a few small issues with this function, I think.

1. We do not use `existsSync()` to check if the file already exists.
2. We should be using `writeAsBytes()` instead of the sync variant. The bytes could be potentially very large, resulting in UI jank otherwise.

navaronbracke (changes requested):
Some minor suggestions.

Overall this looks okay, _but_ we now introduce another bimodal behavior in the API:
desktop opening a file chooser and optionally also saving the file depending on the bytes.

I think it is time we think about addressing https://github.com/miguelpruivo/flutter_file_picker/issues/1469

I will start thinking on a possible API surface for version 10.0.0 to see if this can be improved.

lib/src/utils.dart
vicajilau: Resolved and agreed on the second point, I love it. But on the first one, I would like to understand why we want to check if the file already exists, wouldn't it be better to overwrite the file in that case? From my point of view, it doesn't matter if the file exists or not, in any case we will write it new or overwrite it. Isn't?

navaronbracke: I wouldn't want the plugin to overwrite the user's files, that could be very disruptive.
I would just throw an ArgumentError if the file at the given path exists.

Although, in practice I don't know if the file picker window internally checks if a file at the given path already exists and disallows selecting such a path in the first place?

vicajilau: Currently Windows already manages whether you want to overwrite files.

image

vicajilau: And it happens the same behaviour on Linux 
image

navaronbracke: What happens if you tell it to replace? It deletes the original file and then returns the path for the now-nonexistent file?
And what about MacOS ?

vicajilau: If you tell it to replace the file, it doesn't delete the original file, it just returns the path and in our new code (if we don't detect whether the file exists or not) it will rewrite the file. So I don't think it's invasive if the user has already been able to accept the rewrite.

navaronbracke: Does it return the exact same file path? (or a file path with `_1` at the end) And what decides that the given path can be rewritten? Is the dart:io File API allowed to write to said file path if you select "replace" ?

(I think that allowing it is probably fine, I just want some context on how it works, so I can understand it better)

vicajilau: It returns exactly the same path.

One issue I just detected in MacOS (I haven't checked the rest of the OS) is that the `pickFiles` method is not reading bytes making harder to test this functionality. But I'm not sure if I should check it and include it in this PR or do it in a new separate issue.

navaronbracke: > It returns exactly the same path.

Interesting. As for the issue with the bytes, you can do that separately.

navaronbracke (approved):

stephane-archer:
Wow amazing change, thank you. Is it live on pub? You just are the best

navaronbracke:
> Wow amazing change, thank you. Is it live on pub? You just are the best

It is available on Pub.dev in version 9.2.0

PR #1733, opened 2025-03-18
Some lint improvements

vicajilau:
This is a minor changes improving some lint warnings

### DIFF
--- a/lib/src/platform_file.dart
+++ b/lib/src/platform_file.dart
@@ -1,6 +1,4 @@
 import 'dart:async';
-// ignore: unnecessary_import
-import 'dart:typed_data';
 
 import 'package:cross_file/cross_file.dart';
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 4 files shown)


lib/src/platform_file.dart
```diff
@@ -31,7 +29,7 @@ class PlatformFile {
   /// ```
   /// final File myFile = File(platformFile.path);
   /// ```
-  /// On web the path points to a Blob URL, if present, which can be cleaned up using [URL.revokeObjectURL].
+  /// On web the path points to a Blob URL, if present, which can be cleaned up using `URL.revokeObjectURL`.
```
navaronbracke, line 32: Can we use a `/// @docImport` for package:web ? (I do not recall in what language version we have docImports)

navaronbracke (changes requested):
Minor nit about a potential doc only import

lib/src/platform_file.dart
vicajilau: `/// @docImport` doesn't work. I found another way to do this :)

navaronbracke (approved):

Issue #1718, opened 2025-02-24
Question about `lockParentWindow` behavior on MacOS


AdamVe:
Hi! In the recent PR #1690 where the `lockParentWindow` functionality was fixed for MacOS #1674, I noticed that the entire app freezes when picking files. For example, animations on the material buttons stop working during this process.

Could you please confirm if this behavior is intended?

https://github.com/user-attachments/assets/305174b2-5022-499c-b633-1a67ea08b45e

navaronbracke:
Well, the NSOpenPanel is a new Window, so presumably your app itself gets paused? Can you double check this with the `didChangeAppLifecycleState()` method from a WidgetsBindingObserver ?

orestesgaolin:
Is that happening on Flutter 3.29? I think this may be related to platform thread merging?

AdamVe:
Thanks for your responses. As @orestesgaolin mentions, native apps shadow the parent window.
I did some tests with other applications and have the following example:

- In the Chrome browser, I loaded a page with animated content.
- I pressed Cmd+S to execute the Save action.
- The file picker for Save opened.
- The parent window was darkened.
- The animated content was still animating in the parent window.
- From my experience, this behavior is common across macOS applications.

https://github.com/user-attachments/assets/fc145c64-6dc2-4e10-bc4c-377c65286865

PR #1734, opened 2025-03-18
Use sheet modal for panels on macOs

AdamVe:
In #1718, I described that on macOS, the file_picker panels pause the parent app's run loop. I did some investigation and found out that using a sheet modal to present the file_picker panels improves the user experience - opening and saving files is similar to what users see in other apps, and the run loop in the parent window is still running.

This PR adds the change needed to use the sheet modal - please let me know if this is something we could eventually have in file_picker.

Notes:
- With this change, the file_picker panels lose the dialog title.
- Documentation for beginSheetModal: https://developer.apple.com/documentation/appkit/nssavepanel/beginsheetmodal(for:completionhandler:)

The recordings show the different behavior.

Before change:
https://github.com/user-attachments/assets/6e333148-8bb0-449d-bcec-f3e46e8e8c64

After change:
https://github.com/user-attachments/assets/74724d8a-a02d-4c45-81cf-d5216d11b64a

Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1718

### DIFF
--- a/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
+++ b/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
@@ -75,6 +75,7 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
             result([pathResult!.path])
           }
         }
+        return
       } else {
         // User dismissed the dialog
         result(nil)
@@ -87,7 +88,7 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
   ) {
     let dialog = NSOpenPanel()
     let args = call.arguments as! [String: Any]
-    
+
     dialog.directoryURL = URL(
       fileURLWithPath: args["initialDirectory"] as? String ?? ""
     )
@@ -104,11 +105,11 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
       if (response == .OK) {
         if let url = dialog.url {
           result(url.path)
+          return
         }
-      } else {
-        // User dismissed the dialog
-        result(nil)
       }
+      // User dismissed the dialog
+      result(nil)
     }
   }
 
@@ -141,11 +142,11 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
       if (response == .OK) {
         if let url = dialog.url {
           result(url.path)
+          return
         }
-      } else {
-        // User dismissed the dialog
-        result(nil)
       }
+      // User dismissed the dialog
+      result(nil)
     }
   }
 
--- a/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
+++ b/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
@@ -7,7 +7,7 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
     let channel = FlutterMethodChannel(
       name: "miguelruivo.flutter.plugins.filepicker",
       binaryMessenger: registrar.messenger)
-    let instance = FilePickerPlugin(registrar:registrar)
+    let instance = FilePickerPlugin(registrar: registrar)
     registrar.addMethodCallDelegate(instance, channel: channel)
   }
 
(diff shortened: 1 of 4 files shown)


macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -7,10 +7,17 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
     let channel = FlutterMethodChannel(
       name: "miguelruivo.flutter.plugins.filepicker",
       binaryMessenger: registrar.messenger)
-    let instance = FilePickerPlugin()
+    let instance = FilePickerPlugin(registrar:registrar)
```
navaronbracke, line 10:
```suggestion
    let instance = FilePickerPlugin(registrar: registrar)
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -45,28 +52,32 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
     let extensions = args["allowedExtensions"] as? [String] ?? []
     applyExtensions(dialog, extensions)
 
-    if dialog.runModal() == NSApplication.ModalResponse.OK {
-      if allowMultiple {
-        let pathResult = dialog.urls
-        if pathResult.isEmpty {
-          result(nil)
+    if let appWindow = getFlutterWindow() {
```
navaronbracke, line 55: Can we use `guard let` for the `getFlutterWindow()` calls, with a `result(nil)` ? Then we can keep the diff pretty and the code will be more readable :)

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -45,28 +52,32 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
...
+      dialog.beginSheetModal(for:appWindow) { response in
```
navaronbracke, line 56:
```suggestion
      dialog.beginSheetModal(for: appWindow) { response in
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -45,28 +52,32 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
...
+        if (response == .OK) {
```
navaronbracke, line 57: There's some oddities with the control flow for the picker windows, due to not using `return` along with a `result(nil)` for unreachable cases. Can you verify that each branch calls the `result` block, so that we do not end up in situations where the result is never called? (that would cause the method call to hang on the Dart side)

I think that refactoring the handlers to use early returns instead of if-else, along with using `result(nil)` as a fallback should fix this. This is mostly due to the old implementation being written like that, though.
The new window let statement just makes it a bit harder to review in the diff :P

navaronbracke (changes requested):
Overall this change is LGTM, just some questions about control flow & formatting.

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
AdamVe: I get what you mean! Now every execution branch gets the `result()` call. Sadly, using `guard let` did not help the PR readability.

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -45,28 +52,34 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
...
-        let pathResult = dialog.urls
-        if pathResult.isEmpty {
-          result(nil)
+    guard let appWindow = getFlutterWindow() else {
+      result(nil)
+      return
+    }
+
+    dialog.beginSheetModal(for: appWindow) { response in
+      if (response == .OK) {
```
navaronbracke, line 61: This block is a bit heavy in nesting.

Can we restructure it as follows?

```
dialog.beginSheetModal(for: appWindow) { response in
   // User dismissed the dialog
   if (response != .OK) {
     result(nil)
     return
   }
   
   if allowMultiple {
     let pathResult = dialog.urls
     
     if pathResult.isEmpty {
       result(nil)
     } else {
       let paths = pathResult.map { $0.path }
       result(paths)
     }
     
     return
   }
   
   let pathResult = dialog.url
   if pathResult == nil {
     result(nil)
     return
   }
   
   result([pathResult!.path])
}
```

navaronbracke (changes requested):
Minor readability suggestion, but otherwise LGTM.

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
navaronbracke: See my new comment for the readability thing :)

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
AdamVe: That is much nicer yes. I update all three block to have the same style.

navaronbracke (changes requested):
The Swift code is LGTM.

I had a look at all three methods for updating the documentation:
The `pickFiles` and `getDirectoryPath` methods already list MacOS as not supporting the dialog title:
https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/file_picker.dart#L72-L73
https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/file_picker.dart#L125-L126

So you will have to update the documentation for `saveFile` only.
Can you add `/// Not supported on macOS.` under this line?

https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/file_picker.dart#L167

This is also going to need a changelog entry and patch version bump, so that we can release this upon merging :)

navaronbracke:
Updated the wording of the OP, so that Github will auto close the issue when this is merged.
Once the checks are green, this can land. Thank you!

AdamVe:
Thank you @navaronbracke

orestesgaolin:
This is so cool, congrats on landing the change. Imho this is so much more clear now that the file is being picked.

PR #1736, opened 2025-03-19
11

ishaquehassan:


navaronbracke:
This feels more like a workaround, rather than an actual fix? Instead of using a fake dialog, I would look into using a more long term solution.

navaronbracke:
Maybe a long term solution is just to have the input element invisible? (instead of `display="none"`)

The input element is doing that here: https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/_internal/file_picker_web.dart#L67

See https://stackoverflow.com/questions/26501221/calling-click-event-of-input-doesnt-work-in-safari

ishaquehassan:
This will not work already tried it. @navaronbracke

navaronbracke:
I'm going to close this pull request, due to several reasons.

Firstly it is using old HTML bindings, it also adds unnecessary formatting changes? And I don't know if this restriction still applies to modern Safari. Besides, I think the same limitation exists for autoplay on video elements, so I don't think that the "hacky workaround" with a dialog is the right thing to do, not to mention it is untranslated, which breaks internationalization for all our end users.

Issue #1735, opened 2025-03-18
saveFile returns null on web platform


aleripe:
**Describe the bug**
`saveFile` returns null on web platform, despite downloading the file properly.

**Platform**
- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Desktop

**Platform OS version**
Windows 10, Chrome browser

**How are you picking?**
```
return await FilePicker.platform.saveFile(
  fileName: name,
  bytes: data,
);
```

**Flutter Version details**
[√] Flutter (Channel stable, 3.29.2, on Microsoft Windows [Versione 10.0.19045.5608], locale it-IT) [791ms]
    • Flutter version 3.29.2 on channel stable at C:\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision c236373904 (5 days ago), 2025-03-13 16:17:06 -0400
    • Engine revision 18b71d647a
    • Dart version 3.7.2
    • DevTools version 2.42.3

[√] Windows Version (10 Pro 64 bit, 22H2, 2009) [3,3s]

[√] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [4,5s]
    • Android SDK at C:\Users\Alessandro\AppData\Local\Android\sdk
    • Platform android-35, build-tools 35.0.0
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.5+-12932927-b750.29)
    • All Android licenses accepted.

[√] Chrome - develop for the web [361ms]
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[!] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.12.5) [360ms]
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.12.35728.132
    X Visual Studio is missing necessary components. Please re-run the Visual Studio installer for the "Desktop development with C++" workload, and include these components:      
        MSVC v142 - VS 2019 C++ x64/x86 build tools
         - If there are multiple build tool versions available, install the latest
        C++ CMake tools for Windows
        Windows 10 SDK

[√] Android Studio (version 2024.3) [31ms]
[...]

PR #1737, opened 2025-03-19
improved documentation on saveFile

vicajilau:
Improved documentation https://github.com/miguelpruivo/flutter_file_picker/issues/1735#issuecomment-2734756153

### DIFF
--- a/lib/src/file_picker.dart
+++ b/lib/src/file_picker.dart
@@ -161,8 +161,6 @@ abstract class FilePicker extends PlatformInterface {
   ///
   /// On the web, this function will start a download for the file with [bytes] and [fileName].
   /// If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].
-  /// On the web, this function will start a download for the file with [bytes] and [fileName].
-  /// If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].
   /// The returned path for the downloaded file will always be `null`.
   /// provide access to the file system paths. The temporary blob URL used for the download
   /// is revoked after the download starts to conserve memory.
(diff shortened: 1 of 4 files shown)


lib/src/file_picker.dart
```diff
@@ -161,6 +161,9 @@ abstract class FilePicker extends PlatformInterface {
   ///
   /// On the web, this function will start a download for the file with [bytes] and [fileName].
   /// If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].
+  /// Note: [saveFile] will always return null on the web. This is because the web does not
```
navaronbracke, line 164: Whatever is done behind the scenes for the Blob url is an implementation detail and can be omitted.

```suggestion
  /// On the web, this function will start a download for the file with [bytes] and [fileName].
  /// If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].
  /// The returned path for the downloaded file will always be `null`.
```

lib/src/file_picker.dart
```diff
@@ -177,7 +180,7 @@ abstract class FilePicker extends PlatformInterface {
   /// will be used.
   ///
   /// The file type filter [type] defaults to [FileType.any]. Optionally,
-  /// [allowedExtensions] might be provided (e.g. `[pdf, svg, jpg]`.). Both
+  /// [allowedExtensions] might be provided (e.g. `[pdf, svg, jpg]`)ccd. Both
```
navaronbracke, line 183: This is a typo?

```suggestion
  /// [allowedExtensions] might be provided (e.g. `[pdf, svg, jpg]`). Both
```

navaronbracke (changes requested):

navaronbracke:
@vicajilau Could you do a rebase? https://github.com/miguelpruivo/flutter_file_picker/pull/1734 already added a changelog entry for 9.2.1 & version update, so that should not be in this diff.

vicajilau:
Damn, I made my changes in the wrong branch...

navaronbracke:
@vicajilau Do we need to do a changelog update? I'm fine with publishing this with the next fix though, since it is very minor.

navaronbracke (approved):

vicajilau:
I'm agree with you :)

PR #1741, opened 2025-03-26
NSSavePanel Not Appearing Due to Missing Entitlements Handling

vicajilau:
**Description:**
This PR addresses an issue where the `NSSavePanel` was not appearing when invoking the file picker on macOS. The root cause was an entitlement issue that was not properly handled, making it unclear why the panel failed to initialize. Now, the plugin correctly checks for the required entitlements (`com.apple.security.files.user-selected.read-write` or `com.apple.security.files.user-selected.read-only`) and returns an appropriate error if they are missing.

**Changes:**
- Implemented `checkEntitlement(requiredMode: EntitlementMode, result: @escaping FlutterResult)` to verify if the app has the correct entitlements before invoking `NSSavePanel`.
- Ensured that `NSSavePanel` is only initialized if the required entitlements are available.
- If entitlements are missing, the plugin now returns a `FlutterError` with a clear message instead of failing silently.

**How to Test:**
1. Run the file picker in an app with missing entitlements.
2. Observe that a clear error is returned instead of the panel failing to appear.
3. Run the file picker in an app with the correct entitlements.
4. Confirm that `NSSavePanel` appears as expected.

**Issue Reference:**
This fixes an issue where `NSSavePanel` would silently fail due to entitlement restrictions, making debugging difficult.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 9.2.2
+### Desktop (macOS)
+- Returns a clear FlutterError if the required entitlements are missing, instead of failing silently.
+
 ## 9.2.1
 ### Desktop (macOS)
 - Present file picker panel as a sheet modal to the Flutter application window. [#1734](https://github.com/miguelpruivo/flutter_file_picker/pull/1734)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 9.2.2
 ### Desktop (macOS)
-- Returns a clear FlutterError if the required entitlements are missing, instead of failing silently.
+- Updated the file picker to check for missing entitlements, instead of failing silently.
 
 ## 9.2.1
 ### Desktop (macOS)
--- a/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
+++ b/macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
@@ -2,181 +2,213 @@ import Cocoa
 import FlutterMacOS
 import UniformTypeIdentifiers
 
+enum EntitlementMode {
+    case requireWrite
+    case readOrWrite
+}
+
 public class FilePickerPlugin: NSObject, FlutterPlugin {
-  public static func register(with registrar: FlutterPluginRegistrar) {
-    let channel = FlutterMethodChannel(
-      name: "miguelruivo.flutter.plugins.filepicker",
-      binaryMessenger: registrar.messenger)
-    let instance = FilePickerPlugin(registrar: registrar)
-    registrar.addMethodCallDelegate(instance, channel: channel)
-  }
-
-  private let registrar: FlutterPluginRegistrar
-
-  init(registrar: FlutterPluginRegistrar) {
-    self.registrar = registrar
-    super.init()
-  }
-
-  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult)
-  {
-    switch call.method {
-    case "pickFiles":
-      handleFileSelection(call, result: result)
-
-    case "getDirectoryPath":
-      handleDirectorySelection(call, result: result)
-
-    case "saveFile":
-      handleSaveFile(call, result: result)
-
-    default:
-      result(FlutterMethodNotImplemented)
+    public static func register(with registrar: FlutterPluginRegistrar) {
+        let channel = FlutterMethodChannel(
+            name: "miguelruivo.flutter.plugins.filepicker",
+            binaryMessenger: registrar.messenger)
+        let instance = FilePickerPlugin(registrar: registrar)
+        registrar.addMethodCallDelegate(instance, channel: channel)
     }
-  }
-
-  private func handleFileSelection(
-    _ call: FlutterMethodCall, result: @escaping FlutterResult
-  ) {
-    let dialog = NSOpenPanel()
-    let args = call.arguments as! [String: Any]
-
-    dialog.directoryURL = URL(
-      fileURLWithPath: args["initialDirectory"] as? String ?? ""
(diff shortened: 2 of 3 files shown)


macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -2,181 +2,213 @@ import Cocoa
...
-      }
-
-      if let url = dialog.url {
-        result(url.path)
-        return
-      }
-
-      result(nil)
+    private func handleFileSelection(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        if checkEntitlement(requiredMode: .readOrWrite, result: result) {
```
navaronbracke, line 43: Can we update `checkEntitlement` to return a `FlutterError?` and if it is not nil, call the method call result with it? I'd rather not have to pass the method call result to this method.

I.e.

```
if let entitlementError = checkEntitlement(requiredMode: .readOrWrite) {
  result(entitlementError)
  return
}

// ...
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 9.2.2
+### Desktop (macOS)
+- Returns a clear FlutterError if the required entitlements are missing, instead of failing silently.
```
navaronbracke, line 3:
```suggestion
- Updated the file picker to check for missing entitlements, instead of failing silently.
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -2,181 +2,213 @@ import Cocoa
...
-        result(url.path)
-        return
-      }
-
-      result(nil)
+    
+    /// Checks if the  entitlements file contains the required entitlement for save files.
+    private func checkEntitlement(requiredMode: EntitlementMode, result: @escaping FlutterResult) -> Bool {
+        guard let task = SecTaskCreateFromSelf(nil) else {
+            result(FlutterError(code: "ENTITLEMENT_CHECK_FAILED", message: "Failed to create security task.", details: nil))
```
navaronbracke, line 172: Can we update the message here?

```suggestion
FlutterError(code: "ENTITLEMENT_CHECK_FAILED", message: "Failed to verify file_picker entitlements.", details: nil)
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -2,181 +2,213 @@ import Cocoa
...
+            return false
+        }
+        
+        let readWriteEntitlement = SecTaskCopyValueForEntitlement(task, "com.apple.security.files.user-selected.read-write" as CFString, nil) as? Bool
```
navaronbracke, line 176: There are no constants for these CFStrings ? Maybe we can use private `static let` for these?

navaronbracke, line 176: https://developer.apple.com/documentation/security/sectaskcopyvalueforentitlement(_:_:_:) does not return a `Bool?`. Should we check against nil instead?

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -2,181 +2,213 @@ import Cocoa
...
+        let readOnlyEntitlement = SecTaskCopyValueForEntitlement(task, "com.apple.security.files.user-selected.read-only" as CFString, nil) as? Bool
+        
+        switch requiredMode {
+        case .requireWrite:
+            if readWriteEntitlement != true {
+                result(FlutterError(code: "ENTITLEMENT_REQUIRED_WRITE", message: "Read-write entitlement is required but not found.", details: nil))
```
navaronbracke, line 182:
```suggestion
FlutterError(code: "ENTITLEMENT_REQUIRED_WRITE", message: "The Read-Write entitlement is required for this action.", details: nil)
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -2,181 +2,213 @@ import Cocoa
...
+                return false
+            }
+            
+        case .readOrWrite:
+            if readWriteEntitlement != true && readOnlyEntitlement != true {
+                result(FlutterError(code: "ENTITLEMENT_NOT_FOUND", message: "Neither read-write nor read-only entitlements found.", details: nil))
```
navaronbracke, line 188:
```suggestion
FlutterError(code: "ENTITLEMENT_NOT_FOUND", message: "Either the Read-Only or Read-Write entitlement is required for this action.", details: nil)
```

navaronbracke (changes requested):
Some suggestions around the entitlement code.

The rest seems to be a formatting diff? (reviewed the change without whitespace for my own sanity :P )

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
vicajilau: Changed inside of a big commit

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
vicajilau: Yes, fixed!

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
vicajilau: Added in another commit!

vicajilau:
Resolved all the points raised 🚀

navaronbracke (approved):
TIL about safe casts from Any? with `as?`

Issue #1743, opened 2025-03-27
Bug: Image duplication in "Pictures" folder on Android


daniJimen:
**Description**
When using the FilePicker class to select images on an Android device, I noticed that the selected images are being duplicated in the Pictures folder. This behavior should not occur, as the user is only selecting an existing image, not copying or moving it.

**Steps to reproduce**
Use the following code to pick files:

```dart
FilePicker.platform.pickFiles(
   type: FileType.custom,
   allowedExtensions: ['pdf', 'jpg', 'png'],
   allowMultiple: true,
);
```
Select an image from the gallery.

Check the Pictures folder on the device.

The selected image appears duplicated in that folder.

**Environment**
- Device: Xiaomi 12T
- Android version: 13
- Flutter version: 3.29.2
- file_picker version: 9.2.2

**Additional notes**
This issue might be related to how the plugin handles temporary files or media storage on Android.

**Additional context**
I think from my point of view, that compression should not be applied when the user does not set a compression level, since, as I checked in the library code, by default 30% compression is being applied, and this makes, in the createImage function that you use when you access the compressImage function, duplicate the image in the pictures folder.

That issue its reported by one user also in stackoverflow: https://stackoverflow.com/questions/78689324/problem-with-file-picker-duplicate-image-creation

vicajilau:
In my opinion we should not compress the image [by default](https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java#L45) , avoiding the problem of image duplication. I have identified this on all platforms.

What do you think @navaronbracke?

navaronbracke:
@vicajilau I do agree we should have `allowCompression` default to false.

Issue #1742, opened 2025-03-26
Android 11 or less - permission denied exception when compressionQuality not 0


FilipObornik:
When I call `FilePicker.platform.pickFiles(type: fileType, allowMultiple: allowMultiple)`, i get exception. It happens only in Android 11 or less. I was able to reproduce it with emulator.

Possible hotfix is to set compressionQuality to 0 - `FilePicker.platform.pickFiles(type: fileType, allowMultiple: allowMultiple, compressionQuality: 0)`

**Platform**
- [x] Android

**Platform OS version**
Android 11 and below

**Details to reproduce the issue**
Provide all the details to reproduce the issue.

**Error Log**
```
E/AndroidRuntime( 9795): FATAL EXCEPTION: Thread-9
E/AndroidRuntime( 9795): Process: com.tymbe.ty_mobil_brigady.stage, PID: 9795
E/AndroidRuntime( 9795): java.lang.RuntimeException: java.io.IOException: Permission denied
E/AndroidRuntime( 9795):        at com.mr.flutter.plugin.filepicker.FileUtils.compressImage(FileUtils.java:111)
E/AndroidRuntime( 9795):        at com.mr.flutter.plugin.filepicker.FilePickerDelegate$1.run(FilePickerDelegate.java:125)
E/AndroidRuntime( 9795):        at java.lang.Thread.run(Thread.java:919)
E/AndroidRuntime( 9795): Caused by: java.io.IOException: Permission denied
E/AndroidRuntime( 9795):        at java.io.UnixFileSystem.createFileExclusively0(Native Method)
E/AndroidRuntime( 9795):        at java.io.UnixFileSystem.createFileExclusively(UnixFileSystem.java:317)
E/AndroidRuntime( 9795):        at java.io.File.createTempFile(File.java:2018)
E/AndroidRuntime( 9795):        at com.mr.flutter.plugin.filepicker.FileUtils.createImageFile(FileUtils.java:120)
E/AndroidRuntime( 9795):        at com.mr.flutter.plugin.filepicker.FileUtils.compressImage(FileUtils.java:98)
E/AndroidRuntime( 9795):        ... 2 more
```

**Screenshots and/or video**
If applicable, add screenshots or video to help explain your problem.

**Flutter Version details**
Tested on 3.27.3 and 3.29.0

**Additional context**
Seem to be related to closed issue https://github.com/miguelpruivo/flutter_file_picker/issues/1534

[...]

PR #1744, opened 2025-03-27
Changed compressionQuality by default to 0

vicajilau:
Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1742
Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1743

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - **BREAKING CHANGE:** The `allowCompression` property has been deprecated in favor of `compressionQuality`, and now defaults to `false`.
 
 ### Android
-- Fixed a permission denied exception Android 11 or lower when `compressionQuality` is not `0`. ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)). [@vicajilau](https://github.com/vicajilau)
+- Fixed a permission denied exception on Android 11 or lower when `compressionQuality` is not `0`. ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)). [@vicajilau](https://github.com/vicajilau)
 - Fixed an issue where images were duplicated to the "Pictures" folder on Android. ([#1743](https://github.com/miguelpruivo/flutter_file_picker/issues/1743)) [@vicajilau](https://github.com/vicajilau)
 
 ## 9.2.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 ## 10.0.0
 ### General
-- The `compressionQuality` property in the `pickFiles` method now defaults to `0`, and `allowCompression` is set to `false`. [@vicajilau](https://github.com/vicajilau).
-- The `allowCompression` property has been deprecated in favor of `compressionQuality`. [@vicajilau](https://github.com/vicajilau).
+- **BREAKING CHANGE:** The `compressionQuality` property in the `pickFiles` method now defaults to `0`.
+- **BREAKING CHANGE:** The `allowCompression` property has been deprecated in favor of `compressionQuality`, and now defaults to `false`.
+
 ### Android
-- **Fixed:** *Permission denied* exception on Android 11 or lower when `compressionQuality` is not `0` ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)).
-- **Fixed:** Image duplication issue in the "Pictures" folder on Android ([#1743](https://github.com/miguelpruivo/flutter_file_picker/issues/1743)).
+- Fixed a permission denied exception Android 11 or lower when `compressionQuality` is not `0`. ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)). [@vicajilau](https://github.com/vicajilau)
+- Fixed an issue where images were duplicated to the "Pictures" folder on Android. ([#1743](https://github.com/miguelpruivo/flutter_file_picker/issues/1743)) [@vicajilau](https://github.com/vicajilau)
 
 ## 9.2.3
 ### Desktop (macOS)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## 10.0.0
 ### General
-
 - **BREAKING CHANGE:** The `compressionQuality` property in the `pickFiles` method now defaults to `0`.
 - **BREAKING CHANGE:** The `allowCompression` property has been deprecated in favor of `compressionQuality`, and now defaults to `false`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
-## 9.3.0
+## 10.0.0
+### General
+
+- **BREAKING CHANGE:** The `compressionQuality` property in the `pickFiles` method now defaults to `0`.
+- **BREAKING CHANGE:** The `allowCompression` property has been deprecated in favor of `compressionQuality`, and now defaults to `false`.
+
+### Android
+- Fixed a permission denied exception Android 11 or lower when `compressionQuality` is not `0`. ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)). [@vicajilau](https://github.com/vicajilau)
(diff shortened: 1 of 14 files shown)


android/src/main/java/com/mr/flutter/plugin/filepicker/FilePickerDelegate.java
```diff
@@ -42,7 +42,7 @@ public class FilePickerDelegate implements PluginRegistry.ActivityResultListener
     private boolean loadDataToMemory = false;
     private String type;
     private boolean allowCompression = true;
```
navaronbracke, line 44: Is it really the `compressionQuality` that dictates whether we compress images?
The `private boolean allowCompression = true;` line seems like the culprit instead?

Like, if we have `allowCompression` set to false, then we shouldn't be looking at the `compressionQuality` at all.
Can we leave the compressionQuality as-is, but change the default `allowCompression` to false instead?

navaronbracke (changes requested):
@vicajilau Does this fix the compression quality bug?

Maybe we should just have `allowCompression` be deprecated (and removed in the Java class).

I think we can look at the value of `compressionQuality` to determine if we should compress the image? Like, if compressionQuality is 0 (which would be the new default), we do not compress.

daniJimen:
@vicajilau @navaronbracke many thanks for your fast resolution, i appreciate that, for me the resolution its perfect! i would add only one think about that and its instead create the compressed image into pictures folder (when the user set any compression level), create the image into folder app to not create a lot of images into a shared folder

vicajilau:
We have to keep in mind that only Android uses these two properties (`allowCompression ` and `compressionQuality `). Compression capability is ignored on all other platforms.

lib/src/file_picker_io.dart
```diff
@@ -93,7 +95,7 @@ class FilePickerIO extends FilePicker {
       _eventSubscription?.cancel();
       if (onFileLoading != null) {
         _eventSubscription = _eventChannel.receiveBroadcastStream().listen(
-              (data) => onFileLoading((data as bool)
+              (data) => onFileLoading((data is bool)
```
navaronbracke, line 98: Nice catch!

ios/file_picker/Sources/file_picker/ImageUtils.m
```diff
@@ -17,7 +17,7 @@ + (BOOL)hasAlpha:(UIImage *)image {
             alpha == kCGImageAlphaPremultipliedFirst || alpha == kCGImageAlphaPremultipliedLast);
 }
 
-// Save the image temporarly in the app's tmp directory
+// Save the image temporally in the app's tmp directory
```
navaronbracke, line 20:
```suggestion
// Temporarily save the image in the app's tmp directory.
```

ios/file_picker/Sources/file_picker/FilePickerPlugin.m
```diff
@@ -114,8 +115,9 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
     
     NSDictionary * arguments = call.arguments;
     BOOL isMultiplePick = ((NSNumber*)[arguments valueForKey:@"allowMultipleSelection"]).boolValue;
-    
-    self.allowCompression = ((NSNumber*)[arguments valueForKey:@"allowCompression"]).boolValue;
+
+    self.compressionQuality = [[arguments valueForKey:@"compressionQuality"] intValue];
```
navaronbracke, line 119: We can just use a local instead of a member field?

```suggestion
    int compressionQuality = [[arguments valueForKey:@"compressionQuality"] intValue];
```

android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
```diff
@@ -113,10 +113,10 @@ public static Uri compressImage(Uri originalImageUri, int compressionQuality, Co
         return compressedUri;
     }
 
-    private static File createImageFile() throws IOException {
-        String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss").format(new Date());
+    private static File createImageFile(Context context) throws IOException {
+        String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(new Date());
         String imageFileName = "JPEG_" + timeStamp + "_";
-        File storageDir = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES);
+        File storageDir = context.getCacheDir();
```
navaronbracke, line 119: This change is to fix https://github.com/miguelpruivo/flutter_file_picker/issues/1743 ?

android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
```diff
@@ -113,10 +113,10 @@ public static Uri compressImage(Uri originalImageUri, int compressionQuality, Co
         return compressedUri;
     }
 
-    private static File createImageFile() throws IOException {
-        String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss").format(new Date());
+    private static File createImageFile(Context context) throws IOException {
+        String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(new Date());
```
navaronbracke, line 117: Good catch for the locale!

navaronbracke (changes requested):
Some small remarks, but otherwise LGTM.

Sadly changing the default compression quality could be seen as a breaking change? (Even though it is used to fix a bug)

android/src/main/java/com/mr/flutter/plugin/filepicker/FileUtils.java
vicajilau: YES!! And this will align android and iOS flows :)

vicajilau:
Everything ready @navaronbracke 🚀

vicajilau:
Conflicts resolved @navaronbracke

CHANGELOG.md
```diff
@@ -1,10 +1,18 @@
+## 9.3.0
```
navaronbracke, line 1:
```suggestion
## 10.0.0
### General

- **BREAKING CHANGE:** The `compressionQuality` property in the `pickFiles` method now defaults to `0`.
- **BREAKING CHANGE:** The `allowCompression` property has been deprecated in favor of `compressionQuality`, and now defaults to `false`.

### Android
- Fixed a permission denied exception Android 11 or lower when `compressionQuality` is not `0`. ([#1742](https://github.com/miguelpruivo/flutter_file_picker/issues/1742)). [@vicajilau](https://github.com/vicajilau)
- Fixed an issue where images were duplicated to the "Pictures" folder on Android. ([#1743](https://github.com/miguelpruivo/flutter_file_picker/issues/1743)) [@vicajilau](https://github.com/vicajilau)
```

pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: A package that allows you to use a native file explorer to pick sin
 homepage: https://github.com/miguelpruivo/plugins_flutter_file_picker
 repository: https://github.com/miguelpruivo/flutter_file_picker
 issue_tracker: https://github.com/miguelpruivo/flutter_file_picker/issues
-version: 9.2.3
+version: 9.3.0
```
navaronbracke, line 6:
```suggestion
version: 10.0.0
```

navaronbracke (changes requested):
We need to update the major version, since this changes the default value of the `compressionQuality`, so people might be surprised if the compression breaks when updating. (going from a default of 30 to 0)

CHANGELOG.md
```diff
@@ -1,10 +1,18 @@
+## 10.0.0
+### General
+- The `compressionQuality` property in the `pickFiles` method now defaults to `0`, and `allowCompression` is set to `false`. [@vicajilau](https://github.com/vicajilau).
```
navaronbracke, line 3: I think you removed the wrong duplicated section here? Now the breaking change label is gone.

vicajilau:
Everything ready as you suggested @navaronbracke 🚀

navaronbracke:
Updated the OP so that the mentioned issues will be closed when this is merged.

navaronbracke (approved):

lib/src/file_picker_io.dart
aprzedecki: its still broken

PR #1752, opened 2025-04-01
Feat: Add ability to auto detect mime type on android

Leapward-Koex:
This is fix for https://github.com/miguelpruivo/flutter_file_picker/issues/1669, taking a different approach to the already open PR here https://github.com/miguelpruivo/flutter_file_picker/pull/1682.

This PR takes into consideration of the [recommendation](https://github.com/miguelpruivo/flutter_file_picker/pull/1682#discussion_r2002650880) to try and automatically infer the mime type from the bytes and file name extension (if supplied) to pass to the `ACTION_CREATE_DOCUMENT` intent instead of making the mime type a field on the API. 

I ended up picking Apache Tika for this as it proved to the most accurate for many types of common files. Using the inbuilt java functionality such as [`URLConnection.guessContentTypeFromStream`](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#guessContentTypeFromStream-java.io.InputStream-) proved to be inaccurate for certain file types such as webp and thus unsuitable for this usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 10.1.0
+### Android
+- Added new parameter `inferMimeType` to `saveFile()` which, when enabled, infers the MIME type using Apache Tika. Fixes incorrect file incrementing if a file with the same name in the destination folder already exists. ([#1669](https://github.com/miguelpruivo/flutter_file_picker/issues/1669)) [@Leapward-Koex](https://github.com/Leapward-Koex)
+
 ## 10.0.0
 ### General
 - **BREAKING CHANGE:** The `compressionQuality` property in the `pickFiles` method now defaults to `0`.
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -37,6 +37,7 @@ android {
         implementation 'androidx.core:core:1.0.2'
         implementation 'androidx.annotation:annotation:1.0.0'
         implementation "androidx.lifecycle:lifecycle-runtime:2.1.0"
+        implementation 'org.apache.tika:tika-core:3.1.0'
     }
 
     if (project.android.hasProperty("namespace")) {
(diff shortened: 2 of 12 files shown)


Shunt22:
Hey @navaronbracke. Please give this PR some love, if you have any spare time. 
#1669 is a big issue for me.

navaronbracke:
Hi @Shunt22 , we recently migrated the codebase from Java to Kotlin.
In that migrating we also added a dependency on Apache Tika, to detect the mime type for the intent.

See  https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt#L203-L207 (detecting the mime type from the bytes)

and https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt#L226-L228 (setting the intent type when calling saveFile)

So I think this PR can be closed as a duplicate.

cc @vicajilau to confirm.

navaronbracke:
@vicajilau We didn't do this, though? Maybe we should still add that?

```
                if (fileName != null && !fileName.isEmpty()) {
                    metadata.set(TikaCoreProperties.RESOURCE_NAME_KEY, fileName);
                }
```

vicajilau:
@navaronbracke Very good point! The truth is that for inferring types in those format types that do not have headers or magic number like CSV which is plain text it could be useful. The idea is that it complements the detector through the metadata. I deal with it in another PR.

navaronbracke:
Alright, then I'll close this PR and we can follow up in that new PR for the `TikaCoreProperties.RESOURCE_NAME_KEY` tweak to the detection.

Issue #1493, opened 2024-04-16
Reimplement linux file picker using native APIs in the app rather than calling out to zenity


Hixie:
The current approach requires that an additional package be installed for the application to work, and causes a new system process to be spawned for the dialog that is separate from the actual app. This is particularly problematic in scenarios where the system is locked down to prevent applications from spawning processes.

There doesn't seem to be any reason why we could implement this using regular message channels to code implemented natively in the app, like a normal plugin.

navaronbracke:
@vicajilau i recall that zenity/quarma were removed a long while ago?

vicajilau:
This is already resolved. The current Linux picker implementation talks to org.freedesktop.portal.Desktop over DBus directly in-process, with no dependency on zenity, kdialog, or any external subprocess. Closing this issue.

Issue #1282, opened 2023-05-17
Not working on Linux-Kylin


cuishijie1991:
hi, When I run my app on Linux-Kylin, the file picker not working with exception :
_"flutter: Exception: Couldn't find the executable zenity in the path"_

I know that the Linux-Kylin is not a version of the operating system supported by the mainstream. But I still hope to get some help message to know how to fix it, thank u!

----------------
Flutter assets will be downloaded from https://storage.flutter-io.cn. Make sure you trust this source!
[✓] Flutter (Channel stable, 3.10.0, on Kylin V10 SP1 5.10.0-8-generic, locale zh_CN.UTF-8)
    • Flutter version 3.10.0 on channel stable at /home/IT/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 84a1e904f4 (8 天前), 2023-05-09 07:41:44 -0700
    • Engine revision d44b5a94c9
    • Dart version 3.0.0
    • DevTools version 2.23.1
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn
[✗] Android toolchain - develop for Android devices
    ✗ Unable to locate Android SDK.
      Install Android Studio from: https://developer.android.com/studio/index.html
      On first launch it will assist you in installing the Android SDK components.
      (or visit https://flutter.dev/docs/get-started/install/linux#android-setup for detailed instructions).
      If the Android SDK has been installed to a custom location, please use
      `flutter config --android-sdk` to update to that location.
[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[✓] Linux toolchain - develop for Linux desktop
    • clang version 10.0.0-4kylin1
    • cmake version 3.16.3
    • ninja version 1.10.0
    • pkg-config version 0.29.1
[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/docs/get-started/install/linux#android-setup for detailed instructions).
[✓] VS Code (version 1.78.1)
    • VS Code at /usr/share/code
    • Flutter extension version 3.64.0
[✓] Connected device (1 available)
    • Linux (desktop) • linux • linux-x64 • Kylin V10 SP1 5.10.0-8-generic
[✓] Network resources
    • All expected network resources are available.

[...]

PR #1759, opened 2025-04-05
feat: introduce new method `pickFileAndDirectoryPaths()`

philenius:
**Changes:**

https://github.com/user-attachments/assets/2e137624-e868-416c-a902-3e38817417e9

* Introduced a new method, `pickFileAndDirectoryPaths()`, which displays a dialog enabling users to select both files and directories simultaneously. The method returns their absolute paths. Only available for macOS. Implements #1105.  
* Updated `README.md`:
   * Updated compatibility chart:
      | Before | After |
      |-|-|
      | image | image |
   * Added screen recording of demo app on macOS
* Bumped package version from 10.0.0 to 10.1.0 
* Updated example Flutter app to contain a button for the newly introduced method `pickFileAndDirectoryPaths()`


lib/src/file_picker.dart
```diff
@@ -108,6 +108,31 @@ abstract class FilePicker extends PlatformInterface {
...
+  ///
+  /// Returns a [Future<List<String>?>] that resolves to a list of absolute
+  /// paths for the selected files and directories. If the user cancels the
+  /// dialog or if the paths cannot be resolved, the method returns `null`.
+  Future<List<String>?> pickFileAndDirectoryPaths({
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+  }) async =>
+      throw UnimplementedError('pickFiles() has not been implemented.');
```
navaronbracke, line 134:
```suggestion
      throw UnimplementedError('pickFileAndDirectoryPaths() has not been implemented.');
```

macos/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -95,7 +98,49 @@ public class FilePickerPlugin: NSObject, FlutterPlugin {
...
+            
+            let pathResult = dialog.urls
+            
+            if pathResult.isEmpty {
+                result(nil)
+            } else {
+                let paths = pathResult.map { $0.path }
+                result(paths)
+            }
+            return
```
navaronbracke, line 140: This return is redundant

```suggestion
```

README.md
```diff
@@ -32,19 +32,21 @@ A package that allows you to use the native file explorer to pick single or mult
...
-| pickFiles()           | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
-| saveFile()            | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
+| API                           | Android            | iOS                | Linux              | macOS              | Windows            | Web                |
+|-------------------------------|--------------------|--------------------|--------------------|--------------------|--------------------|--------------------|
+| `clearTemporaryFiles()`       | :white_check_mark: | :white_check_mark: | :x:                | :x:                | :x:                | :x:                |
+| `getDirectoryPath()`          | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x:                |
+| `pickFileAndDirectoryPaths()` | :x:                | :x:                | :x:                | :white_check_mark: | :x:                | :x:                |
```
navaronbracke, line 47: @philenius Out of curiosity, is there a possibility to add this to other platforms in the future?
Or is this stricly MacOS only?

example/lib/src/file_picker_demo.dart
```diff
@@ -1,7 +1,8 @@
 import 'package:file_picker/file_picker.dart';
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'dart:convert';
+import 'dart:io' as io;
```
navaronbracke, line 5: Don't import `dart:io`. This would break the example when run on the web.

Instead, use https://pub.dev/packages/file in the example app, and guard it with `kIsWeb`.

example/lib/src/file_picker_demo.dart
```diff
@@ -33,11 +55,32 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         .addListener(() => _extension = _fileExtensionController.text);
   }
 
+  Widget _buildFilePickerResultsWidget(
+      {required int itemCount,
+      required Widget Function(BuildContext, int) itemBuilder}) {
+    final ScrollController scrollController = ScrollController();
```
navaronbracke, line 61: This scroll controller is not disposed. I would just omit the scroll controller / scrollbar? It's only a sample anyway.

example/lib/src/file_picker_demo.dart
```diff
@@ -33,11 +55,32 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+    return Container(
+      height: MediaQuery.of(context).size.height * 0.50,
```
navaronbracke, line 63: Use the inherited aspect of MediaQuery

```suggestion
      height: MediaQuery.sizeOf(context).height * 0.50,
```

example/lib/src/file_picker_demo.dart
```diff
@@ -33,11 +55,32 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         .addListener(() => _extension = _fileExtensionController.text);
   }
 
+  Widget _buildFilePickerResultsWidget(
+      {required int itemCount,
+      required Widget Function(BuildContext, int) itemBuilder}) {
```
navaronbracke, line 60: Missing comma, this results in a weird formatting diff.

```suggestion
      required Widget Function(BuildContext, int) itemBuilder,}) {
```

example/lib/src/file_picker_demo.dart
```diff
@@ -33,11 +55,32 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+    final ScrollController scrollController = ScrollController();
+    return Container(
```
navaronbracke, line 62: This container is redundant? See https://dart.dev/tools/linter-rules/avoid_unnecessary_containers

```suggestion
    return SizedBox(
```

example/lib/src/file_picker_demo.dart
```diff
@@ -50,17 +93,78 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
       _isLoading = false;
-      _fileName =
-          _paths != null ? _paths!.map((e) => e.name).toString() : '...';
-      _userAborted = _paths == null;
+      _userAborted = hasUserAborted;
+      _resultsWidget = _buildFilePickerResultsWidget(
+        itemCount: pickedFilesAndDirectories?.length ?? 0,
+        itemBuilder: (BuildContext context, int index) {
+          final String name =
+              new io.File(pickedFilesAndDirectories![index]).statSync().type ==
```
navaronbracke, line 154: Avoid using `new` as it is redundant since Dart 2.0

```suggestion
              new io.File(pickedFilesAndDirectories![index]).statSync().type ==
```

example/lib/src/file_picker_demo.dart
```diff
@@ -122,20 +244,29 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         fileName: _defaultFileNameController.text,
         initialDirectory: _initialDirectoryController.text,
         lockParentWindow: _lockParentWindow,
-        bytes: _paths?.first.bytes,
+        bytes: utf8.encode('Hello, world!'),
```
navaronbracke, line 247: What if the file isn't a text file? I'm aware that the previous line was an error, though.

example/lib/src/file_picker_demo.dart
```diff
@@ -354,6 +480,13 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
                           icon: const Icon(Icons.folder),
                         ),
                       ),
+                      SizedBox(
+                        width: 250,
+                        child: FloatingActionButton.extended(
+                            onPressed: () => _pickFileAndDirectoryPaths(),
+                            label: Text('Pick files and directories'),
+                            icon: const Icon(Icons.folder_open)),
```
navaronbracke, line 488: Missing comma resulting in weird formatting indent.

```suggestion
                            icon: const Icon(Icons.folder_open),),
```

example/lib/src/file_picker_demo.dart
```diff
@@ -423,64 +556,10 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-                                    )
-                                  : _saveAsFileName != null
-                                      ? ListTile(
-                                          title: const Text('Save file'),
-                                          subtitle: Text(_saveAsFileName!),
-                                        )
-                                      : const SizedBox(),
+                          : _resultsWidget,
                 ),
                 SizedBox(
```
navaronbracke, line 561:
```suggestion
                const SizedBox(
```

navaronbracke (changes requested):
New file picker API looks good to me. Mainly remarks on the example app though.

example/lib/src/file_picker_demo.dart
philenius: I understand your point but what do you want me to do? :D I cannot provide an example file or example bytes for all possible file formats (jpg, png, pdf, txt, md, etc. pp.).

philenius:
Thank you for your in-depth code review, @navaronbracke  :)

example/lib/src/file_picker_demo.dart
philenius: I have replaced it with `import 'package:file/local.dart';` and execute the code only when we're not on web. Is that what you meant?

```dart
...

          String name = 'File path:';
          if (!kIsWeb) {
            final fs = LocalFileSystem();
            name = fs.isFileSync(pickedFilesAndDirectories![index])
                ? 'File path:'
                : 'Directory path:';
          }

...
```

README.md
navaronbracke: Well, then it's fine to have the API implemented as-is. _If_ there is a future possibility for other platforms, that can be addressed later.

example/lib/src/file_picker_demo.dart
navaronbracke: Exactly!

example/lib/src/file_picker_demo.dart
navaronbracke: Well, I was only asking this, since in theory the user could set `_extension` to one for a non-plaintext one? (I.e. .pdf, like you mentioned) What if we just have the extension hardcoded to `.txt` ?

philenius: That's correct. Gotcha. I hardcoded it as you said. At least on macOS, it works as expected (haven't tested it on Linux/Windows):
ScreenRecording2025-04-05at20 04 10-ezgif com-video-to-gif-converter

vicajilau: I can confirm this is working properly on Linux 
image

vicajilau: But if I try to pick a pdf file when I try to save fails. Look @philenius, this is what @navaronbracke is trying to tell you :)
[Screencast from 2025-04-06 18-10-05.webm](https://github.com/user-attachments/assets/cabe07d0-6db4-48c5-bc6f-4836bbe8888f)

philenius: I already implemented the proposal https://github.com/miguelpruivo/flutter_file_picker/pull/1759#discussion_r2029934872 in commit af47334deffb82c6ae457579470eb6703748ff26. This improved the behavior of the demo app on macOS.

@vicajilau, I understand your point that the demo app isn't working perfectly fine on Linux (and probably Windows, too). But please tell me what you want me to do, @navaronbracke and @vicajilau.

In the end, this is just a demo app made for developers. Does the demo app really need to work for all kind of scenarios? This pull request is not about making the perfect demo app. Instead this pull request is about adding a new feature for macOS.

Please approve this pull request or provide an idea how you want this issue in the demo app to be fixed. I'd be happy to provide further fixes.

navaronbracke: This is fine as it is now. Like you said, it's an example app.

navaronbracke (approved):

stephane-archer:
You guys are the best ! Thank you 🙏

geotro:
@philenius thank you so much for making it happen!  What would it take to implement this on Windows and Linux as well?

philenius:
thanks, @geotro :)

Providing the same functionality on other platforms requires an operating system API that we could call.

* **iOS & Android:** I'm neither familiar with Android nor with iOS, hence I can't answer on that.

* **Linux:** On Linux, our current implementation is based on executing one of the three command line tools `qarma`, `kdialog` or `zenity`. This approach is quite buggy and if we knew how to implement file picker dialogs in C++, we would have already done it (e.g. as requested in #1493, #1282, #1270). According to the [docs](https://help.gnome.org/users/zenity/stable/file-selection.html.en), `zenity` has no option to pick files and directories simultaneously. The same applies to [`qarma`](https://github.com/luebking/qarma/blob/master/Qarma.cpp#L782). As long as these tools do not provide this functionality, there is no way to implement this feature on Linux. If we reimplement the Linux file picking in C++, there might be a chance to implement `pickFileAndDirectoryPaths()`.

* **Windows:**  Our Windows implementation for picking files is based on `OPENFILENAMEW`. Again, [according to the docs](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew#:~:text=or%20Open).-,Flags,Type%3A%20DWORD,-A%20set%20of) there seems to be no flag to pick both, files and directories, simultaneously. If we port our implementation from `OPENFILENAMEW` to [`FileOpenDialog`](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/ne-shobjidl_core-_fileopendialogoptions), we might have a chance of implementing `pickFileAndDirectoryPaths()`. But again, we'd have to play around with these APIs.

PR #1766, opened 2025-04-12
Improve mimetype detection on Android

vicajilau:


android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -200,10 +203,19 @@ object FileUtils {
         return mimeType.substringAfter("/")
     }
 
-    fun getMimeTypeForBytes(bytes: ByteArray?): String {
+    fun getMimeTypeForBytes(fileName: String?, bytes: ByteArray?): String {
         val tika = Tika()
-        val mimeType = tika.detect(bytes)
-        return mimeType
+        val detector = tika.detector
```
navaronbracke, line 208: Can we invert this condition, to avoid the negation? Also, we can avoid using `tika.detector` if it is not needed.

```
if (fileName.isNullOrEmpty()) {
  return tika.detect(bytes)
}

val detector = tika.detector
val stream = TikaInputStream.get(bytes)
val metadata = Metadata()

metadata.set(TikaCoreProperties.RESOURCE_NAME_KEY, fileName)

return detector.detect(stream, metadata).toString()
```

navaronbracke (changes requested):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Noted. Changed!

vicajilau: Resolved @navaronbracke :)

navaronbracke (approved):

example/lib/src/file_picker_demo.dart
```diff
@@ -234,13 +235,15 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
 
     try {
       pickedSaveFilePath = await FilePicker.platform.saveFile(
-        allowedExtensions: ["txt"],
+        allowedExtensions: (_extension?.isNotEmpty ?? false)
```
navaronbracke, line 238: @vicajilau Why was this reverted? IIRC the original fixed "txt" extension was to fix the bytes on line 246.

vicajilau: I have restored that on this PR

navaronbracke: Perhaps we could redo this so it works for both? (Like if the file is plaintext, it encodes "Hello World", but otherwise it uses the picked file bytes?)

navaronbracke (approved):

Issue #1626, opened 2024-11-06
Crash on IOS while slide down to close picker delegate


virskor:
**Describe the bug**
file_picker: ^8.1.3
while slide down to close picker delegate, app crashed 
```
flutter: [MethodChannelFilePicker] Platform exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
```
but click cancel button to close works well.

**Platform**
- [✅] iOS

**Platform OS version**
What version did it happen?
IOS 18.1
**How are you picking?**
```dart
    final FilePickerResult? result = await FilePicker.platform.pickFiles(
      allowMultiple: false,
      type: FileType.custom,
      lockParentWindow: false,
      allowCompression: false,
      allowedExtensions: ['csv'],
    );

    /// 没有选择文件
    if (result == null || result.files.isEmpty) {
      return;
    }
```

**Details to reproduce the issue**
Provide all the details to reproduce the issue.

**Error Log**
```
flutter: onReceiveClientId: d3fc2febef5f60571f55c1483276349c
flutter: [MethodChannelFilePicker] Platform exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(multiple_request, Cancelled by a second request, null, null)
#0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:648)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:334)
<asynchronous suspension>
#2      MethodChannel.invokeListMethod (package:flutter/src/services/platform_channel.dart:520)
<asynchronous suspension>
#3      FilePickerIO._getPath (package:file_picker/src/file_picker_io.dart:103)
<asynchronous suspension>
```

**Screenshots and/or video**
If applicable, add screenshots or video to help explain your problem.

https://github.com/user-attachments/assets/92e9311d-50b1-498e-9223-06015bfd76da

**Flutter Version details**
Please, post the output of your `flutter doctor -v`, preferably, while running the issued device/simulator.

**Additional context**

virskor:
```
[...]

PR #1788, opened 2025-04-30
Dont clear _result of previous method call when multiple method calls are performed at once on iOS

Leapward-Koex:
When the iOS plugin detects that a method call is already in progress it returns a error result for the subsequent call **but sets the previously saved `_result` to `nil`**.

This causes a native app crash when the first (inprogress call) calls `- (void)documentPickerWasCancelled` as it tries to call `_result(nil)` but _result has been set to `nil` from the subsequent call. Fixes https://github.com/miguelpruivo/flutter_file_picker/issues/1626

This also fixes an issue where in dart if you call FilePicker.platform.saveFile twice without waiting for the first call to finish and you catch the exception from the second call, that the first call's future will never resolve (as the _result was nilled by the second call).

Example of this second issue
```dart
       String? pickedSaveFilePath;
       final future1 = FilePicker.platform.saveFile(
           allowedExtensions: [".dat"],
           type: FileType.custom,
           dialogTitle: "Title",
           fileName: "myfile.dat",
           initialDirectory: "<a valid directory>",
           bytes: new Uint8List(100));
 
       try {
         await Future.delayed(Duration(seconds: 5));
         final future2 = FilePicker.platform.saveFile(
             allowedExtensions: [".dat"],
             type: FileType.custom,
             dialogTitle: "Title",
             fileName: "myfile.dat",
             initialDirectory: "<a valid directory>",
             bytes: new Uint8List(100));
         await future2;
       } catch (e) {
         // Expected due to calling FilePicker.platform.saveFile twice without waiting for the first call to finish.
         debugPrint(e.toString());
       }
       pickedSaveFilePath = await future1;
       // This code is unreachable as future1 will never complete.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 10.1.3
 ### iOS
 - Fixed an app crash when calling `.saveFile` twice and cancelling the native save operation via the UI [#1626](https://github.com/miguelpruivo/flutter_file_picker/issues/1626) [@Leapward-Koex](https://github.com/Leapward-Koex)
-- Fixed a .saveFile future never completing when `.saveFile` is called twice without waiting for the first completion. [@Leapward-Koex](https://github.com/Leapward-Koex)
+- Fixed a `.saveFile` future never completing when `.saveFile` is called twice without waiting for the first completion. [@Leapward-Koex](https://github.com/Leapward-Koex)
 
 ## 10.1.2
 ### Android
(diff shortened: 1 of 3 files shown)


navaronbracke (changes requested):
Can you fix the merge conflict and increment the patch version?

Otherwise this is LGTM.

Leapward-Koex:
Sure, I've fixed the conflict and bumped the version to 10.1.5. @navaronbracke

navaronbracke (approved):

PR #1805, opened 2025-05-12
(feat) add broader uri to path conversion handling

ViscousPot:
This branch has been rebased on the latest master

Main changes included are to add support for older Android versions and accurately converting paths across the board.

#### Tested with:
-  Documents folder on Android 10
-  Downloads folder on Android 10
-  Documents folder on Android KitKat

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -162,7 +162,7 @@ object FileUtils {
                         putExtra(Intent.EXTRA_MIME_TYPES, it.toTypedArray())
                     }
                 }
-                
+
             }
 
 
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -7,7 +7,6 @@ import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.net.Uri
 import android.os.Build
-import android.database.Cursor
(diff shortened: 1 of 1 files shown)


daniJimen (approved):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -506,13 +514,74 @@ object FileUtils {
...
+    }
+
     @JvmStatic
-    fun getFullPathFromTreeUri(treeUri: Uri?, con: Context): String? {
+    fun getFullPathFromTreeUri(uri: Uri, treeUri: Uri?, con: Context): String? {
         if (treeUri == null) {
             return null
         }
 
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
```
daniJimen, line 550: why its needed to control sdk version? in what part of code force you to use only in versions lower than 4.4?

ViscousPot: Ah yeah, the minimum SDK is 21 👌🏾 
I'll remove this from the PR 🙏🏾

vicajilau:
Hi @ViscousPot,

Thanks for the PR. While we appreciate the effort to improve path resolution on older Android versions, the plugin has moved towards a full SAF (Storage Access Framework) implementation in version 12.

We are now discouraging the use of absolute paths on Android, as they are increasingly unreliable and restricted by Scoped Storage. Instead, we now provide `identifier` (URIs) and `safHandle` so developers can interact with the files using the official Android APIs, which is the robust way to handle this moving forward.

Since this PR adds significant complexity for a legacy behavior that we are moving away from, we've decided not to merge it.

PR #1808, opened 2025-05-13
Switch to AndroidLifecycle

jmperezfminsait:


vicajilau:
Hi, thanks for the PR.

I’m not sure I understand the reason behind this change. Why should we replace `FlutterLifecycleAdapter.getActivityLifecycle(activityBinding)` with a direct cast from `activityBinding.lifecycle`?

As far as I know, the `FlutterLifecycleAdapter` is the recommended and safe way to retrieve the lifecycle from the activity binding. This approach ensures compatibility across different versions of the embedding and various `Activity` implementations.

This usage is also documented in the official `flutter_plugin_android_lifecycle` package on pub.dev:
👉 [https://pub.dev/packages/flutter\_plugin\_android\_lifecycle](https://pub.dev/packages/flutter_plugin_android_lifecycle)

Could you please clarify the motivation behind this modification?

Thanks!

Issue #1833, opened 2025-07-07
File picker cancels upload when browser extensions intervene


mugglmenzel:
There is an issue in file picker web that could be resolved by introducing a configuration parameter.

There are various extensions, particularly for Chrome Enterprise, which scan selected files with exfiltration filters. When a scan happens, the window usually loses focus.

The following line then prevents the upload to continue after a scan:
https://github.com/miguelpruivo/flutter_file_picker/blob/9e1ace52cf31c327c0b979c173f9b0fdd5647546/lib/_internal/file_picker_web.dart#L169

When any extension intervenes, the line cancels the upload because focus on the window was lost. The line should either be removed or a configuration parameter should allow to skip the line (e.g. cancelUploadOnWindowBlur or cancelUploadOnWindowFocusLost).

PR #1834, opened 2025-07-07
Introduce bool flag to control upload behavior on window blur

mugglmenzel:
See issue #1833 for more details.

Feature:
This PR adds a new configuration flag for uploads on web which prevents upload cancellation when window focus is lost. 

Reason:
Various extensions (e.g. exfiltration scanners) process upload files and remove focus from a window. Thereby uploads are cancelled due to the windows focus cancellation listener.


vicajilau:
@mugglmenzel Many thanks for this and sorry the delay!! Could you update this fork with master and add an entry in changelog for 10.3.0 version?

PR #1847, opened 2025-07-30
Type 'bool' is not a subtype of type 'Map' when onFileLoading is used

aprzedecki:
During file handling, data param is a Map with type: progress field

### DIFF
--- a/lib/src/file_picker_io.dart
+++ b/lib/src/file_picker_io.dart
@@ -100,7 +100,7 @@ class FilePickerIO extends FilePicker {
                 onFileLoading(data ? FilePickerStatus.picking : FilePickerStatus.done);
               },
               onError: (error) => throw Exception(error),
-        );
+            );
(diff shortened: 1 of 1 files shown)


vicajilau:
@aprzedecki  The current version has already fixed this bug. Could you confirm that it has been fixed so we can close the PR?

vicajilau:
Yes! This has been resolved in the latest version 10.3.0

navaronbracke (approved):
@aprzedecki Can you bump the patch version and add a changelog entry? Then this can go in the next hotfix.

aprzedecki:
> @aprzedecki Can you bump the patch version and add a changelog entry? Then this can go in the next hotfix.

Done. Cheers

vicajilau:
Good point guys!!

navaronbracke:
Filed #1864

PR #1848, opened 2025-07-31
fix issue loading

daniJimen:
fix issue loading


navaronbracke (changes requested):
This is also going to need a version bump for the patch version + a changelog entry

vicajilau (approved):

navaronbracke (approved):

PR #1849, opened 2025-08-03
Fix picking CSV files on Android

SoftWyer:
This fixes an issue on Android where the file picker does not allow CSV files to be picked when using the Custom option.

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -302,6 +302,15 @@ object FileUtils {
             return null
         }
 
+        // On Android, the CSV mime type from getMimeTypeFromExtension() returns
+        // "text/comma-separated-values" which is non-standard and doesn't filter
+        // CSV files in Google Drive.
+        // (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)
+        // (see https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java#439)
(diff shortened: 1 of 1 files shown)


SoftWyer:
@vicajilau thanks for the heads up on your change 👍

It seems like we're addressing slightly different problems as this change is primarily about _picking_ files and _filtering_ on the CSV type using code like this:

```dart
final files = await FilePicker.platform.pickFiles(
  type: FileType.custom,
  allowedExtensions: ['csv'],
);
```

The filtering works fine when picking files on the Android file system (although I might be misremembering), but it doesn't work when picking files on external file systems like Google Drive.

This is because Android defines the CSV mime-type as `text/comma-separated-values` which is non-standard and doesn't match the `text/csv` standard type. so it finds no files.

I don't think your change will address this specific problem as the mime type is correctly specified as `text/csv`.

Ideally I'd prefer it if the API supported `allowedMimeTypes: []` directly and avoid the archaic file extension conversion. If I had the time I'd create a PR for that 😉

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -323,6 +332,10 @@ object FileUtils {
             }
 
             mimes.add(mime)
+            if(allowedExtensions[i].equals(CSV_EXTENSION)) {
```
vicajilau, line 335: This flow is not totally correct because you are adding mime twice, on 334 line and if is also CSV it's adding another mime with `text/csv`.

miguelpruivo: My bad. Probably due to resolving the conflicts on GitHub web on my phone eheh. Good catch. Thanks!

PR #1860, opened 2025-08-10
Android: Fixed build error and some vitamins

vicajilau:
- Fixed a build issue related to the CSV file selection flow accidentally introduced in #1849.
- Migrated from Java to Kotlin some code related with the first topic.
- Removed warnings related with the first topic too.
- Updated the CI workflow to run tests on both stable and beta Flutter channels.
- Added Logcat messages for improved tracking and debugging.

I have tested CSV implementation and the mime tipe for CSV is detected and filtered properly.

### DIFF
--- a/example/lib/src/file_picker_demo.dart
+++ b/example/lib/src/file_picker_demo.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use, deprecated_member_use_from_same_package
+
 import 'package:file/local.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:flutter/foundation.dart';


.github/workflows/main.yml
```diff
@@ -12,12 +12,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        flutter: [ '3.24.0', '3.27.0', '3.29.0', '3.32.0' ]
+        flutter: [ '3.24.0', '3.27.0', '3.29.0', '3.32.0', 'stable', 'beta' ]
```
navaronbracke, line 15: Currently if we run this on "stable", we run it twice? (stable is 3.32.x at the moment)

navaronbracke (changes requested):

navaronbracke (approved):

navaronbracke:
@vicajilau Can you address this analysis warning?

```
Run flutter analyze
Analyzing flutter_file_picker...                                

   info • 'value' is deprecated and shouldn't be used. Use initialValue instead. This will set the initial value for the form field. This feature was deprecated after v3.33.0-1.0.pre • example/lib/src/file_picker_demo.dart:369:25 • deprecated_member_use
```

This does mean that running these on beta also works, though!

PR #1867, opened 2025-08-15
Update Flutter version matrix in CI pipeline

vicajilau:
This pull request makes a minor update to the CI workflow configuration by adding Flutter version `3.32.0` to the test matrix because `stable` now is not `3.32.0` anymore, [3.35.0 just arrived](https://medium.com/flutter/whats-new-in-flutter-3-35-c58ef72e3766). This ensures that automated tests will run against the full list of Flutter releases.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        flutter: [ '3.24.0', '3.27.0', '3.29.0', 'stable', 'beta' ]
+        flutter: [ '3.24.0', '3.27.0', '3.29.0', '3.32.0', 'stable', 'beta' ]
     name: Tests on Flutter ${{ matrix.flutter }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        flutter: [ '3.24.0', '3.27.0', '3.29.0', 'stable', 'beta' ]
+        flutter: [ '3.24.0', '3.27.0', '3.29.0', '3.32.0', 'stable', 'beta' ]
```
navaronbracke, line 15: Do we really need to test against 3.24 then? What is our policy about minimum supported versions?
Because now we run 6x.

navaronbracke (approved):

.github/workflows/main.yml
vicajilau: I love that question! In my point of view between no more than 3 versions would be enough.. what do you think?

PR #1881, opened 2025-09-03
Update compile sdk to 36 to support 16kb page size

shivanshtalwar0:
Without this android app with ndk 28 isn't compiling and 16kb page support is made mandatory by google as it is supported from ndk 28 this minor fix makes app working again


navaronbracke:
This is alrady in progress at https://github.com/miguelpruivo/flutter_file_picker/pull/1882

PR #1882, opened 2025-09-04
Feature/update dependencies

daniJimen:
**API Level 36 Support**

- The SDK now fully supports API level 36, ensuring compatibility with the latest platform features and system behaviors.


navaronbracke (changes requested):
This is going to need a changelog entry & version bump.

PR #1901, opened 2025-10-14
fix(Android): modify filtering logic when type is 'media'

ChoiHyunjin:
### Android
- Fixes an issue where all files were displayed when selecting the ‘media’ type.
- Refactors parts of the code.


vicajilau:
@navaronbracke This code is now obsolete given the current state of the master branch, and image and video selection has been improved in previous versions. Additionally, MediaStore only supports image and video selection, so I think we can close this PR.

Issue #1921, opened 2025-12-09
Class name conflict


sqcsabbey:
Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened inssues. Also, check if it isn't described on the [Wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki), specially on [Troubleshooting](https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting) page.

Also, sometimes a simple `flutter clean` and `flutter build` again with latest file_picker version, may end up by fixing cached issues, so I encourage you to first do so.

**Describe the bug**
A clear and concise description of what the bug is. If the issue happens to be on Android, please make sure that it also happens with a different device/simulator and/or version.

**Platform**

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
What version did it happen?
iOS 26

**How are you picking?**
```
Please, post the snippet here
```

**Details to reproduce the issue**
After running the [UIScene migration](https://docs.flutter.dev/release/breaking-changes/uiscenedelegate), 'flutter run' fails with the following message: 

> objc[1185]: Class FileUtils is implemented in both /System/Library/PrivateFrameworks/OSAnalytics.framework/OSAnalytics (0x1fdd78f70) and /private/var/containers/Bundle/Application/9CA56EA6-B1A0-4576-A2FB-B02FF6D0BCBA/Runner.app/Frameworks/file_picker.framework/file_picker (0x1013715d8). 
> This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed. 

**Error Log**
Please, post the _full_ console log of your issue, if applicable.

```
Launching lib/main.dart on Craig’s iPhone in debug mode...
Automatically signing iOS for device deployment using specified development team in Xcode project: QUJN68ZHKN
Running pod install...
Running Xcode build...
Xcode build done.                                           53.0s
Installing and launching...
objc[1249]: Class FileUtils is implemented in both /System/Library/PrivateFrameworks/OSAnalytics.framework/OSAnalytics (0x1fdd78f70) and /private/var/containers/Bundle/Application/DC54F5FE-0AD5-4567-AA17-E957F32F7FEB/Runner.app/Frameworks/file_picker.framework/file_picker (0x104cc55d8). This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
```

**Screenshots and/or video**
[...]

PR #1927, opened 2025-12-18
refactor(ios): Rename FileUtils to FilePickerUtils

vicajilau:
Renames the `FileUtils` class and source files to `FilePickerUtils` in the iOS implementation. This includes updating the class definition, implementation, and the umbrella header import.

Fixed: https://github.com/miguelpruivo/flutter_file_picker/issues/1921

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 10.3.9
+### iOS
+- Rename FileUtils to FilePickerUtils [#1921](https://github.com/miguelpruivo/flutter_file_picker/issues/1921)
+
 ## 10.3.8
 ### Android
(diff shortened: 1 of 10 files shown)


navaronbracke:
> @navaronbracke From what I can see, [sharedDarwinSource](https://docs.flutter.dev/packages-and-plugins/developing-packages#shared-ios-and-macos-implementations) only serves to unify flows on iOS and macOS. But I wouldn't mind rewriting the Objective-C code in Swift. Shall we create an issue to follow up on this?

Correct. My suggestion is that we rewrite the MacOS portion in Swift, and unify with iOS in the same refactor.
We can follow up with a new feature request issue.

Maybe we can do that in two passes though (one Obj-C -> Swift for MacOS PR, one sharedDarwinSource:true PR)

PR #1928, opened 2025-12-18
Release version to 10.3.8

vicajilau:
Updates `pubspec.yaml` and `CHANGELOG.md` to revert the version from 10.3.9 to 10.3.8.


vicajilau:
This way we can combine the two publications into a single release 🚀

PR #1929, opened 2025-12-19
ci: Dynamically fetch Flutter versions for the CI test matrix instead…

vicajilau:
This pull request updates the GitHub Actions workflow to dynamically fetch and use the latest stable Flutter versions for testing, instead of hardcoding them. This makes it easier to keep tests up-to-date with new Flutter releases.

**Workflow improvements:**

* Added a new `setup_flutter_versions` job that fetches the five latest stable Flutter versions (plus "stable" and "beta" channels) from the Flutter releases API and exposes them as an output for downstream jobs.
* Updated the `test_flutter_versions` job to use the dynamically fetched Flutter versions from the `setup_flutter_versions` job, replacing the previously hardcoded list.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,24 @@ on:
     branches: [ master ]
 
 jobs:
+  setup_flutter_versions:
+    runs-on: ubuntu-latest
+    outputs:
+      flutter-versions: ${{ steps.id_flutter_versions.outputs.versions }}
+    steps:
+      - name: Fetch Flutter versions
+        id: id_flutter_versions
+        run: |
+          VERSIONS=$(curl -s https://storage.googleapis.com/flutter_infra_release/releases/releases_linux.json | jq -c '[.releases | map(select(.channel == "stable")) | .[:5] | map(.version)] | flatten + ["stable", "beta"]')
+          echo "versions=$VERSIONS" >> $GITHUB_OUTPUT
+
   test_flutter_versions:
+    needs: setup_flutter_versions
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -7,12 +7,24 @@ on:
...
 jobs:
+  setup_flutter_versions:
+    runs-on: ubuntu-latest
+    outputs:
+      flutter-versions: ${{ steps.id_flutter_versions.outputs.versions }}
+    steps:
+      - name: Fetch Flutter versions
+        id: id_flutter_versions
+        run: |
+          VERSIONS=$(curl -s https://storage.googleapis.com/flutter_infra_release/releases/releases_linux.json | jq -c '[.releases | map(select(.channel == "stable")) | group_by(.version | split(".")[:2] | join(".")) | map(.[0]) | sort_by(.release_date) | reverse | .[:5] | map(.version)] | flatten + ["stable", "beta"]')
```
navaronbracke: My bash-fu isn't very strong, so just to double check that this does the following?

Grab the last 4 stable releases, taking the most recent patch version (so for example 3.35.99) and add the current "stable" + "beta" to it?

Can we move this to a script that gets its own unit test?

TIL about https://storage.googleapis.com/flutter_infra_release/releases/releases_linux.json

Also, we now test 6! stable releases, which is way mor than Flutter's own policy of "stable" - 2?

navaronbracke (changes requested):

navaronbracke (changes requested):

navaronbracke (approved):

Issue #1731, opened 2025-03-17
Improve Package Score and Platform Compatibility


vicajilau:
**Description:**
The `flutter_file_picker` package currently has some platform-related issues that, while not affecting its overall score, could cause confusion for users. Specifically:

- The package does not correctly declare support for various platforms (`Android`, `iOS`, `Linux`, `macOS`, `Web`).
- The package fails the WASM compatibility check, which may impact future scoring.

**Steps to Reproduce:**
1. Visit the package score page: [file_picker score](https://pub.dev/packages/file_picker/score)
2. Observe the warnings related to platform support and WASM compatibility.

**Suggested Improvements:**
The problem occurs because imports are running on different platforms causing the parser to not detect that the applications are supported. I consider restructuring the project to improve maintainability, following an approach similar to [pdf_combiner](https://github.com/vicajilau/pdf_combiner) where folder and file organization is more structured.

These changes will not only clarify platform compatibility but also prepare the package for potential future changes in scoring criteria.

**Additional Context:**
- Visit the pdf_combiner package score page: [pdf_combiner score](https://pub.dev/packages/pdf_combiner/score)

Would love to hear your thoughts on this @miguelpruivo  and @navaronbracke! 🚀

navaronbracke:
I do agree on all of this. I think this could be a good first step / low hanging fruit to start fixing up the platform interface.

The first thing that I would do is the following:
1. `flutter create .` the project again, to recreate the plugin setup / registration for all 6 supported platforms (Android / iOS / MacOS / Windows / Linux / web)
2. while doing that, move the web implementation into the lib/ folder directly, instead of having it in an `_internal` folder
   - I think that this folder could contribute to the bad score for web? Not sure.
3. the plugin registration should register the existing IO implementation for Android & iOS
   - there are some `Platform.is` checks in there that are very weirdly placed, for example https://github.com/miguelpruivo/flutter_file_picker/blob/abfdf9772b84302fdfd4a1ada911e3c5b371f580/lib/src/file_picker_io.dart#L148 (the super call is just the unimplemented stub, which is very weird)
   - ideally we can just get away with the Platform.is checks for Windows / MacOS / Linux, since those register their own implementations
4. The existing native implementations should be left untouched, this is only a cleanup of the plugin registration
5. Once we have this properly set up, we can migrate web to WASM properly in a follow-up PR

Thoughts @vicajilau ?

miguelpruivo:
[...]

PR #1931, opened 2025-12-20
feat: Add Wasm support and fix pub.dev compatibility score

vicajilau:
This pull request adds support for compiling the package to WebAssembly (Wasm) by introducing conditional exports to avoid using `dart:io` on unsupported platforms, and provides a stub implementation for unsupported platforms. It also updates the macOS example project to require macOS 10.15 and includes a generated plugin package reference.

Platform support and compatibility:

* Added a conditional export in `lib/file_picker.dart` to use `file_picker_io_stub.dart` when `dart:io` is unavailable (such as on Wasm), enabling compilation for WebAssembly and other non-IO platforms.
* Introduced `lib/src/file_picker_io_stub.dart`, which provides a stub `FilePickerIO` implementation that throws `UnimplementedError` for all methods on platforms without `dart:io` support.
* Updated the README to mention support for WebAssembly (Wasm) compilation.

https://github.com/miguelpruivo/flutter_file_picker/issues/1731#issuecomment-3677047875


navaronbracke (changes requested):
I don't think that this is the right way to do things.

The stub is not needed if we follow the best practices for setting up plugins.

For example, another plugin that I do maintenance on does the following:

1) Declare the supported platforms correctly in the pubspec.yaml: https://github.com/juliansteenbakker/mobile_scanner/blob/develop/pubspec.yaml#L36
(note the `fileName` part for web)
2) It delcares a platform interface: https://github.com/juliansteenbakker/mobile_scanner/blob/develop/lib/src/mobile_scanner_platform_interface.dart#L12

`file_picker` _has_ a platform interface, see https://github.com/miguelpruivo/flutter_file_picker/blob/master/lib/src/file_picker.dart#L30

but the plugin itself _is_ the platform interface, which is wrong. (this is where file_picker does not follow the best practices, which leads to issues like the one mentioned)

We should have a `class FilePicker` which includes a `FilePickerPlaftorm extends PlatformInterface` instance.

3) The web impl should provide its own implementation class of the `FilePickerPlaftorm` and can import `package:web` / `dart:js_interop` safely
4) The native impls (Android + iOS / MacOS / Windows / Linux) should implement `FilePickerPlaftorm` separately, with their own Dart classes
5) Clients of the plugin can then call `FilePicker().fooMethod` directly (or if its static, omit the parens), because now they need to access `FilePicker.platform`, which is wrong as well.

I outlined this in https://github.com/miguelpruivo/flutter_file_picker/issues/1731#issuecomment-2730881752

So I will have to decline the approach in this PR. The above will be a breaking change, though, but will benefit us in the long term.

navaronbracke (changes requested):

navaronbracke (changes requested):

navaronbracke (changes requested):

navaronbracke (changes requested):

navaronbracke (approved):
I went ahead and fixed that last comment about the enum docs here on Github.

LGTM, lets merge this! (unless if you have changes left for this PR?)

Issue #1933, opened 2025-12-23
can run web example


archko:
on mine macbook pro m2, run desktop app is ok.
but rund web example app get error

**Platform**

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Desktop

Waiting for connection from debug service on Chrome...
../lib/src/platform/windows/file_picker_windows.dart:3:8: Error: Dart library 'dart:ffi' is not available on this platform.
import 'dart:ffi';
       ^
Context: The unavailable library 'dart:ffi' is imported through these packages:

    web_entrypoint.dart => package:file_picker_example => package:file_picker => dart:ffi
    web_entrypoint.dart => package:file_picker_example => package:file_picker => package:ffi => dart:ffi
    web_entrypoint.dart => package:file_picker_example => package:file_picker => package:win32 => dart:ffi
    web_entrypoint.dart => package:file_picker_example => package:file_picker => package:win32 => package:ffi => dart:ffi
    ...

Detailed import paths for (some of) the these imports:

    web_entrypoint.dart => package:file_picker_example/main.dart => package:file_picker_example/src/file_picker_demo.dart => package:file_picker/file_picker.dart => package:file_picker/src/platform/windows/file_picker_windows.dart => dart:ffi
    web_entrypoint.dart => package:file_picker_example/main.dart => package:file_picker_example/src/file_picker_demo.dart => package:file_picker/file_picker.dart => package:file_picker/src/platform/windows/file_picker_windows.dart => package:ffi/ffi.dart => package:ffi/src/allocation.dart => dart:ffi
    web_entrypoint.dart => package:file_picker_example/main.dart => package:file_picker_example/src/file_picker_demo.dart => package:file_picker/file_picker.dart => package:file_picker/src/platform/windows/file_picker_windows.dart => package:ffi/ffi.dart => package:ffi/src/arena.dart => dart:ffi
    web_entrypoint.dart => package:file_picker_example/main.dart => package:file_picker_example/src/file_picker_demo.dart => package:file_picker/file_picker.dart => package:file_picker/src/platform/windows/file_picker_windows.dart => package:ffi/ffi.dart => package:ffi/src/utf16.dart => dart:ffi
[...]

PR #1935, opened 2025-12-27
ci: add dedicated web build job

vicajilau:
- Adds a dedicated job to build the example app for Web platform.
- Fixed: https://github.com/miguelpruivo/flutter_file_picker/issues/1933

### DIFF
--- a/lib/file_picker.dart
+++ b/lib/file_picker.dart
@@ -2,9 +2,12 @@ export 'src/file_picker.dart';
 export 'src/api/platform_file.dart';
 export 'src/api/file_picker_result.dart';
 export 'src/api/file_picker_types.dart';
+
+// Platform-specific implementations are exported only for plugin registration.
+// These exports are hidden on Web to avoid dart:ffi and dart:io compatibility issues.
 export 'src/platform/linux/file_picker_linux.dart'
-    if (dart.library.html) 'src/web_hysteresis.dart';
+    if (dart.library.html) 'src/platform/web/file_picker_web.dart';
 export 'src/platform/macos/file_picker_macos.dart'
-    if (dart.library.html) 'src/web_hysteresis.dart';
+    if (dart.library.html) 'src/platform/web/file_picker_web.dart';
 export 'src/platform/windows/file_picker_windows.dart'
-    if (dart.library.html) 'src/web_hysteresis.dart';
+    if (dart.library.html) 'src/platform/web/file_picker_web.dart';
--- a/lib/src/web_hysteresis.dart
+++ b/lib/src/web_hysteresis.dart
@@ -1,1 +1,0 @@
-// This file is intentionally empty to hide desktop-specific exports on Web.


lib/src/web_hysteresis.dart
```diff
@@ -0,0 +1 @@
+// This file is intentionally empty to hide desktop-specific exports on Web.
```
navaronbracke: This file is not needed?

lib/file_picker.dart
```diff
@@ -2,7 +2,9 @@ export 'src/file_picker.dart';
 export 'src/api/platform_file.dart';
 export 'src/api/file_picker_result.dart';
 export 'src/api/file_picker_types.dart';
-export 'src/platform/file_picker_method_channel.dart';
-export 'src/platform/macos/file_picker_macos.dart';
-export 'src/platform/linux/file_picker_linux.dart';
-export 'src/platform/windows/file_picker_windows.dart';
+export 'src/platform/linux/file_picker_linux.dart'
```
navaronbracke: Why do we export the desktop implementations?

I don't think that there is _specific_ functionality, tied to either the MacOS/Linux/Windows implementations of `FilePickerPlatform`, so exporting these has no use for clients. (neither implementation adds its own methods on top of `FilePickerPlatform`, they all just implement `FilePickerPlatform`)

IF we want to add specific functionality for only MacOS/Linux/Windows/Web/Android/iOS, we should:

1) Make each platform use a federated package, which implements file_picker
2) Or provide specific methods to FilePickerPlatform, which are only implemented by the specific platform implementations, _without leaking the unavailable imports_ back to clients.

The exports above do leak the unavailable imports into other platforms, which is the bug in question.

For now, I would just not export these implementations to clients and keep them as internal

navaronbracke (changes requested):

lib/src/web_hysteresis.dart
vicajilau: Yes, the file `web_hysteresis.dart` is necessary. It is used as an “empty” file in Dart’s conditional exports to prevent the desktop‑specific implementations (Linux, macOS, Windows) from being imported when compiling for the Web.

vicajilau: True, this can ber removed

lib/file_picker.dart
vicajilau: @navaronbracke You're right ideally, these platform implementations shouldn't be exported as they don't add platform-specific public APIs. However, they must be exported because of how Dart's plugin registration system works.

The `pubspec.yaml` declares `dartPluginClass` for each desktop platform:

```yaml
macos:
  dartPluginClass: FilePickerMacOS
windows:
  dartPluginClass: FilePickerWindows  
linux:
  dartPluginClass: FilePickerLinux
```

This causes Flutter to auto-generate `dart_plugin_registrant.dart` which imports and calls `FilePickerMacOS.registerWith()` etc. from the main `file_picker` package. If we don't export these classes, the generated code fails to compile.

The conditional exports solution avoids the Web compilation issue by replacing desktop implementations with the Web implementation when `dart.library.html` is available:

```dart
export 'src/platform/linux/file_picker_linux.dart'
    if (dart.library.html) 'src/platform/web/file_picker_web.dart';
```

I've verified that this works correctly:

- macOS builds successfully
- Web builds successfully

For a cleaner architecture, as you suggested, we'd need to refactor into [federated plugins](https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins). That would be a larger change beyond the scope of fixing the Web build issue. What do you think?

.github/workflows/main.yml
```diff
@@ -99,6 +99,7 @@ jobs:
           flutter pub get
           flutter build linux
 
+
```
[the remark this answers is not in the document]
vicajilau: Fixed!

navaronbracke (approved):

Issue #1911, opened 2025-11-17
FilePicker.platform.pickFileAndDirectoryPaths doesn't have a dialogTitle option like FilePicker.platform.pickFiles


stephane-archer:
FilePicker.platform.pickFileAndDirectoryPaths doesn't have a dialogTitle option like FilePicker.platform.pickFiles

PR #1937, opened 2026-01-05
feat: Shared Darwin Source Implementation

vicajilau:
Draft PR for shared Darwin source implementation. @navaronbracke please review.


lib/src/platform/file_picker_method_channel.dart
```diff
@@ -67,7 +70,11 @@ class MethodChannelFilePicker extends FilePickerPlatform {
     String? initialDirectory,
   }) async {
     try {
-      return await methodChannel.invokeMethod('dir', {});
+      return await methodChannel.invokeMethod('dir', {
```
navaronbracke: This is a little out of scope for this PR, but these were missing options on MacOS?
Does this fix https://github.com/miguelpruivo/flutter_file_picker/issues/1911 ? If yes, please add this to the PR description, using Github "fixes" syntax, then that issue can be closed properly.

macos/file_picker/Package.swift
navaronbracke: This file is gone, but there is no Package.swift in the Darwin source.

darwin/file_picker.podspec
```diff
@@ -0,0 +1,28 @@
...
+Pod::Spec.new do |s|
+  s.name                  = 'file_picker'
+  s.version               = '0.0.1'
+  s.summary               = 'A flutter plugin to show native file picker dialogs'
+  s.description           = <<-DESC
+A flutter plugin to show native file picker dialogs.
+                       DESC
+  s.homepage              = 'https://github.com/miguelpruivo/plugins_flutter_file_picker'
+  s.license               = { :file => '../LICENSE' }
+  s.author                = 'Miguel Ruivo'
```
navaronbracke: The authors line is not the same between old/new.

darwin/file_picker.podspec
```diff
@@ -0,0 +1,28 @@
...
+  s.source                = { :path => '.' }
+  s.source_files = 'Classes/**/*'
+  s.ios.dependency 'Flutter'
+  s.osx.dependency 'FlutterMacOS'
+  s.ios.deployment_target = '14.0'
+  s.osx.deployment_target = '10.14'
+
+  # Flutter.framework does not contain a i386 slice.
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
```
navaronbracke: Since the preprocessor definitions are now gone (THANK YOU! yes I hated them haha), we should remove any mention of those from the Wiki / README

navaronbracke (changes requested):
This is an initial review of everything, but the Swift sources, since those are still WIP it seems.

example/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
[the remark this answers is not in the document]
vicajilau: Since there are no third-party dependencies, the file is not generated.

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -16,6 +16,7 @@
 		97C146FC1CF9000F007C117D /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FA1CF9000F007C117D /* Main.storyboard */; };
 		97C146FE1CF9000F007C117D /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FD1CF9000F007C117D /* Assets.xcassets */; };
 		97C147011CF9000F007C117D /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FF1CF9000F007C117D /* LaunchScreen.storyboard */; };
+		D35FCDBC272B0CC60B85A8F6 /* Pods_Runner.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5128A6E93B86112246D017A1 /* Pods_Runner.framework */; };
```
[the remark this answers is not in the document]
vicajilau: We should currently support both, Flutter isn't forcing us to make that switch yet. Possibly in the future, as soon as Flutter deprecates the use of CocoaPods, we'll have to. But the important thing here is that I forgot to add support for SPM. I'm working on it.

darwin/file_picker.podspec
navaronbracke: Dominik has fixed bugs in the plugin before, so I do prefer keeping both names. (technically we could also list ourselves here now)

PR #1939, opened 2026-01-08
Implement shared Darwin source code between iOS and macOS

vicajilau:
This PR implements `sharedDarwinSource` between iOS and macOS platforms, using the macOS implementation as the base for unified code.

## Changes

- Enabled `sharedDarwinSource: true` for both iOS and macOS platforms in `pubspec.yaml`
- Updated iOS minimum version to iOS 14, simplifying the codebase by leveraging modern APIs (iOS 13 doesn't support [`PHPickerViewController`](https://developer.apple.com/documentation/photosui/phpickerviewcontroller) and [`PHPickerResult`](https://developer.apple.com/documentation/photosui/phpickerresult-swift.struct))


darwin/file_picker/Package.swift
```diff
@@ -0,0 +1,37 @@
...
+    ],
+    products: [
+        // TODO: Update your library and target names.
+        // If the plugin name contains "_", replace with "-" for the library name.
+        .library(name: "file-picker", targets: ["file_picker"])
+    ],
+    dependencies: [],
+    targets: [
+        .target(
+            // TODO: Update your target name.
```
navaronbracke: This TODO can be removed

darwin/file_picker/Package.swift
```diff
@@ -0,0 +1,37 @@
...
+            name: "file_picker",
+            dependencies: [],
+            resources: [
+                // TODO: If your plugin requires a privacy manifest
```
navaronbracke: This TODO needs to be addressed. We do have a PrivacyManifest.xcprivacy

darwin/file_picker/Package.swift
```diff
@@ -0,0 +1,37 @@
+// swift-tools-version: 5.9
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "file_picker",
+    platforms: [
+        .iOS("14.0"),
```
navaronbracke: Can you put the change to iOS 14.0 in the changelog?

darwin/file_picker/Sources/file_picker/PrivacyInfo.xcprivacy
navaronbracke: I know that the old PrivacyInfo.xcprivacy was directly in the Sources dir, but that is technically against the recommendation to put it inside `Sources/file_picker/Resources/PrivacyInfo.xcprivacy`

Can you move the file to that subfolder?

For example, `file_selector` does it also: https://github.com/flutter/packages/blob/main/packages/file_selector/file_selector_ios/ios/file_selector_ios/Sources/file_selector_ios/Resources/PrivacyInfo.xcprivacy

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -0,0 +1,595 @@
...
+
+    init(registrar: FlutterPluginRegistrar, channel: FlutterMethodChannel) {
+        self.registrar = registrar
+        self.channel = channel
+        super.init()
+    }
+    
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        #if os(iOS)
+        self.result = result
```
navaronbracke: This feels like the wrong palce to do this? Not all methods resolve the result at a later time, in an escaping way

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -0,0 +1,595 @@
...
+        case "clear":
+             #if os(iOS)
+             let cleaned = FilePickerUtils.clearTemporaryFiles()
+             result(cleaned)
+             self.result = nil
+             #else
+             result(true)
+             #endif
+        case "any", "custom":
+             // Unified entry point for file picking with type filters if distinct from pickFiles in some logic,
```
navaronbracke: Can we remove this comment? It feels rather noisy in this place.

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -0,0 +1,595 @@
...
+             handleFileSelection(call, result: result)
+             #endif
+        default:
+            result(FlutterMethodNotImplemented)
+        }
+    }
+    
+    // MARK: - Common Handlers
+    
+    private func handleFileSelection(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
```
navaronbracke: @vicajilau I don't like how this file now becomes one big file with everything in it.
It makes reviewing it quite the task.

Can you create the `/darwin/file_picker/Sources/file_picker` folder in a separate PR,
which then also contains `FilePickerUtils.swift` and `FilePickerMacOS.swift` / `FilePickerIOS.swift` ?

Then I'll review those first. Afterwards, this PR will be much easier to review. (since then this PR is actually setting up the usage of the FilePickerPlugin.swift + usage of the delegates)

navaronbracke (changes requested):

darwin/file_picker/Package.swift
vicajilau: Fixed :)

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -0,0 +1,595 @@
...
+             // but usually mapped to pickFiles in Dart. 
+             // Wait, the Dart code calls 'pickFiles' which sends 'any', 'custom', 'audio', etc. as method names ONLY in legacy code or strictly typed calls?
+             // Checking Dart code: MethodChannelFilePicker invokes `methodChannel.invokeListMethod(type, ...)`
+             // So 'type' IS the method name.
+            handleFileSelection(call, result: result)
+        case "video", "image", "media":
+            #if os(iOS)
+            handleMediaSelection(call, result: result)
+            #else
+            // On macOS these are just file filters in `pickFiles` usually, but if sent as method name:
```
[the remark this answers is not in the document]
vicajilau: Done!

PR #1962, opened 2026-03-08
fix: expose cancelUploadOnWindowBlur in public pickFiles API

### LINKED ISSUE #1961: cancelUploadOnWindowBlur parameter in pickFiles is not part of the public API
The `cancelUploadOnWindowBlur` parameter was added to `FilePickerWeb.pickFiles()` in commit [b40af72](https://github.com/miguelpruivo/flutter_file_picker/pull/1834/changes/b40af72385b4819e53ed91fb04988f36fd8fe6d6) to allow callers to opt out of the window `focus`-based cancellation heuristic on the web. However, it was only added to the `FilePickerWeb` override in file_picker_web.dart and **not** to the base class `FilePicker.pickFiles()` in file_picker.dart.

This means the parameter is not part of the package's public API. While it's technically possible to reach it by importing the internal file and casting:

```dart

import 'package:file_picker/_internal/file_picker_web.dart';



final result = await (FilePicker.platform as FilePickerWeb).pickFiles(

  cancelUploadOnWindowBlur: false,

);

```

this relies on an implementation detail (`_internal/` path) and could break without notice in a future release.

The standard call site `FilePicker.platform.pickFiles(cancelUploadOnWindowBlur: false)` does **not** compile because the base class signature doesn't declare the parameter.

**Proposed fix:**

Add `cancelUploadOnWindowBlur` (with default `true` to preserve existing behavior) to the base class signature in file_picker.dart:

```dart

Future<FilePickerResult?> pickFiles({

    String? dialogTitle,

    String? initialDirectory,

    FileType type = FileType.any,

    List<String>? allowedExtensions,

    Function(FilePickerStatus)? onFileLoading,

    @Deprecated(

        'allowCompression is deprecated and has no effect. Use compressionQuality instead.')

    bool allowCompression = false,

    int compressionQuality = 0,

    bool allowMultiple = false,

    bool withData = false,

    bool withReadStream = false,

    bool readSequential = false,

    bool cancelUploadOnWindowBlur = true,  // <-- add here

  }) async =>

      throw UnimplementedError('pickFiles() has not been implemented.');

```

Documentation should note this parameter is web-only and is ignored on other platforms.

vicajilau:
Exposes the cancelUploadOnWindowBlur parameter in the public FilePicker.pickFiles API. This parameter was previously internal to the web implementation, making it hard to use without internal casts. Adding it to the public API ensures consistency and allows users to opt-out of the window focus-based cancellation on web. Fixes #1961

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -134,7 +134,7 @@ object FileUtils {
     /**
      * Creates and launches an intent for the given file type.
      *
-     * This method is responsible for creating the appropriate intent based on the [type] of file
+     * This method is responsible for creating the appropriate intent based on the type of file
      * that is requested to be picked.
      *
      * This may be either a directory, a regular file, or a gallery pick.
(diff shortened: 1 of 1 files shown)


navaronbracke (changes requested):
This needs a patch version bump / changelog entry.

navaronbracke:
@vicajilau I suppose this happened because it is a named parameter? Are there other named parameters that are missing?

navaronbracke (changes requested):

navaronbracke (approved):

Issue #1845, opened 2025-07-28
[MacOS] File Picker broken when sandboxing is off


azukaar:
**Describe the bug**
When using the picker with MacOS in non sandboxed mode, you manually check for the entitlement for read/write (whichever is needed) to be explicitely set, but **you cannot** set it, when sandbox is off, it is automatically removed even if set in the config file, making the plugin completely unusable under those conditions.

**Platform**

- [x] MacOS

**Platform OS version**
18.4

**How are you picking?**
```

  static Future<String> uploadConfigFile() async {
    FilePickerResult? result = await FilePicker.platform.pickFiles(
      type: FileType.custom,
      allowedExtensions: ['yaml', 'yml', 'json'],
      allowMultiple: false,
    );

    if (result != null && result.files.isNotEmpty) {
      final file = result.files.first;

      String? fileContent;

      if (file.bytes != null) {
        // For web platform, use bytes
        fileContent = String.fromCharCodes(file.bytes!);
      } else if (file.path != null) {
        // For mobile/desktop platforms, use file path
        final selectedFile = File(file.path!);
        fileContent = await selectedFile.readAsString();
      }

      if (fileContent != null && fileContent.isNotEmpty) {
        return fileContent;
      } else {
        throw Exception('File is empty or could not be read');
      }
    } else {
      throw Exception('File upload cancelled');
    }
  }
```

**Details to reproduce the issue**
Change the entitlement file from <true /> to <false /> on the App's entitlement xml

egonbeermat:
Just discovered the same issue here on MacOS with Hardened Runtime capability enabled. FilePicker not usable in this situatio, not sure how to setup Xcode and project in this scenario.

egonbeermat:
I'll try later, thanks! In the interim, I found file_selector package and that had no issues, and provides all the functionality I need, so plugged that in, instead.

IanCarrasco:
Same issue as well!

PR #1964, opened 2026-03-11
Add ability to disable macOS entitlements checks when not needed

PWrzesinski:
This PR resolves: https://github.com/miguelpruivo/flutter_file_picker/issues/1845

Not all macOS applications run with the Sandbox enabled and the current check for entitlements requires it. This PR adds a new method `skipEntitlementsChecks()` that can be called before any other call to FilePicker to disable the entitlements check. I don't think I have permission to update the wiki for this project, so the relevant info should be updated when this is merged.

As an alternative, I considered detecting when Sandbox was not enabled and skipping the check in such cases, but decided to go with the manual way to delegate this decision to the package's users as this affects the security of the end users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Updated Android example app to use Flutter-managed SDK versions (Gradle 8+, Kotlin, targetSdk 36).
 
 ### Desktop (macOS)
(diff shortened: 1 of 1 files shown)


navaronbracke (approved):
LGTM with nits

PR #1981, opened 2026-04-11
chore(windows): migrate win32 to 6.0.1

### LINKED ISSUE #1980: update win32 dependency to 6.0.0
There is a major update of [win32](https://pub.dev/packages/win32/changelog) plugin with significant API updates and it is being picked up by other plugins.

It would be great if the `file_picker` would use the latest `win32` dependency.

Alternatively it might be a good idea to use the [filepicker_windows](https://pub.dev/packages/filepicker_windows) plugin which is maintained by the same folks as `win32` plugin and it provides higher level Dart API for native Windows file picker.

vicajilau:
## Summary
- bump win32 dependency from ^5.9.0 to ^6.0.0
- migrate Windows folder dialog implementation to win32 v6 COM API changes
- replace removed helpers (TEXT/convertToIID) and update enum usage (FOS_*, SIGDN_*)
- adopt explicit COM lifetime management and proper CoTaskMemFree cleanup

## Validation
- dart analyze
- flutter test test/file_picker_windows_test.dart

Closes #1980


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 11.1.0
+### Windows
+- Migrated `win32` dependency to `6.0.0` and updated the Windows folder picker implementation to match the new Win32/COM APIs. [#1980](https://github.com/miguelpruivo/flutter_file_picker/issues/1980)
```
navaronbracke, line 6: Since the new win32 dependency requires dart 3.11, this is a breaking change, so this should not be a minor version bump.

lib/src/platform/windows/file_picker_windows.dart
```diff
@@ -112,83 +112,60 @@ class FilePickerWindows extends FilePickerPlatform {
...
-        if (!SUCCEEDED(hr)) throw WindowsException(hr);
-      } finally {
-        free(title);
-      }
-
-      if (initialDirectory != null) {
-        final folder = TEXT(initialDirectory);
-        final riid = convertToIID(IID_IShellItem);
-        final ppv = calloc<Pointer>();
+      return using((arena) {
```
navaronbracke, line 129: This is an equivalent of try-with-resources, and the arena frees pointers for us?

lib/src/platform/windows/file_picker_windows.dart
```diff
@@ -112,83 +112,60 @@ class FilePickerWindows extends FilePickerPlatform {
...
-        hr = fileDialog.getResult(ppv);
-        if (!SUCCEEDED(hr)) throw WindowsException(hr);
-
-        final item = IShellItem(ppv.cast());
-        final pathPtr = calloc<Pointer<Utf16>>();
-        hr = item.getDisplayName(SIGDN.SIGDN_FILESYSPATH, pathPtr);
-        if (SUCCEEDED(hr)) {
-          selectedPath = pathPtr.value.toDartString();
+        final item = arena.adopt(selectedItem);
+        final pathPtr = item.getDisplayName(SIGDN_FILESYSPATH);
```
navaronbracke, line 168: This is a separate pointer by design? (I'm not too familiar with the new API yet)

navaronbracke (changes requested):

lib/src/platform/windows/file_picker_windows.dart
vicajilau: Yes, the purpose of this change is to use a pattern equivalent to `try-with-resources`. The `using(arena)` block manages the lifecycle of FFI/COM resources registered with the arena and ensures they are released when the scope ends, even if an exception occurs. Technically, this reduces the risk of memory leaks and simplifies manual cleanup.

**Important note**: Only resources managed by the arena are automatically released (for example, those obtained with arena.com or arena.adopt). Pointers that require a specific Windows API freer still need to be explicitly released, that is why `CoTaskMemFree(pathPtr)` remains in a finally block.

lib/src/platform/windows/file_picker_windows.dart
vicajilau: Yes, exactly.

It is a separate pointer by design:
- item is a COM object interface pointer (managed via arena.adopt, so it gets Release called).
- pathPtr from getDisplayName(SIGDN_FILESYSPATH) is a newly allocated PWSTR buffer returned by the shell API.
- That string buffer is not owned by the COM wrapper, so it must be freed explicitly with CoTaskMemFree(pathPtr).

So the split cleanup is intentional: COM lifetime via arena, returned string buffer via CoTaskMemFree.

navaronbracke:
The test failures on the matrix are intentional, since that is tied to the minimum framework version. We will have to update the matrix when this lands, too.

vicajilau:
I've updated the pipeline to check the last three versions. When Flutter 3.44 is released next month, the supported versions will be 3.44, 3.41, and 3.38. Is that correct? @navaronbracke

CHANGELOG.md
vicajilau: Fixed!

vicajilau:
I wanted to get the branch ready for when Flutter 3.41 is released. Now, only the first of the last three versions should fail because it’s not supported. When version 3.41 is released, this should work by restarting the pipeline.

lib/src/platform/web/file_picker_web.dart
```diff
@@ -226,7 +231,8 @@ class FilePickerWeb extends FilePickerPlatform {
     // Start a download by using a click event on an anchor element that contains the Blob.
     HTMLAnchorElement()
       ..href = url
-      ..target = 'blank' // Always open the file in a new tab.
+      ..target =
+          'blank' // Always open the file in a new tab.
```
navaronbracke: Could you move the comment above the assignment? Then I think the formatter puts the assignment on a single line

navaronbracke (approved):
LGTM with nit

example/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: An example of how to use the file_picker plugin.
 publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 
 environment:
-  sdk: ">=2.19.0 <4.0.0"
+  sdk: ">=3.10.0 <4.0.0"
```
navaronbracke, line 6: @vicajilau I just realized, can we put an explicit page with setting here and in the main pubspec now?

Then we can enforce the formatter line length: https://dart.dev/tools/dart-format#page-width

I have a global line length setting in a vscode/settings.json, but that would otherwise override this when I format using the IDE

vicajilau: Good point!! Done :)

MSOB7YY:
hello, any eta on when would this get merged? (no rush just askin ^^)

vicajilau:
> hello, any eta on when would this get merged? (no rush just askin ^^)

Technically, in the next version of Flutter, scheduled for next month 🚀

kjxbyz:
Can you release a beta version first?

MohamedGawdat:
we need this pr ASAP please

Skyost:
> we need this pr ASAP please

@MohamedGawdat Just replace `flutter_file_picker: ^version` by :

```yaml
file_picker:
  git:
    url: https://github.com/miguelpruivo/flutter_file_picker
    ref: chore/win32-6-migration-1980
```

Just keep in mind that it's a free and open-source project. Mainteners do their best to keep it up to date and free of bugs, but they're not paid for this work :wink:

PR #1982, opened 2026-04-12
Running `dart format .` in the repository

vicajilau:
I've formatted the files using the `dart format .` command. This PR contains only the output generated by running that command.

### DIFF
--- a/lib/src/platform/web/file_picker_web.dart
+++ b/lib/src/platform/web/file_picker_web.dart
@@ -55,7 +55,8 @@ class FilePickerWeb extends FilePickerPlatform {
   }) async {
     if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
       throw Exception(
-          'You are setting a type [$type]. Custom extension filters are only allowed with FileType.custom, please change it or remove filters.');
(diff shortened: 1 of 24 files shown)


lib/src/platform/web/file_picker_web.dart
```diff
@@ -226,7 +231,8 @@ class FilePickerWeb extends FilePickerPlatform {
     // Start a download by using a click event on an anchor element that contains the Blob.
     HTMLAnchorElement()
       ..href = url
-      ..target = 'blank' // Always open the file in a new tab.
+      ..target =
```
navaronbracke: Can you put the comment above the assignment and reformat this?

navaronbracke (approved):
LGTM, with nit

lib/src/platform/web/file_picker_web.dart
vicajilau: Done!!

navaronbracke:
Why does it fail on stable, but not on 3.38? Is this because of the formatting fixes that landed after Dart 3.8? (I know there were some fixes for some minor issues in the formatter)

PR #1984, opened 2026-04-13
fix(darwin): avoid false FlutterMacOS SourceKit errors

vicajilau:
## Summary
- guard macOS-specific code paths with `canImport(FlutterMacOS)` in Darwin sources
- wrap `FilePickerPlugin` compilation so SourceKit/VS Code does not try to resolve Flutter macOS types when unavailable
- keep runtime behavior unchanged for actual iOS/macOS builds

## Why
VS Code SourceKit reported `No such module 'FlutterMacOS'` even though the plugin built and worked correctly. This PR removes those false-positive diagnostics while preserving normal plugin functionality.

## Files changed
- darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
- darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift

## Validation
- project builds continue to work as before
- SourceKit false-positive errors for `FlutterMacOS` are no longer reported in VS Code for this code path

### DIFF
--- a/darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
+++ b/darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
@@ -1,10 +1,11 @@
 #if os(iOS)
 import Flutter
-#elseif os(macOS)
+#elseif os(macOS) && canImport(FlutterMacOS)
 import FlutterMacOS
 #endif
 import Foundation
 
+#if os(iOS) || (os(macOS) && canImport(FlutterMacOS))
 public class FilePickerPlugin: NSObject, FlutterPlugin {
     public static func register(with registrar: FlutterPluginRegistrar) {
 #if os(iOS)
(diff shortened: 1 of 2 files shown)


vicajilau:
@navaronbracke This is because I was seeing this error:

image

Neither the current state of the main branch nor my changes affect functionality.

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
```diff
@@ -1,10 +1,11 @@
 #if os(iOS)
 import Flutter
-#elseif os(macOS)
+#elseif os(macOS) && canImport(FlutterMacOS)
```
navaronbracke, line 3: TIL: https://stackoverflow.com/a/73925056

Do we need this for iOS too? (I think not, because of how the Flutter module is structured)
And this is only because VSCode does not handle it correctly? (I always write Swift using Xcode :P)

navaronbracke (approved):

darwin/file_picker/Sources/file_picker/FilePickerPlugin.swift
vicajilau: > Do we need this for iOS too? 

I don't think so either. I think like you!

> And this is only because VSCode does not handle it correctly? 

Yes, that is right. 

> (I always write Swift using Xcode :P)

Me too, that's why I didn't notice it yesterday in the PR. Today, when I opened the project in VSCode, I noticed it!!

PR #1985, opened 2026-04-13
fix(android): normalize duplicate saveFile names to preserve extension

### LINKED ISSUE #1947: Better handling of existing files
```dart
final resultPath = await FilePicker.platform.saveFile(
        fileName: "Ascent.bak",
        bytes: Uint8List.fromList(dbFileBytes),
        type: FileType.custom,
        allowedExtensions: ['bak'],
      );
```

The above line for android saves the file as `Ascent.bak (1)` if the file aready exists in the directory. It should use the name `Ascent (1).bak` if file already exists.

vicajilau:
## Summary
- improve Android saveFile duplicate name handling to keep the extension suffix in the expected position
- add fallback name normalization when providers return duplicate names with extension in the wrong place
- add a changelog entry under 12.0.0 for this Android fix

## Related
- Fixes #1947

## Notes
- This addresses cases where duplicate names can appear as  and normalizes to  when possible via SAF rename APIs.


test/test_files/example.sqlite
navaronbracke, line 1: These two test files should be removed?

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -330,6 +334,51 @@ object FileUtils {
...
+        }
+
+        val extension = originalFileName.substringAfterLast('.', "")
+        if (extension.isBlank()) {
+            return uri
+        }
+
+        val currentName = getFileName(uri, context) ?: return uri
+        val escapedExtension = Pattern.quote(extension)
+        val duplicatedExtensionPattern = Regex("^(.*)\\.${escapedExtension} \\((\\d+)\\)$")
```
navaronbracke, line 355: Can you explain these patterns in a comment in the code?

navaronbracke (approved):

test/test_files/example.sqlite
vicajilau: True, removed!

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Yes, done!

Gabbar-v7:
@vicajilau  Thanks for the fix man

Gabbar-v7:
I might be wrong here, I find this fix to have a minor flaw. Example is shown in image below. Basically suppose a file with name file.sqlite is saved, again saving a file with same name has desired behavoiur i.e. file (1).sqlite. However, saving the file again saves it as `file.bak (1)`, saving the file again saves it as file (2).sqlite
 
image

PR #1986, opened 2026-04-16
fix(android): normalize duplicate save names and avoid nested collision suffixes

### LINKED ISSUE #1947: Better handling of existing files

vicajilau:
## What was happening

On Android document providers, saving a file with an existing name can trigger provider-side collision suffixes (for example: `example.sqlite (1)`).

Our previous rename normalization could still end up accepting provider-generated nested names like:
- `example (1) (1).sqlite`
- `example (1) (2).sqlite`

instead of clean sequential names:
- `example (1).sqlite`
- `example (2).sqlite`
- `example (3).sqlite`

## Root cause

The provider may apply its own collision suffix even after `renameDocument(...)` is requested.  
So a requested target name is not guaranteed to be the final persisted display name.

## Fix

In `FileUtils.maybeRenameGenericMimeDuplicate(...)`:

- Normalize/parse collision patterns (`name.ext (N)`, `name (N).ext`, plain `name.ext`).
- Recursively clean `baseName` suffixes to prevent nested accumulation.
- Add retry loop with bounded attempts.
- After each rename attempt, verify with `actualName == targetName`.
- If provider returns a different name (auto-suffixed), increment suffix and retry until a clean normalized name is achieved.
- Keep a final warning only when retries are exhausted.

## Result

Duplicate saves now converge to clean names (`example (2).sqlite`, `example (3).sqlite`, etc.) instead of nested names (`example (1) (1).sqlite`).

## Validation

Manually verified repeated saves against existing `example.sqlite`:
1. first duplicate -> `example (1).sqlite`
2. next duplicates -> `example (2).sqlite`, `example (3).sqlite`, ...
3. no nested collision suffixes observed

Closes: https://github.com/miguelpruivo/flutter_file_picker/issues/1947

Many thanks @Gabbar-v7 for your comments here https://github.com/miguelpruivo/flutter_file_picker/pull/1985#issuecomment-4251323738

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -363,9 +363,8 @@ object FileUtils {
         var success = false
         var currentSuffix = suffix ?: 0
         var attempts = 0
-        val maxAttempts = MAX_RENAME_ATTEMPTS
 
-        while (!success && attempts < maxAttempts) {
+        while (!success && attempts < MAX_RENAME_ATTEMPTS) {
             val targetName = when {
                 currentSuffix > 0 -> "$baseName ($currentSuffix).$extension"
                 else -> "$baseName.$extension"
@@ -396,10 +395,10 @@ object FileUtils {
             } catch (ex: Exception) {
                 currentSuffix++
                 attempts++
-                if (attempts >= maxAttempts) {
+                if (attempts >= MAX_RENAME_ATTEMPTS) {
                     Log.w(
                         TAG,
-                        "Failed to normalize saved document name from '$currentName' to '$targetName' after $maxAttempts attempts. MIME=$mimeType, error=$ex"
+                        "Failed to normalize saved document name from '$currentName' to '$targetName' after $MAX_RENAME_ATTEMPTS attempts. MIME=$mimeType, error=$ex"
                     )
                 }
             }
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -37,7 +37,6 @@ import java.io.InputStream
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
-import java.util.regex.Pattern
 
 object FileUtils {
     private const val TAG = "FilePickerUtils"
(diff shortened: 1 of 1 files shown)


android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -350,37 +349,91 @@ object FileUtils {
...
-        // Desired style "name (N).ext"; leave unchanged.
-        val alreadyNormalizedPattern = Regex("^.* \\((\\d+)\\)\\.${escapedExtension}$")
-        // Regular "name.ext"; leave unchanged.
-        val hasExtensionPattern = Regex("^.*\\.${escapedExtension}$")
-
-        val targetName = when {
-            duplicatedExtensionPattern.matches(currentName) -> {
-                val match = duplicatedExtensionPattern.matchEntire(currentName) ?: return uri
-                "${match.groupValues[1]} (${match.groupValues[2]}).$extension"
+        val androidCollisionRegex = Regex("^(.*)\\.$escapedExtension \\((\\d+)\\)$")
```
navaronbracke: Can you put a comment above `androidCollisionRegex` too? (The other two regex have an example comment as well)

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -350,37 +349,91 @@ object FileUtils {
...
+                val match = plainNameRegex.matchEntire(currentName)!!
+                match.groupValues[1] to null
+            }
+            else -> {
+                currentName to null
+            }
+        }
+
+        // Clean up baseName if it already has one or more " (M)" suffixes to prevent nesting
+        val baseNameSuffixRegex = Regex("^(.*) \\((\\d+)\\)$")
```
navaronbracke: This reduces `"foo.ext (M) (N) (O)"` to `"foo.ext (M)"` ?

Can we put this normalization into a new private method?

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -350,37 +349,91 @@ object FileUtils {
...
-            duplicatedExtensionPattern.matches(currentName) -> {
-                val match = duplicatedExtensionPattern.matchEntire(currentName) ?: return uri
-                "${match.groupValues[1]} (${match.groupValues[2]}).$extension"
+        val androidCollisionRegex = Regex("^(.*)\\.$escapedExtension \\((\\d+)\\)$")
+        // Desired style "name (N).ext"
+        val normalizedCollisionRegex = Regex("^(.*) \\((\\d+)\\)\\.$escapedExtension$")
+        // Regular "name.ext"
+        val plainNameRegex = Regex("^(.*)\\.$escapedExtension$")
+
+        var (baseName, suffix) = when {
```
navaronbracke: Can we put this into a new private method?

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -350,37 +349,91 @@ object FileUtils {
...
+            val baseSuffix = baseMatch.groupValues[2].toInt()
+            baseName = realBase
+            suffix = (suffix ?: 0) + baseSuffix
+        }
+
+        var finalUri = uri
+        var success = false
+        var currentSuffix = suffix ?: 0
+        var attempts = 0
+        val maxAttempts = 100
```
navaronbracke: Why is max attempts 100? That seems very high?

navaronbracke (changes requested):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Done!

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Done!

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Lowered to 20!

vicajilau:
I have tested and it is working as expected now:

image

Gabbar-v7:
It works. Thanks man.

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -350,39 +350,110 @@ object FileUtils {
...
+        // Clean up baseName if it already has one or more " (M)" suffixes to prevent nesting.
+        val normalized = normalizeBaseNameAndSuffix(baseName, suffix)
+        baseName = normalized.first
+        suffix = normalized.second
+
+        var finalUri = uri
+        var success = false
+        var currentSuffix = suffix ?: 0
+        var attempts = 0
+        val maxAttempts = MAX_RENAME_ATTEMPTS
```
navaronbracke: This local val is not needed?

navaronbracke (approved):
LGTM with nit

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: True, fixed!

PR #1988, opened 2026-04-16
Fix web default for pickFiles withData

### LINKED ISSUE #1987: withData no longer defaults to true on web
**Describe the bug**

Since v11.0.0, `withData` no longer defaults to `true` on web, despite the documentation stating: *"Defaults to `true` on web, `false` otherwise."*

The refactor from instance-based to static methods introduced a regression. `FilePicker.pickFiles()` declares `bool withData = false` and passes it explicitly to `FilePickerPlatform.instance.pickFiles()`, overriding the web implementation's default of `true`.

Before 11.0.0, calling `FilePicker.platform.pickFiles()` invoked the web implementation directly, which had `withData = true`. Now the static method always passes `false`, so `file.bytes` is always `null` on web unless `withData: true` is set explicitly.

**Platform**

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Desktop

**Platform OS version**

macOS 26.4.1 — Chrome 147.0.7727.57

**How are you picking?**

```dart
final result = await FilePicker.pickFiles(
  type: FileType.custom,
  allowedExtensions: ['pdf'],
);
// result.files.first.bytes is null on web (expected non-null)
```

1. Call `FilePicker.pickFiles()` on web without specifying `withData`
2. Access `result.files.first.bytes`
3. `bytes` is `null` — before 11.0.0 it contained the file data

**Root cause**

In `file_picker.dart`, the static method:
```dart
static Future<FilePickerResult?> pickFiles({
  bool withData = false, // ← always false
}) {
  return FilePickerPlatform.instance.pickFiles(
    withData: withData, // ← passes false explicitly, overriding web's default true
  );
}
```

In `file_picker_web.dart`, the default is `true` but never used:
```dart
Future<FilePickerResult?> pickFiles({
  bool withData = true, // ← dead default, always receives explicit false
}) async { ... }
```

**Flutter Version details**

```
Flutter 3.41.6 • channel stable
Dart 3.11.4
file_picker 11.0.2
```

**Additional context**

Workaround: pass `withData: true` explicitly on web calls.

vicajilau:
## Summary
- fix FilePicker.pickFiles to default withData to kIsWeb
- keep non-web default behavior unchanged (false)
- add changelog entry referencing #1987 and agreement from #1961

## Context
Per the agreement discussed in #1961, withData should default to true on web.
This PR aligns the public API default with that behavior.

Fixes #1987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Improved `saveFile` naming behavior when a duplicate file exists, normalizing duplicate names to keep the extension suffix (for example, `file (1).bak` instead of `file.bak (1)`). [#1947](https://github.com/miguelpruivo/flutter_file_picker/issues/1947)
 
 ### Web
-- Fixed `pickFiles()` default `withData` behavior in the public API to correctly default to `true` on web when not explicitly provided, as agreed when `cancelUploadOnWindowBlur` was introduced. [#1987](https://github.com/miguelpruivo/flutter_file_picker/issues/1987) [#1961](https://github.com/miguelpruivo/flutter_file_picker/issues/1961)
+- Fixed `pickFiles()` default `withData` behavior in the public API to correctly default to `true` on web when not explicitly provided. [#1987](https://github.com/miguelpruivo/flutter_file_picker/issues/1987)
 
 ## 11.0.2
 ### Android
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Android
 - Improved `saveFile` naming behavior when a duplicate file exists, normalizing duplicate names to keep the extension suffix (for example, `file (1).bak` instead of `file.bak (1)`). [#1947](https://github.com/miguelpruivo/flutter_file_picker/issues/1947)
 
+### Web
+- Fixed `pickFiles()` default `withData` behavior in the public API to correctly default to `true` on web when not explicitly provided, as agreed when `cancelUploadOnWindowBlur` was introduced. [#1987](https://github.com/miguelpruivo/flutter_file_picker/issues/1987) [#1961](https://github.com/miguelpruivo/flutter_file_picker/issues/1961)
+
 ## 11.0.2
 ### Android
 - Fixed a Path Traversal vulnerability (CWE-22) when resolving file paths from external content providers. [#1967](https://github.com/miguelpruivo/flutter_file_picker/issues/1967)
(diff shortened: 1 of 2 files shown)


lib/src/file_picker.dart
```diff
@@ -1,6 +1,6 @@
 import 'dart:async';
-import 'dart:typed_data';
 
+import 'package:flutter/foundation.dart';
 import 'package:file_picker/src/platform/file_picker_platform_interface.dart';
 import 'package:file_picker/src/api/file_picker_result.dart';
 import 'package:file_picker/src/api/file_picker_types.dart';
```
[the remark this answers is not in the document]
vicajilau: typed_data is included in foundation

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 ### Android
 - Improved `saveFile` naming behavior when a duplicate file exists, normalizing duplicate names to keep the extension suffix (for example, `file (1).bak` instead of `file.bak (1)`). [#1947](https://github.com/miguelpruivo/flutter_file_picker/issues/1947)
 
+### Web
+- Fixed `pickFiles()` default `withData` behavior in the public API to correctly default to `true` on web when not explicitly provided. [#1987](https://github.com/miguelpruivo/flutter_file_picker/issues/1987)
+
```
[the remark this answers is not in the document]
vicajilau: In the changelog we do not need to link the old one, only the current. The user will be able to read inside of the issue all the details.

lib/src/file_picker.dart
```diff
@@ -1,6 +1,6 @@
 import 'dart:async';
-import 'dart:typed_data';
 
+import 'package:flutter/foundation.dart';
```
navaronbracke, line 3: Interesting, the lint that usually flags unnecessary imports (because all used symbols are exported from another import), does not flag this one?

navaronbracke (approved):

lib/src/file_picker.dart
vicajilau: It is because foundation includes typed_data 😃

navaronbracke: No, I mean that this https://dart.dev/tools/diagnostics/unnecessary_import didn't trigger when I tried it in Dartpad

lib/src/file_picker.dart
```diff
@@ -62,7 +62,7 @@ abstract final class FilePicker {
     Function(FilePickerStatus)? onFileLoading,
     int compressionQuality = 0,
     bool allowMultiple = false,
-    bool withData = false,
+    bool withData = kIsWeb,
```
homeserve-lsaudon, line 65: smart 🧠

Issue #721, opened 2021-05-21
Get directory with write access on Android


minusium:
Hey. I have been writing an app, which asks user to select a directory. then the app put multiple files on that directory and these files are not media. I have a problem here, because your library doesn't have any parameter request these.
When I use `getDirectoryPath()`, I get a path for a directory, which can be wrong if the user selects a directory from a storage provider (like google drive), Also I can't write any file there, even if I request `WRITE_EXTERNAL_STORAGE` permission, I'll get `OS Error: Operation not permitted` error.
I'm using `MANAGE_EXTERNAL_STORAGE` for workaround but please add parameters to `getDirectoryPath()` to request persistable and write permission.

Look at this code, from [Signal Android App](https://github.com/signalapp/Signal-Android):
```java
Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
if (Build.VERSION.SDK_INT >= 26) {
    intent.putExtra(DocumentsContract.EXTRA_INITIAL_URI, SignalStore.settings().getLatestSignalBackupDirectory());
}
intent.addFlags(Intent.FLAG_GRANT_PERSISTABLE_URI_PERMISSION |                                     
                          Intent.FLAG_GRANT_WRITE_URI_PERMISSION |
                          Intent.FLAG_GRANT_READ_URI_PERMISSION);
try {
    fragment.startActivityForResult(intent, requestCode);
} catch (ActivityNotFoundException e) {                 
    Toast.makeText(fragment.requireContext(), R.string.BackupDialog_no_file_picker_available, Toast.LENGTH_LONG).show();
}
```

miguelpruivo:
I'll take a look into it. Thank you!

brightseagit:
Hello, @miguelpruivo do you have any updates on this issue.
Also, recently Google stated:
> Starting August 2021, new apps will be required to target API level 30 (Android 11) and use the Android App Bundle publishing format. Starting November 2021, all app updates will be required to target API level 30 (Android 11). 

So with this change we were no longer able to use api 29 to bypass these restrictions.

miguelpruivo:
@brightseagit so, there is nothing we can do here?

brightseagit:
[...]

PR #1989, opened 2026-04-18
Implement Storage Access Framework (SAF) with persistable URI grants

### LINKED ISSUE #1825: Feature request – Support Storage-Access-Framework *persistable* URI grants on Android 10 +
## Is your feature request related to a problem?
Yes. On Android 10 (API 29) and above, any URI returned by an `ACTION_OPEN_DOCUMENT*` intent becomes inaccessible after a reboot **unless** the app calls `ContentResolver.takePersistableUriPermission(uri, flags)`.

Without this capability, developers who need a durable music- or document library must either:

- Request the Play-Store-sensitive **MANAGE_EXTERNAL_STORAGE** permission, or
- Depend on a second plugin (`saf`, `shared_storage`) just to persist the grant.

Both alternatives add review friction or rely on libraries that are no longer actively maintained.

Official guidance: <https://developer.android.com/training/data-storage/shared/documents-files>

---

### 1 · Android-only options object

```dart
enum SafGrant  { transient, persist }      // lifetime
enum SafAccess { readOnly, readWrite }     // mode

class AndroidSafOptions {
  final SafGrant  grant;        // default → SafGrant.transient
  final SafAccess access;       // default → SafAccess.readOnly
  final bool autoPersist;       // default → true
  const AndroidSafOptions({
    this.grant = SafGrant.transient,
    this.access = SafAccess.readOnly,
    this.autoPersist = true,
  });
}

### 2 · Usage examples

```dart
// Pick multiple files and keep read access across reboots
  allowMultiple: true,
  androidSafOptions: const AndroidSafOptions(
    grant: SafGrant.persist,
    access: SafAccess.readOnly,
  ),
);

// Let the user choose a folder we can also write to
final folderUri = await FilePicker.platform.getDirectoryPath(
  androidSafOptions: const AndroidSafOptions(
    grant: SafGrant.persist,
    access: SafAccess.readWrite,
  ),
);

### 3 · Rich result type

```dart
class SafHandle {
  final Uri uri;               // e.g. content://com.android.externalstorage…
  final SafAccess access;      // readOnly / readWrite
  Future<void> release();      // relinquish grant
}

class PlatformFile {
  final String?   path;        // unchanged
  final SafHandle? safHandle;  // new
}
```
### 4 · Helper utilities (optional)

```dart
class FilePickerAndroidSaf {
  static Future<List<Uri>> listPersistedUris();
  static Future<void> releaseAll();
}
```

### 5 · Describe alternatives you’ve considered

- **`saf`** – last released **two years** ago; it fails to build on current Flutter (3.22+) because of outdated Android support libraries.
- **`shared_storage`** – functionally solid but now marked **“discontinued”**; its last update was 16 months ago, so future maintenance is uncertain.
- **`media_store_plus`** – works well for files inside the standard `Music/` collection, but **cannot** open arbitrary folders selected by the user.

Bringing this capability directly into **file_picker** would consolidate everything into a single, actively maintained dependency.

### 6 · Additional context

- Android 13 introduced granular `READ_MEDIA_*` permissions, but they still don’t allow directory traversal. Persistable SAF grants remain Google’s recommended method for long-term access to user-selected files and folders.
- Apps can hold **512** persisted grants on API 30+, and **128** on older versions.
- The proposed `AndroidSafOptions` mirrors the plugin’s existing platform-specific options style (e.g., `lockParentWindow` for Windows).

Related closed issue: [#721](https://github.com/miguelpruivo/flutter_file_picker/issues/721) – “Get directory with write access on Android” (closed as stale but continues to receive interest).

### 7 · Are you willing to contribute?

Yes. I will wait for maintainer feedback before opening a pull request. Once approved, I can:

1. Add `AndroidSafOptions`, `SafHandle`, and helper utilities.
2. Implement `takePersistableUriPermission` / `releasePersistableUriPermission` in the Android code.
3. Update the example app, documentation, and include an integration test on an Android 14 emulator.

Thank you!

vicajilau:
# PR Description: Android SAF Persistable URI Grants

## Description
This PR implements support for **Android Storage Access Framework (SAF) Persistable URI Grants**, allowing applications to maintain long-term access to files and directories even after device reboots.

### Key Changes
- **Configuration API**: Introduced `AndroidSAFOptions` to configure grant types (`transient` vs `lifetime`) and access modes (`readOnly` vs `readWrite`).
- **Platform Integrity**: Followed the suggested refactoring by:
    - Using explicit `AndroidSAF` prefixes.
    - Implementing `final class` for API models.
    - Moving `MethodChannel` logic to the platform interface layer.
- **Model Separation**: Created `AndroidPlatformFile` (a subclass of `PlatformFile`) to hold SAF-specific data without polluting the generic base class.
- **Native Implementation**: Updated `FileUtils.kt` and `FilePickerDelegate.kt` to handle `takePersistableUriPermission` and provide a mechanism to release grants.
- **Example App Improvements**: Updated the demo app with:
    - UI controls for SAF options.
    - Interactive "Release Grant" functionality that dynamically updates the list.
    - Clearer feedback on SAF status.

## Related Issues
- Fixes #1825
- Addresses #721

## Verification performed
- [x] `flutter analyze` passes with no issues.
- [x] Verified native logic for permission handling.
- [x] Manual verification in the example app for picking files/folders and releasing permissions.

---
The implementation respects the federated plugin architecture and maintains backward compatibility for other platforms.


android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -54,23 +54,47 @@ object FileUtils {
...
+                 if (isPersist && autoPersist) {
+                     try {
+                         val flags = if (isReadWrite) {
+                            android.content.Intent.FLAG_GRANT_READ_URI_PERMISSION or android.content.Intent.FLAG_GRANT_WRITE_URI_PERMISSION
+                         } else {
+                            android.content.Intent.FLAG_GRANT_READ_URI_PERMISSION
+                         }
+                         activity.contentResolver.takePersistableUriPermission(uri, flags)
+                     } catch (e: SecurityException) {
+                         Log.e(TAG, "Failed to take persistable URI permission for $uri", e)
```
navaronbracke, line 85: Should we pipe this through result.error() ? Or can the user still do thing with the URI if this fails?

lib/src/api/android_saf_handle.dart
```diff
@@ -0,0 +1,32 @@
+import '../../file_picker.dart';
```
navaronbracke: Do we use relative uris in other places?

lib/src/api/android_saf_handle.dart
```diff
@@ -0,0 +1,32 @@
...
+import '../platform/file_picker_platform_interface.dart';
+
+/// An opaque handle to an Android Storage Access Framework [uri].
+final class AndroidSAFHandle {
+  const AndroidSAFHandle({
+    required this.accessMode,
+    required this.uri,
+  });
+
+  factory AndroidSAFHandle.fromMap(Map<String, dynamic> map) {
```
navaronbracke:
```suggestion
  factory AndroidSAFHandle.fromMap(Map<String, Object?> map) {
```

example/lib/src/file_picker_demo.dart
```diff
@@ -102,21 +112,46 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-          );
-        },
-      );
+
+      void updateResults() {
+        _resultsWidget = _buildFilePickerResultsWidget(
+          itemCount: pickedFiles?.length ?? 0,
+          itemBuilder: (BuildContext context, int index) {
+            final path =
+                pickedFiles!.map((e) => e.path).toList()[index].toString();
```
navaronbracke: This line reads a little messy.

- what if picked files is null or empty
- we could omit the to list with a for literal? (map files to paths)
- string interpolation, instead of toString() ?

example/lib/src/file_picker_demo.dart
```diff
@@ -102,21 +112,46 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+            return ListTile(
+              leading: Text(
+                index.toString(),
+                style:
```
navaronbracke: Formatting is a bit meh here?

example/lib/src/file_picker_demo.dart
```diff
@@ -102,21 +112,46 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+                    const TextStyle(fontSize: 15, fontWeight: FontWeight.bold),
+              ),
+              title: Text(
+                  "File path (SAF Grant: ${pickedFiles![index] is AndroidPlatformFile}):"),
```
navaronbracke: `pickedFiles![index] is AndroidPlatformFile` is duplicated multiple times?

example/lib/src/file_picker_demo.dart
```diff
@@ -102,21 +112,46 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-              style: TextStyle(fontSize: 15, fontWeight: FontWeight.bold),
-            ),
-            title: Text("File path:"),
-            subtitle: Text(path),
-          );
-        },
-      );
+
+      void updateResults() {
+        _resultsWidget = _buildFilePickerResultsWidget(
```
navaronbracke: Should we put the results wdget in its own file? It is getting a bit big.

example/lib/src/file_picker_demo.dart
```diff
@@ -102,21 +112,46 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+            return ListTile(
+              leading: Text(
+                index.toString(),
+                style:
+                    const TextStyle(fontSize: 15, fontWeight: FontWeight.bold),
+              ),
+              title: Text(
+                  "File path (SAF Grant: ${pickedFiles![index] is AndroidPlatformFile}):"),
+              subtitle: Text(path),
+              trailing: pickedFiles![index] is AndroidPlatformFile
```
navaronbracke: I'd make a separate nullable var for the trailing widget, at the top of the itemBuilder method, and build it once, then reference it here.

That would clean up the formatting, too.

example/lib/src/file_picker_demo.dart
```diff
@@ -215,15 +258,40 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-        itemBuilder: (BuildContext context, int index) {
-          return ListTile(
-            title: const Text('Directory path:'),
-            subtitle: Text(pickedDirectoryPath ?? ''),
-          );
-        },
-      );
+
+      void updateResults() {
+        _resultsWidget = _buildFilePickerResultsWidget(
```
navaronbracke: Should this also live as its own widget, in a new file?

example/lib/src/file_picker_demo.dart
```diff
@@ -215,15 +258,40 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+          itemCount: pickedDirectoryPath != null ? 1 : 0,
+          itemBuilder: (BuildContext context, int index) {
+            return ListTile(
+              title: const Text('Directory path:'),
+              subtitle: Text(pickedDirectoryPath ?? ''),
+              trailing: pickedDirectoryPath != null &&
```
navaronbracke: Put the icon button in a separate var, pickedDirectoryPath can be a local variable, so we get null type promotion, and the SAF handle could be a final variable? (once this is its own widget)

example/lib/src/file_picker_demo.dart
```diff
@@ -215,15 +258,40 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
+                      pickedDirectoryPath!.startsWith('content://')
+                  ? IconButton(
+                      icon: const Icon(Icons.delete_forever),
+                      onPressed: () {
+                        AndroidSAFHandle(
+                          uri: Uri.parse(pickedDirectoryPath!),
+                          accessMode: AndroidSAFAccessMode.readWrite,
```
navaronbracke: Is this really always readWrite? I don't believe it :p

example/lib/src/file_picker_demo.dart
```diff
@@ -425,6 +493,30 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
                         value: _multiPick,
                       ),
                     ),
+                    ConstrainedBox(
+                      constraints: const BoxConstraints.tightFor(width: 400.0),
+                      child: SwitchListTile.adaptive(
+                        title: Text(
+                          'SAF Persist (Android 10+)',
```
navaronbracke: Do we need a support query? Do people with Android 9 don't have to do some fiddely code checks before using it? Or do we no-op / throw an error?

(we can add that later, though)

lib/src/api/android_saf_options.dart
```diff
@@ -0,0 +1,59 @@
...
+  lifetime(1);
+
+  const AndroidSAFGrant(this.rawValue);
+
+  /// The raw value of this grant.
+  final int rawValue;
+}
+
+/// The configuration options for working with Android's Storage Access Framework.
+/// Only applicable when reading APIs on Android 10+ (API 29+).
```
navaronbracke:
```suggestion
/// Only supported on Android 10+ (API 29+).
```

lib/src/api/android_saf_options.dart
```diff
@@ -0,0 +1,59 @@
...
+  /// The access mode to use with the Android Storage Access Framework.
+  ///
+  /// Defaults to [AndroidSAFAccessMode.readOnly].
+  final AndroidSAFAccessMode accessMode;
+
+  /// Whether to automatically execute `takePersistableUriPermission` on the native side.
+  /// Defaults to `true`.
+  final bool persistGrant;
+
+  Map<String, dynamic> toMap() {
```
navaronbracke:
```suggestion
  Map<String, Object?> toMap() {
```

lib/src/api/platform_file.dart
```diff
@@ -14,14 +16,24 @@ class PlatformFile {
...
+    final file = PlatformFile(
       name: data['name'],
       path: data['path'],
       bytes: data['bytes'],
       size: data['size'],
       identifier: data['identifier'],
       readStream: readStream,
     );
+
+    if (data.containsKey('safHandle') && data['safHandle'] != null) {
```
navaronbracke: Use in if-case expression?

```
if (data case { 'safHandle': final Map<String, Object?> safHandle }) {
  // ...
}
```

lib/src/platform/web/file_picker_web.dart
```diff
@@ -38,6 +39,16 @@ class FilePickerWeb extends FilePickerPlatform {
     return target;
   }
 
+  @override
+  Future<String?> getDirectoryPath({
```
navaronbracke, line 43: Good catch!

lib/src/platform/linux/file_picker_linux.dart
```diff
@@ -44,6 +45,7 @@ class FilePickerLinux extends FilePickerPlatform {
     bool readSequential = false,
     int compressionQuality = 0,
     bool cancelUploadOnWindowBlur = true,
+    AndroidSAFOptions? androidSafOptions,
```
navaronbracke, line 48: One of the things that we should try and address with the API refactor is having "Options" classes, so adding a new field is not a breaking change.

Also, we should investigate if we can have platform specific features like these, available only in the relevant platforms. (for example the releaseSafGrant method should only exist in a future FilePickerAndroid implementation)

lib/src/api/platform_file.dart
```diff
@@ -100,3 +112,41 @@ class PlatformFile {
     return 'PlatformFile(${kIsWeb ? '' : 'path $path'}, name: $name, bytes: $bytes, readStream: $readStream, size: $size)';
   }
 }
+
+/// An extension of [PlatformFile] that includes Android's Storage Access Framework
```
navaronbracke:
```suggestion
/// A [PlatformFile] implementation that includes a handle to a Android's Storage Access Framework document URI.
```

lib/src/api/platform_file.dart
```diff
@@ -100,3 +112,41 @@ class PlatformFile {
...
+  }) : super(
+          path: file.path,
+          name: file.name,
+          size: file.size,
+          bytes: file.bytes,
+          readStream: file.readStream,
+          identifier: file.identifier,
+        );
+
+  /// The SAF handle representing the URI grant on Android in case it's a persistent grant.
```
navaronbracke: If the grant is not persist, then the user does not get a AndroidPlatformFile? So that is redundant to mention?

```suggestion
  /// The handle to the Storage Access Framework URI.
```

lib/src/api/platform_file.dart
```diff
@@ -100,3 +112,41 @@ class PlatformFile {
...
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    return super == other &&
+        other is AndroidPlatformFile &&
+        other.safHandle == safHandle;
+  }
+
+  @override
+  int get hashCode {
+    return super.hashCode ^ safHandle.hashCode;
```
navaronbracke: Use Object.hash()?

lib/src/api/platform_file.dart
```diff
@@ -100,3 +112,41 @@ class PlatformFile {
...
+          identifier: file.identifier,
+        );
+
+  /// The SAF handle representing the URI grant on Android in case it's a persistent grant.
+  /// Available if `AndroidSAFOptions` enabled `grant: AndroidSAFGrant.persist`.
+  final AndroidSAFHandle safHandle;
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
```
navaronbracke, line 136: Can we bail out quickly if other is not an AndroidPlatformFile ?

navaronbracke (changes requested):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Even if `takePersistableUriPermission` fails with a `SecurityException`, the user **can still do things with the URI**. The app still retains the temporary read/write permissions granted by the system's `Intent` that originally returned the URI. 

Because of this temporary permission, `flutter_file_picker` can still successfully read the file contents, copy it to the cache directory, and return a valid result to Flutter for immediate use. 

If we pipe this through `result.error()`, the entire file picking operation will abort, and the Flutter side will just get an exception. The user's selection would be discarded entirely, which is an aggressive degradation and a frustrating user experience.

lib/src/api/android_saf_handle.dart
vicajilau: Fixed!

lib/src/api/platform_file.dart
vicajilau: Good catch! Fixed

lib/src/api/platform_file.dart
vicajilau: Absolutely!!

lib/src/platform/linux/file_picker_linux.dart
vicajilau: So as not to confuse what you're saying about this new feature, after this PR I'd love to give you a preliminary overview of what you're talking about

example/lib/src/file_picker_demo.dart
vicajilau: I just ran `dart format .` 😄

example/lib/src/file_picker_demo.dart
vicajilau: You are right! Changed

lib/src/platform/linux/file_picker_linux.dart
vicajilau: Once this pull request has been merged, it would be great if you could take a look at this: https://github.com/miguelpruivo/flutter_file_picker/pull/1990

navaronbracke: I'll get to that next week!

example/lib/src/file_picker_demo.dart
vicajilau: Refactored!

example/lib/src/file_picker_demo.dart
vicajilau: I have refactored the example project a little bit

example/lib/src/file_picker_demo.dart
vicajilau: Refactored example project

example/lib/src/file_picker_demo.dart
vicajilau: Same!

example/lib/src/file_picker_demo.dart
vicajilau: Done! Same in example project

lib/src/api/platform_file.dart
vicajilau: Good catch!!

vicajilau:
I have reviewed each comment @navaronbracke 🚀

example/android/app/src/main/kotlin/com/mr/flutter/plugin/filepicker/file_picker_example/MainActivity.kt
```diff
@@ -1,5 +1,22 @@
...
+class MainActivity : FlutterActivity() {
+	override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+		super.configureFlutterEngine(flutterEngine)
+
+		MethodChannel(
+			flutterEngine.dartExecutor.binaryMessenger,
+			"com.mr.flutter.plugins.filepicker/android_saf_support",
+		).setMethodCallHandler { call, result ->
+			when (call.method) {
+				"getSdkInt" -> result.success(Build.VERSION.SDK_INT)
```
navaronbracke: Do we really need this in the example app? In the long term I would put it in a `FilePickerAndroid` implementation on the Dart side (which forwards to this exact code, but in FilePickerPlugin.kt.

For now I would just return a `result.error()`, with an unsupported error code, in FilePickerPlugin.kt.

example/lib/src/file_picker_demo.dart
```diff
@@ -14,6 +19,10 @@ class FilePickerDemo extends StatefulWidget {
 }
 
 class _FilePickerDemoState extends State<FilePickerDemo> {
+  static const MethodChannel _androidSafSupportChannel = MethodChannel(
```
navaronbracke: This part and related code can be removed, per my other comment.

example/lib/src/file_picker_demo.dart
```diff
@@ -288,260 +340,38 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-                            )
-                          : _resultsWidget,
-                ),
-                const SizedBox(height: 10.0),
-              ],
-            ),
-          ),
-        ),
-      ),
+      scaffoldKey: _scaffoldKey,
```
navaronbracke: Why do we have all this redirection? This was fine as it was, although we could make the children inside the view local variables in the build method, to improve formatting:

- The Wrap widgets could be separate final fields
- The _pickingType Dropdown field could use a for literal for the items and could be a separate final in the build method
- `_pickingType == FileType.custom` should not create an empty sized box (this could have been a collection-if)
- There is a redundant Builder() widget? The inner context is never used?
   - the loading indicator/user aborted widgets could be separate final fields
   - the ternaries are redundant? Just use if-else since this is valid to do in a collection statement

example/lib/src/file_picker_demo_view.dart
```diff
@@ -0,0 +1,341 @@
+import 'package:file_picker/file_picker.dart';
```
navaronbracke: This widget can be removed (see my other comment as to why we don't really need it)

example/lib/src/file_picker_results.dart
```diff
@@ -0,0 +1,24 @@
...
+    required this.itemBuilder,
+  });
+
+  final int itemCount;
+  final IndexedWidgetBuilder itemBuilder;
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      height: MediaQuery.sizeOf(context).height * 0.50,
```
navaronbracke, line 16: FractionallySizedBox?

example/lib/src/picked_directory_result.dart
```diff
@@ -0,0 +1,53 @@
...
+                    accessMode: accessMode,
+                  );
+                  safHandle.releaseGrant();
+                  onDirectoryRemoved();
+                },
+              )
+            : null;
+
+        return ListTile(
+          title: const Text('Directory path:'),
```
navaronbracke: Is it really a directory though? :P

example/lib/src/picked_files_results.dart
```diff
@@ -0,0 +1,48 @@
...
+
+class PickedFilesResults extends StatelessWidget {
+  const PickedFilesResults({
+    super.key,
+    required this.pickedFiles,
+    required this.onRemoveAndroidFile,
+  });
+
+  final List<PlatformFile>? pickedFiles;
+  final void Function(int index, AndroidPlatformFile androidPlatformFile)
```
navaronbracke: Private typedef, so the formatting is nicer?

example/lib/src/picked_files_results.dart
```diff
@@ -0,0 +1,48 @@
...
+      itemCount: pickedFiles?.length ?? 0,
+      itemBuilder: (BuildContext context, int index) {
+        final pickedFile = pickedFiles![index];
+        final AndroidPlatformFile? androidPlatformFile =
+            pickedFile is AndroidPlatformFile ? pickedFile : null;
+        final Widget? trailingWidget = androidPlatformFile == null
+            ? null
+            : IconButton(
+                icon: const Icon(Icons.delete_forever),
+                onPressed: () =>
```
navaronbracke, line 33: Use `() {}` for formatting reasons?

lib/src/api/android_saf_options.dart
```diff
@@ -0,0 +1,59 @@
...
+  ///
+  /// Defaults to [AndroidSAFGrant.transient].
+  final AndroidSAFGrant grant;
+
+  /// The access mode to use with the Android Storage Access Framework.
+  ///
+  /// Defaults to [AndroidSAFAccessMode.readOnly].
+  final AndroidSAFAccessMode accessMode;
+
+  /// Whether to automatically execute `takePersistableUriPermission` on the native side.
```
navaronbracke:
```suggestion
  /// Whether to persist the [grant], so that it is preserved across device reboots.
  ///
  /// Defaults to `true`.
```

lib/src/api/platform_file.dart
```diff
@@ -87,16 +98,53 @@ class PlatformFile {
...
   }
 
   @override
   String toString() {
     return 'PlatformFile(${kIsWeb ? '' : 'path $path'}, name: $name, bytes: $bytes, readStream: $readStream, size: $size)';
   }
 }
+
+/// A [PlatformFile] implementation that includes a handle to a Android's Storage Access Framework document URI.
+/// specifics, returned when picking files on Android 10+ with SAF options enabled.
```
navaronbracke: The last part of this paragraph is incorrect?

```suggestion
```

lib/src/api/platform_file.dart
```diff
@@ -87,16 +98,53 @@ class PlatformFile {
...
+          path: file.path,
+          name: file.name,
+          size: file.size,
+          bytes: file.bytes,
+          readStream: file.readStream,
+          identifier: file.identifier,
+        );
+
+  /// The handle to the Storage Access Framework URI.
+  /// Available if `AndroidSAFOptions` enabled `grant: AndroidSAFGrant.persist`.
```
navaronbracke: If you get a `AndroidPlatformFile`, you always have the revokable grant?
If it was transient, should you not have a `AndroidPlatformFile` in the first place?

```suggestion
```

navaronbracke (changes requested):

example/android/app/src/main/kotlin/com/mr/flutter/plugin/filepicker/file_picker_example/MainActivity.kt
vicajilau: Agree! Fixed :)

example/lib/src/file_picker_demo.dart
vicajilau: Done!

example/lib/src/file_picker_demo.dart
vicajilau: Fixed!!

example/lib/src/file_picker_results.dart
vicajilau: In this specific layout, I’d keep the current SizedBox height instead of switching to FractionallySizedBox.

The results list sits inside a Column within a SingleChildScrollView, so the vertical constraints are effectively unbounded. FractionallySizedBox depends on bounded parent constraints to resolve its height factor, which can lead to unstable/unintended sizing here.
Using SizedBox(height: MediaQuery.sizeOf(context).height * 0.5) keeps the ListView explicitly bounded and predictable.

If we want to improve it, I can clamp the height (for example min/max) rather than change to FractionallySizedBox

example/lib/src/picked_files_results.dart
vicajilau: Fixed!

example/lib/src/picked_files_results.dart
vicajilau: I used the arrow form because this callback is a single expression, but I can switch to () { ... } if you prefer it for consistency/readability

example/lib/src/file_picker_demo_view.dart
vicajilau: Fixed!

vicajilau:
@navaronbracke once more time!! 🚀

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -55,8 +57,14 @@ class MethodChannelFilePicker extends FilePickerPlatform {
         withData,
         withReadStream,
         compressionQuality,
+        androidSafOptions,
       );
 
+  @override
+  Future<void> releaseSAFGrant(String uri) async {
```
navaronbracke, line 64: I wonder if the method name should match here?

lib/src/api/android_saf_options.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  /// The access mode to use with the Android Storage Access Framework.
+  ///
+  /// Defaults to [AndroidSAFAccessMode.readOnly].
+  final AndroidSAFAccessMode accessMode;
+
+  /// Whether to persist the [grant], so that it is preserved across device reboots.
+  ///
+  /// Defaults to `true`.
+  /// Defaults to `true`.
```
navaronbracke: This is repeated?

lib/src/api/android_saf_options.dart
```diff
@@ -0,0 +1,61 @@
+/// The access mode that is coupled to an [AndroidSAFGrant].
+enum AndroidSAFAccessMode {
+  /// Only allow read permission for a URI.
+  readOnly(0),
```
navaronbracke: I noticed that both of these pass the enum name to the API and not the index?

If the enum is renamed then this would break.

On the other hand the names follow the Android values, so the index would be redundant?

1. Remove the `rawValue`
2. Or, pass the index instead of name?

example/lib/src/picked_directory_result.dart
```diff
@@ -0,0 +1,55 @@
...
+                    accessMode: accessMode,
+                  );
+                  safHandle.releaseGrant();
+                  onDirectoryRemoved();
+                },
+              )
+            : null;
+
+        return ListTile(
+          title: Text(isContentUri ? 'Directory URI:' : 'Directory path:'),
```
navaronbracke:
```suggestion
          title: Text(isContentUri ? 'Content URI:' : 'Filesystem path:'),
```

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerPlugin.kt
```diff
@@ -130,6 +130,17 @@ class FilePickerPlugin : MethodCallHandler, FlutterPlugin,
                 result.success(activity?.applicationContext?.let { clearCache(it) })
             }
 
+            "releaseSafHandle" -> {
+                val uriStr = arguments?.get("uri") as? String
```
navaronbracke, line 134: Should we do nothing if the SDK INT is too low for SAF?

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
navaronbracke, line 85: @vicajilau One thing I forgot: Does the SecurityException happen on Android 9? (where SAF is not supported)

navaronbracke (changes requested):
Just some minor things left!

I do like how the example app got way more readable!

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Regarding Android 9: I believe you might be confusing SAF with **Scoped Storage**. SAF (Storage Access Framework) and `takePersistableUriPermission` were actually introduced all the way back in Android 4.4 (API 19), so SAF is fully supported on Android 9. Scoped Storage is what was introduced later in Android 10 (API 29).

As for the `SecurityException`, it can definitely happen on any Android version (API 19 to 34) and we must catch it to prevent app crashes for three main reasons:
1. When picking specific media types (`image/*`, `video/*`, `audio/*`), the plugin intentionally uses `Intent.ACTION_GET_CONTENT`. URIs returned from this intent do not support persistable permissions, so calling `takePersistableUriPermission` on them will reliably throw a `SecurityException`.
2. The OS enforces a strict concurrent limit on persisted URIs per app (128 on Android 10 and below, 512 on Android 11+). If the app hits this limit, the system throws a `SecurityException`.
3. Some OEM-modified file explorers occasionally fail to attach the persistable grant flag to the returned intent.

Catching it ensures a graceful fallback: the app just keeps the default temporary read/write access for the current session instead of crashing

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerPlugin.kt
vicajilau: True, fixed!

lib/src/api/android_saf_options.dart
vicajilau: Yes... Fixed!

lib/src/api/android_saf_options.dart
vicajilau: Fixed!!

lib/src/platform/file_picker_method_channel.dart
vicajilau: Although it isn't necessary, it's good practice for the method in Dart and the Method Channel to call the same thing. Fixed!!

vicajilau:
I've just tested it thoroughly and can confirm that it works @navaronbracke with the last changes 🚀

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerPlugin.kt
```diff
@@ -130,6 +130,22 @@ class FilePickerPlugin : MethodCallHandler, FlutterPlugin,
                 result.success(activity?.applicationContext?.let { clearCache(it) })
             }
 
+            "releaseSafGrant" -> {
+                val uriStr = arguments?.get("uri") as? String
+                if (uriStr != null) {
```
navaronbracke:
```suggestion
                if (uriStr == null) {
                  result.success(null)
                  return
                }
                
                try {
                 ...
                }
```

lib/src/api/platform_file.dart
```diff
@@ -14,14 +16,24 @@ class PlatformFile {
...
+    final file = PlatformFile(
       name: data['name'],
       path: data['path'],
       bytes: data['bytes'],
       size: data['size'],
       identifier: data['identifier'],
       readStream: readStream,
     );
+
+    final safHandle = data['safHandle'];
```
navaronbracke: Didn't this have an if-case before? Does it work if we use the if-case with Map<Object?, Object?>

navaronbracke (approved):
LGTM with nits

lib/src/api/platform_file.dart
vicajilau: Yes, now that is restored. Everything works as expected!

PR #1990, opened 2026-04-19
Refactor: Introduce FilePickerOptions to encapsulate platform properties

vicajilau:
This PR is a first draft addressing the feedback in https://github.com/miguelpruivo/flutter_file_picker/pull/1989#discussion_r3104886821.

### Design Decisions

In this first approach, we have taken the following design decisions to improve API scalability and prevent future breaking changes:

1. **New `FilePickerOptions` Abstraction**: Platform-specific properties have been deprecated in the root `pickFiles` method and grouped inside a new `FilePickerOptions` wrapper (with properties like `webOptions`, `androidOptions`, etc.). The DartDoc documentation has been explicitly updated for each deprecated parameter to state exactly which new abstraction they map to, avoiding any confusion.
2. **Shared Properties Remain at the Root**: Properties that are common across multiple platforms (e.g., `lockParentWindow` which applies to both Windows and Linux desktop) have been kept at the root level to avoid parameter duplication. Their documentation has been updated to state clearly which platforms they affect and which ones ignore them.

This ensures smoother scalability for future platform-specific features without introducing breaking changes to the main API signature.


lib/src/platform/linux/file_picker_linux.dart
navaronbracke, line 38: @vicajilau The desktop window specific options (dialogTitle / lockParentWindow) should probably go into a `class FilePickerDesktopOptions`, from which the Linux,MacOS,Windows options inherit?

lib/src/api/file_picker_options.dart
```diff
@@ -0,0 +1,70 @@
+import 'package:file_picker/src/api/android_saf_options.dart';
+
+/// Cross-platform wrapper for all platform-specific options.
+///
+/// This groups all platform-specific configurations into a single object, mitigating
+/// breaking changes when new options or new platforms are added.
+class FilePickerOptions {
+  const FilePickerOptions({
```
navaronbracke:
```suggestion
  /// Create a new [FilePickerOptions] instance.
  const FilePickerOptions({
```

lib/src/api/file_picker_options.dart
```diff
@@ -0,0 +1,70 @@
...
+    this.androidOptions = const FilePickerAndroidOptions(),
+    this.iosOptions = const FilePickerIosOptions(),
+    this.webOptions = const FilePickerWebOptions(),
+    this.windowsOptions = const FilePickerWindowsOptions(),
+    this.macosOptions = const FilePickerMacosOptions(),
+    this.linuxOptions = const FilePickerLinuxOptions(),
+  });
+
+  final FilePickerAndroidOptions androidOptions;
```
navaronbracke, line 26: Add doc comments for these?

lib/src/api/file_picker_options.dart
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/// Platform-specific options for Web.
+class FilePickerWebOptions {
+  const FilePickerWebOptions({
+    this.readSequential = false,
+    this.cancelUploadOnWindowBlur = true,
+  });
+
+  /// Keeps the file loading sequentially if true.
```
navaronbracke: Can this doc comment be improved? The flag mentions `read`, but the comment says `loading` ?

Also document the default value

lib/src/api/file_picker_options.dart
```diff
@@ -0,0 +1,70 @@
...
+  final bool readSequential;
+
+  /// Prevents upload cancellation when window focus is lost.
```
navaronbracke:
```suggestion
  /// Whether to prevent upload cancellation when window focus is lost.
```

Also document the default value

lib/src/api/file_picker_options.dart
vicajilau: Fixed!

lib/src/api/file_picker_options.dart
vicajilau: Fixed!

lib/src/platform/linux/file_picker_linux.dart
vicajilau: @navaronbracke Currently, the convention we follow is that when a parameter affects more than one platform (even if not all of them), we keep it as a top-level parameter and explicitly state in the documentation which platforms it applies to and which ones ignore it. 

I don't think it makes sense to group `dialogTitle` and `lockParentWindow` into a `FilePickerDesktopOptions` class because neither of these parameters actually affect *all* desktop platforms. Specifically, they are both ignored on macOS.

Regarding the viability of supporting them on macOS in the future: because the plugin presents the `NSOpenPanel` as a sheet (`beginSheetModal(for:)`), macOS inherently locks the parent window by default, making `lockParentWindow` redundant. Furthermore, Apple's official documentation explicitly states that titles are not displayed when the panel is presented as a sheet. Therefore, `dialogTitle` cannot be supported on macOS unless we completely change the native implementation to open as a standalone window instead of a sheet, which might degrade the native user experience.

Given that these options aren't truly 'Desktop-wide', I believe the current approach of keeping them as top-level parameters with clear documentation is the most semantically correct option, while also avoiding breaking changes and unnecessary boilerplate for developers.

vicajilau:
Hey @navaronbracke ,

I'm currently looking into migrating the `androidSafOptions` parameter inside `getDirectoryPath` into an options wrapper, similar to what we did for `pickFiles` to keep the API clean. 

However, I have a concern: if we reuse the exact same `FilePickerOptions` class for `getDirectoryPath`, developers will see autocompletion for properties that don't apply there (like `webOptions`, which contains `readSequential`), and might get confused or frustrated when they are ignored.

I see three main ways we can handle this, and I'd love to hear your thoughts:

## Option 1: Reuse `FilePickerOptions` with Assertions
We keep a single `FilePickerOptions` class across all methods to minimize boilerplate. To prevent developers from passing unsupported options, we add assertions inside the methods. For example, in `getDirectoryPath`:
```dart
assert(options.webOptions == const FilePickerWebOptions(), 'webOptions are not supported for getDirectoryPath and will be ignored.');
```
*Pros:* Keeps the API surface very small and warns developers during debug if they pass unsupported options.
*Cons:* The IDE will still suggest the irrelevant parameters.

## Option 2: Reuse `FilePickerOptions` without Assertions (Silent Ignore)
We use the exact same `FilePickerOptions` class everywhere, but we simply ignore the properties that don't apply to the method being called, without throwing any assertions.
*Pros:* The absolute simplest approach with the fewest classes to maintain.
*Cons:* High risk of developers passing unsupported parameters, getting confused when they do nothing, and opening issues on GitHub complaining that it doesn't work.

## Option 3: Method-Specific Option Wrappers
We introduce strictly typed wrappers for each method. For example, for `getDirectoryPath`, we would create a `GetDirectoryPathOptions` class containing its own set of platform-specific classes. Platforms that don't currently have applicable options for this method would simply use an empty class to leave room for future scalability.
```dart
class GetDirectoryPathOptions {
  final GetDirectoryPathAndroidOptions androidOptions;
  final GetDirectoryPathWebOptions webOptions; // Empty for now
  final GetDirectoryPathIosOptions iosOptions; // Empty for now
  final GetDirectoryPathWindowsOptions windowsOptions; // Empty for now
  final GetDirectoryPathMacosOptions macosOptions; // Empty for now
  final GetDirectoryPathLinuxOptions linuxOptions; // Empty for now
}
```
*Pros:* Perfect strict typing. The IDE only suggests what actually works for that specific method on each platform, completely preventing user errors while remaining perfectly scalable.
*Cons:* Introduces a significant amount of boilerplate and class explosion, as we go from having one set of option classes to multiple sets (one for `pickFiles`, one for `getDirectoryPath`, etc.).

Which approach do you think fits the philosophy of the package better? This applies to all FilePicker methods: saveFile, etc...

navaronbracke:
I have the exact same concern. Mainly also because we still need to figure out what the redesigned plugin API will look like. (some things will change in that aspect)

I think we should put this on hold once we figured that out. My idea would be:
- keep the main API as idiomatic as possible: pickDirectory, pickFile, pickFiles, ...
- keep the main API as cross platform as possible, minimizing options that are used on say one specific platform (SAF options is one such thing, the window blur flag is another example), to avoid bloating the main API. For example the compressionQuality does not apply to directories etc
- put platform specific methods in the platform specific implementations (releaseGrant should only exist on FilePickerAndroid)

What we could do is ask for opinions in the package-authors channel on the Flutter Discord

vicajilau:
Hi @navaronbracke,

Regarding your feedback, I believe the best approach moving forward would be to go with **Option 3** as I described earlier. 

I lean towards this option mainly because of its strict typing. By using method-specific wrappers, the IDE will only suggest what actually works for that specific method on each platform. This completely prevents user errors while keeping the API perfectly scalable for future additions.

If you agree with this direction, I'm happy to start implementing it right away in this PR. I just want to get your "go" before jumping into the code.

vicajilau:
This is what it would look like if we implemented it. The upside? New platform-specific properties would scale without breaking the API. The downside? A bit of boilerplate code for classes that currently don't have any properties.

I don't mind it at first glance, what do you think @navaronbracke?

navaronbracke:
@vicajilau While I agree on the intent here, I think this will make the federated plugin route impossible?

Currently we have a barrel file that uses conditional imports to prevent issues on the web (WASM/JS vs IO).
The sole reason that works right now is because the toplevel `file_picker` API does not deviate from its API signature on all platforms.

However, there are signs that it should deviate for platform specific things:
- releasing Blobs in FilePickerWeb (this depends on Blob from package:web)
- releasing Android specific SAF handles in FilePickerAndroid (The SAFHandle class should also be moved to FilePickerAndroid)
- entitlements check skipping for FilePickerMacOS

All those things should not live in the generic FilePickerPlatform API, but rather in their platform specific implementations. That is my opinion on a future federated plugin system, where each platform has an implementation.

Example:

```
if (FilePickerPlatform.instance case final FilePickerAndroid implementation) {
  await implementation.releaseSafHandle(handle);
}
```

However, the options that you introduce here make that a little harder to achieve.

Thoughts?

vicajilau:
@navaronbracke I think we can agree that this approach is rather verbose, so we'll set it aside for now. I'll work on creating the federated plugin. I am closing this PR

PR #1992, opened 2026-04-19
Using GTK to Implement File Selection on Linux Platform

yz-java:
I originally intended to modify the Linux platform to support lockParentWindow based on dbus, but I found that obtaining the main window ID requires using gtk native.
The underlying Flutter Linux platform is based on GTK, and the GTK file selector supports synchronization, asynchrony, and UI customization. To implement these functions, simply call the GTK API.
DBUS is a great choice across platforms, but this advantage can be ignored on the Flutter Linux platform (based on GTK).
Next, I will implement asynchronous selection based on GTK API (lockParentWindow=false)


vicajilau (changes requested):
Thanks for the contribution!! The Linux improvements are valuable, especially the multi-directory selection support.

I’m requesting changes for one architectural concern before we can move forward: why should GTK become the primary implementation instead of keeping XDG Desktop Portal (D-Bus) as the default path?

From a compatibility and maintenance perspective, Portal is generally the safer default across desktop environments (GNOME/KDE), aligns better with sandboxed distributions (Flatpak/Snap), and follows the current Linux permission model more closely.

Could you please justify, with concrete cases, where Portal is insufficient and GTK is required as the primary path?

A hybrid strategy (Portal first, GTK fallback when needed) may be a better fit unless there is clear evidence that GTK-first improves behavior across the supported Linux matrix.

yz-java:
> Thanks for the contribution!! The Linux improvements are valuable, especially the multi-directory selection support.
> 
> I’m requesting changes for one architectural concern before we can move forward: why should GTK become the primary implementation instead of keeping XDG Desktop Portal (D-Bus) as the default path?
> 
> From a compatibility and maintenance perspective, Portal is generally the safer default across desktop environments (GNOME/KDE), aligns better with sandboxed distributions (Flatpak/Snap), and follows the current Linux permission model more closely.
> 
> Could you please justify, with concrete cases, where Portal is insufficient and GTK is required as the primary path?
> 
> A hybrid strategy (Portal first, GTK fallback when needed) may be a better fit unless there is clear evidence that GTK-first improves behavior across the supported Linux matrix.

I originally intended to modify the Linux platform to support lockParentWindow based on dbus, but I found that obtaining the main window ID requires using gtk native.
The underlying Flutter Linux platform is based on GTK, and the GTK file selector supports synchronization, asynchrony, and UI customization. To implement these functions, simply call the GTK API.
DBUS is a great choice across platforms, but this advantage can be ignored on the Flutter Linux platform (based on GTK).
Next, I will implement asynchronous selection based on GTK API (lockParentWindow=false)

PR #1993, opened 2026-04-23
Using GTK to Implement File Selection on Linux Platform

yz-java:
Support multi directory selection and lockParentWindow

### DIFF
--- a/example/lib/src/file_picker_demo.dart
+++ b/example/lib/src/file_picker_demo.dart
@@ -179,6 +179,52 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
     });
   }
 
+  void _pickDirectoryPaths() async {
+    List<String>? pickedDirectories;
+    bool hasUserAborted = true;
+    _resetState();
+
+    try {
+      pickedDirectories = await FilePicker.pickDirectoryPaths(
+        dialogTitle: _dialogTitleController.text,
+        initialDirectory: _initialDirectoryController.text,
+        allowMultiple: _multiPick,
+        lockParentWindow: _lockParentWindow,
+      );
+      hasUserAborted = pickedDirectories == null;
+    } on PlatformException catch (e) {
+      _logException('Unsupported operation: $e');
+    } catch (e) {
+      _logException(e.toString());
+    }
+    if (!mounted) return;
+
+    setState(() {
+      _isLoading = false;
+      _userAborted = hasUserAborted;
+      _resultsWidget = FilePickerResultsList(
+        itemCount: pickedDirectories?.length ?? 0,
+        itemBuilder: (BuildContext context, int index) {
+          String name = 'File path:';
+          if (!kIsWeb) {
+            final fs = LocalFileSystem();
+            name = fs.isFileSync(pickedDirectories![index])
+                ? 'File path:'
+                : 'Directory path:';
+          }
+          return ListTile(
+            leading: Text(
+              index.toString(),
+              style: TextStyle(fontSize: 15, fontWeight: FontWeight.bold),
+            ),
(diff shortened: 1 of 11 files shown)


navaronbracke:
@vicajilau this seems redundant?

vicajilau:
> @vicajilau this seems redundant?

Yeah, I don't think maintaining a dedicated GTK implementation makes sense now that we rely on **XDG Desktop Portals** (`org.freedesktop.portal.FileChooser` via DBus):

1. **Linux Freedesktop Standard**: XDG Desktop Portal is the official, universal standard across modern Linux distributions (Ubuntu, Fedora, Arch, etc.).
2. **Sandbox Friendly**: Direct GTK calls fail inside Flatpak and Snap sandboxes, whereas XDG Portals communicate over DBus and work seamlessly without requiring sandbox escape permissions.
3. **Native DE Integration**: XDG Portals automatically delegate to the user's active Desktop Environment (rendering native KDE/KIO dialogs on KDE Plasma and GNOME dialogs on GNOME/GTK desktops). Hardcoding GTK bypasses this and forces GTK UI on non-GTK desktops.
4. **Maintenance Overhead**: Since the portal already uses GTK under the hood on GTK-based environments, maintaining a separate native GTK C/C++ fallback adds unnecessary code complexity without providing any real functional benefit.

For these reasons, sticking to XDG Desktop Portal as the sole Linux implementation is cleaner, more robust, and future-proof.

yz-java:
> > @vicajilau this seems redundant?
> 
> Yeah, I don't think maintaining a dedicated GTK implementation makes sense now that we rely on **XDG Desktop Portals** (`org.freedesktop.portal.FileChooser` via DBus):
> 
> 1. **Linux Freedesktop Standard**: XDG Desktop Portal is the official, universal standard across modern Linux distributions (Ubuntu, Fedora, Arch, etc.).
> 2. **Sandbox Friendly**: Direct GTK calls fail inside Flatpak and Snap sandboxes, whereas XDG Portals communicate over DBus and work seamlessly without requiring sandbox escape permissions.
> 3. **Native DE Integration**: XDG Portals automatically delegate to the user's active Desktop Environment (rendering native KDE/KIO dialogs on KDE Plasma and GNOME dialogs on GNOME/GTK desktops). Hardcoding GTK bypasses this and forces GTK UI on non-GTK desktops.
> 4. **Maintenance Overhead**: Since the portal already uses GTK under the hood on GTK-based environments, maintaining a separate native GTK C/C++ fallback adds unnecessary code complexity without providing any real functional benefit.
> 
> For these reasons, sticking to XDG Desktop Portal as the sole Linux implementation is cleaner, more robust, and future-proof.

Does the Linux platform support lockParentWindow in the new version?

vicajilau:
Yes, it does. `lockParentWindow` maps to the XDG portal's `modal` option in all three entry points (`pickFiles`, `getDirectoryPath`, `saveFile`).

There is also `FilePickerLinuxOptions.parentWindow`, which matters here: the portal needs a window identifier (`x11:0x...` or `wayland:...`) to parent the dialog to. Without it, `modal` has no window to attach to.

Your question surfaced a bug, though. Passing a plain `LinuxOptions(lockParentWindow: true)` silently drops the flag, because the implementation narrows to `FilePickerLinuxOptions` and falls back to a default instance when the type does not match:

```
LinuxOptions(lockParentWindow: true)           -> modal = false
FilePickerLinuxOptions(lockParentWindow: true) -> modal = true
```

That is the opposite of what our own deprecation message tells people to use, since it points at `LinuxOptions.lockParentWindow`. Opening an issue and a fix for it. Thanks for asking the question that found it.

Issue #1932, opened 2025-12-22
☔ file_picker public API refactor


navaronbracke:
This issue is an umbrella issue, to list some of the things that we want to do for `file_picker` 12.0.0

- [x] https://github.com/miguelpruivo/flutter_file_picker/issues/1731 Fix the package compatibility issues with the supported platforms. https://github.com/miguelpruivo/flutter_file_picker/pull/1931
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1469 A redesign of the public Dart API, this still needs a design discussion, but is planned for this release.
- [x] Removing the deprecated `allowCompression` flag. This needs to be added to the changelog. https://github.com/miguelpruivo/flutter_file_picker/pull/1931
- [x] Rewriting the MacOS portion in Swift
- [x] Migrating iOS & MacOS to use `sharedDarwinSource` (this depends on the point above to be completed first)

Besides the above, I would also look into fixing the following bugs:

- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1825 This looks like we need to call `ContentResolver.takePersistableUriPermission(uri, flags)` when using `ACTION_OPEN_DOCUMENT`
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1911 This is a missing option. This depends on the design doc for the public API, as it can be discussed there.
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1904 To be looked into.
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1850 This looks like a case of forgetting https://developer.android.com/reference/android/content/Intent#resolveActivity(android.content.pm.PackageManager)
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1384 This can be looked into with the refactor of iOS / MacOS to a shared Darwin source
- [ ] https://github.com/miguelpruivo/flutter_file_picker/issues/1918 Looks like we should remove a specific line in the code

navaronbracke:
That is acceptable, but we should document it in the public API. IIRC Flutter will soon (or already does?) have a minimum of iOS 13

navaronbracke:
@vicajilau I've updated the release target to 12.0 for the big refactor.

daniJimen:
[...]

PR #2007, opened 2026-05-12
Feature/fix add loading into save (#1918)

### LINKED ISSUE #1918: [Bug]onFileLoading callback from pickFiles() incorrectly triggered by saveFile() in file_picker plugin on Android
The onFileLoading callback provided to FilePicker.platform.pickFiles() is unexpectedly triggered when saving files using FilePicker.platform.saveFile(), especially after a file has already been picked. This occurs when selecting any directory in Google Drive and saving a file.

**Platform**

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
What version did it happen?
NA

**How are you picking?**
NA

Provide all the details to reproduce the issue.

1. Call FilePicker.platform.pickFiles() with the onFileLoading callback to open the internal file picker.
2. After picking a file, call FilePicker.platform.saveFile() to open the internal file picker for saving a file.
3. Once the file is saved, the previously provided onFileLoading callback is triggered again.

Expected Behavior
The onFileLoading callback should only be triggered for the active file picking operation and not when saving files via saveFile().
Actual Behavior
The onFileLoading callback from the previous pickFiles() call is triggered when saving a file using saveFile().

**Error Log**
NA

**Screenshots and/or video**
NA

**Flutter Version details**
```
[✓] Flutter (Channel stable, 3.24.3, on macOS 14.7.8 23H730 darwin-arm64, locale en-US)
    • Flutter version 3.24.3 on channel stable at /Users/ks6281/Documents/GitHub/flutter
    • Framework revision 2663184aa7 (1 year, 1 month ago), 2024-09-11 16:27:48 -0500
    • Engine revision 36335019a8
    • Dart version 3.5.3
    • DevTools version 2.37.3

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
    • Android SDK at /Users/ks6281/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • ANDROID_HOME = /Users/ks6281/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.11+0-17.0.11b1207.24-11852314)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.11+0-17.0.11b1207.24-11852314)

[✓] VS Code (version 1.104.3)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.112.0

[✓] VS Code (version 1.94.2)
    • VS Code at /Users/ks6281/Downloads/Visual Studio Code.app/Contents
    • Flutter extension version 3.112.0

[✓] Connected device (4 available)
    • Kavya (mobile)                  • 00008130-001E74AE11FA001C • ios            • iOS 18.6.2 22G100
    • macOS (desktop)                 • macos                     • darwin-arm64   • macOS 14.7.8 23H730 darwin-arm64
    • Mac Designed for iPad (desktop) • mac-designed-for-ipad     • darwin         • macOS 14.7.8 23H730 darwin-arm64
    • Chrome (web)                    • chrome                    • web-javascript • Google Chrome 141.0.7390.66

[✓] Network resources

• No issues found!
```

**Additional context**
This issue only occurs if FilePicker.platform.pickFiles() is called before saveFile().().

daniJimen:
perf: offload file saving to a background isolate and update loading status

- Implement `_saveBytesIsolateEntry` to handle file writing off the main thread using isolates.
- Use `TransferableTypedData` for efficient cross-isolate communication when saving bytes.
- Move byte-saving logic from the native platform side to the Dart side in `MethodChannelFilePicker`.
- Trigger `onFileLoading` status updates during the file saving process.

fix: https://github.com/miguelpruivo/flutter_file_picker/issues/1918

### DIFF
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -119,17 +119,25 @@ class FilePickerUtils {
   }
 }
 
-Future<void> _saveBytesIsolateEntry(List<dynamic> args) async {
-  final SendPort send = args[0] as SendPort;
-  final String path = args[1] as String;
-  final TransferableTypedData transferable = args[2] as TransferableTypedData;
-  try {
-    final Uint8List bytes = transferable.materialize().asUint8List();
-    final file = File(path);
-    await file.writeAsBytes(bytes);
-    send.send({'result': 'ok'});
-  } catch (e) {
-    send.send({'error trying to save file': e});
+Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
+  // Decode expected message shape using pattern matching to avoid explicit
+  // casts. Expected: [SendPort send, String path, TransferableTypedData transferable]
+  if (args case [SendPort send, String path, TransferableTypedData transferable]) {
+    try {
+      final Uint8List bytes = transferable.materialize().asUint8List();
+      final file = File(path);
+      await file.writeAsBytes(bytes);
+      send.send(true);
+    } catch (e) {
+      send.send(e);
+    }
+    return;
+  }
+
+  if (args.isNotEmpty && args[0] is SendPort) {
+    final Object? first = args[0];
+    if (first is SendPort) {
+      first.send(Exception('Invalid isolate arguments'));
+    }
   }
 }
-
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -146,10 +146,7 @@ Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
     return;
   }
 
-  if (args.isNotEmpty && args[0] is SendPort) {
-    final Object? first = args[0];
-    if (first is SendPort) {
-      first.send(Exception('Invalid isolate arguments'));
-    }
+  if (args case [final SendPort port, ...]) {
+    port.send(Exception('Invalid isolate arguments'));
   }
 }
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -106,7 +106,9 @@ class FilePickerUtils {
 
       final result = await receivePort.first;
       receivePort.close();
-      if (result is Map && result['error trying to save file'] != null) {
+      if (result is Exception) {
+        throw result;
+      }
         throw result['error trying to save file'];
       }
     }
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -106,8 +106,8 @@ class FilePickerUtils {
 
       final result = await receivePort.first;
       receivePort.close();
-      if (result is Map && result['error trying to save file'] != null) {
-        throw result['error trying to save file'];
+      if (result is Exception) {
+        throw result;
       }
     }
   }
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -119,7 +119,6 @@ class FilePickerUtils {
   }
 }
 
-// Entry point for the spawned isolate. Args: [SendPort, String path, TransferableTypedData]
 Future<void> _saveBytesIsolateEntry(List<dynamic> args) async {
   final SendPort send = args[0] as SendPort;
   final String path = args[1] as String;
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -98,7 +98,11 @@ class FilePickerUtils {
       final receivePort = ReceivePort();
       final transferable = TransferableTypedData.fromList([bytes]);
 
-      await Isolate.spawn(_saveBytesIsolateEntry, [receivePort.sendPort, path, transferable]);
+      await Isolate.spawn(_saveBytesIsolateEntry, [
+        receivePort.sendPort,
+        path,
+        transferable,
+      ]);
 
       final result = await receivePort.first;
       receivePort.close();
@@ -122,7 +126,11 @@ class FilePickerUtils {
 Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
   // Decode expected message shape using pattern matching to avoid explicit
   // casts. Expected: [SendPort send, String path, TransferableTypedData transferable]
-  if (args case [SendPort send, String path, TransferableTypedData transferable]) {
+  if (args case [
+    SendPort send,
+    String path,
+    TransferableTypedData transferable,
+  ]) {
     try {
       final Uint8List bytes = transferable.materialize().asUint8List();
       final file = File(path);
--- a/lib/src/file_picker_utils.dart
+++ b/lib/src/file_picker_utils.dart
@@ -1,6 +1,8 @@
 import 'dart:io';
+import 'dart:isolate';
 import 'dart:typed_data';
 
+import 'package:flutter/foundation.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:path/path.dart';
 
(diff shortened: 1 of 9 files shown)


lib/src/file_picker_utils.dart
```diff
@@ -108,3 +118,18 @@ class FilePickerUtils {
         (codeUnit >= 97 && codeUnit <= 122); // a-z
   }
 }
+
+Future<void> _saveBytesIsolateEntry(List<dynamic> args) async {
```
navaronbracke: I have several questions with this approach:

1. Isolates are not supported on the web (you will need a conditional import I guess, with a no-op on the web?)
2. Can we return a nullable Exception through the send port? (which we could rethrow if non null) The current code with a Map seems a bit odd?
3. Can we use `List<Object?> args`, with an `if-case` expression to decode the map? Then we don't need the `as` checks

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -168,20 +169,52 @@ class MethodChannelFilePicker extends FilePickerPlatform {
...
+
+      await FilePickerUtils.saveBytesToFile(bytes, savedPath);
+
+      if (onFileLoading != null) {
+        onFileLoading(FilePickerStatus.done);
+      }
+
+      return savedPath;
+    } on PlatformException catch (e) {
+      print('[$_tag] Platform exception: $e');
```
navaronbracke: This print is redundant? The rethrow will surface the error anyway

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -168,20 +169,52 @@ class MethodChannelFilePicker extends FilePickerPlatform {
...
+      rethrow;
+    } catch (e) {
+      print(
```
navaronbracke: This seems odd to have a catch for? If the method is not found, we get a PlatformException back anyway (i.e. the result of `MethodNotImplemented`)

navaronbracke (changes requested):

lib/src/file_picker_utils.dart
daniJimen: 1- FilePickerUtils its never used in web platform, that,s the cause for its not needed to use any restriction in that case, the others points are solved now
2-resolved
3-resolved

lib/src/platform/file_picker_method_channel.dart
daniJimen: you are right, fixed!

lib/src/platform/file_picker_method_channel.dart
daniJimen: fixed! now we rethrow directly the error in catch

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -168,20 +163,46 @@ class MethodChannelFilePicker extends FilePickerPlatform {
...
 
-    return methodChannel.invokeMethod("save", {
-      "fileName": fileName,
-      "fileType": type.name,
-      "initialDirectory": initialDirectory,
-      "allowedExtensions": allowedExtensions,
-      "bytes": bytes,
-    });
+    try {
+      await _eventSubscription?.cancel();
```
navaronbracke: Should we not cancel the subscription at the very end of saveFile() as well? (in a finally block, which also sets it to null)

lib/src/file_picker_utils.dart
```diff
@@ -108,3 +122,30 @@ class FilePickerUtils {
         (codeUnit >= 97 && codeUnit <= 122); // a-z
   }
 }
+
+Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
```
navaronbracke, line 130: Make it a DartDoc comment? Then it shows up in the IDE

```suggestion
/// Save the given bytes to a file, using a separate [Isolate].
///
/// The [args] is expected to contain a [SendPort], the [String] file path
/// and the [TransferableTypedData] bytes, in this order.
Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
```

lib/src/file_picker_utils.dart
```diff
@@ -108,3 +122,30 @@ class FilePickerUtils {
...
+  if (args case [
+    SendPort send,
+    String path,
+    TransferableTypedData transferable,
+  ]) {
+    try {
+      final Uint8List bytes = transferable.materialize().asUint8List();
+      final file = File(path);
+      await file.writeAsBytes(bytes);
+      send.send(true);
```
navaronbracke:
```suggestion
      send.send(null);
```

lib/src/file_picker_utils.dart
```diff
@@ -108,3 +122,30 @@ class FilePickerUtils {
...
+    } catch (e) {
+      send.send(e);
+    }
+    return;
+  }
+
+  if (args.isNotEmpty && args[0] is SendPort) {
```
navaronbracke: This could also use an if case?

```
if (args case [final SendPort port, ...]) {
  port.send(Exception('Invalid isolate arguments'));
}
```

lib/src/file_picker_utils.dart
```diff
@@ -93,8 +95,20 @@ class FilePickerUtils {
...
+
+      await Isolate.spawn(_saveBytesIsolateEntry, [
+        receivePort.sendPort,
+        path,
+        transferable,
+      ]);
+
+      final result = await receivePort.first;
+      receivePort.close();
+      if (result is Map && result['error trying to save file'] != null) {
```
navaronbracke: This still needs to be updated? Also, Is the stack trace of the original exception preserved?

```suggestion
      if (result is Exception) {
        throw result;
      }
```

navaronbracke (changes requested):

lib/src/platform/file_picker_method_channel.dart
daniJimen: you are right, changed

lib/src/file_picker_utils.dart
daniJimen: done!

lib/src/file_picker_utils.dart
daniJimen: done!

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -149,14 +149,10 @@ class MethodChannelFilePicker extends FilePickerPlatform {
...
-    } on PlatformException catch (e) {
-      print('[$_tag] Platform exception: $e');
-      rethrow;
     } catch (e) {
-      print(
-        '[$_tag] Unsupported operation. Method not found. The exception thrown was: $e',
-      );
       rethrow;
+    } finally {
+      await _eventSubscription?.cancel();
```
navaronbracke, line 155:
```suggestion
      await _eventSubscription?.cancel();
      _eventSubscription = null;
```

navaronbracke (approved):
LGTM with one nit.

Also, could it be that we forgot changelog entries for your latest fixes in the PR's that got merged recently?

lib/src/file_picker_utils.dart
```diff
@@ -108,3 +122,31 @@ class FilePickerUtils {
         (codeUnit >= 97 && codeUnit <= 122); // a-z
   }
 }
+
+/// Save the given bytes to a file, using a separate [Isolate].
+///
+/// The [args] is expected to contain a [SendPort], the [String] file path
+/// and the [TransferableTypedData] bytes, in this order.
+Future<void> _saveBytesIsolateEntry(List<Object?> args) async {
+  // Decode expected message shape using pattern matching to avoid explicit
```
navaronbracke: This can be removed, as the doc comment above says the same thing

daniJimen: done!

daniJimen:
> Also, could it be that we forgot changelog entries for your latest fixes in the PR's that got merged recently?

sorry, now changelog its updated, i have in local the changelog changes

navaronbracke (approved):

navaronbracke:
@daniJimen Since 12.0.0-beta.3 was not yet published, I fixed up the changelog here, while also leaving out the non-user facing changes from it.

PR #2008, opened 2026-05-12
fix(android): update error code to explorer_not_found when file explorer is missing (#1850)

### LINKED ISSUE #1850: FilePicker.platform.pickFiles() hangs indefinitely when no compatible app is available
## Description:
When calling FilePicker.platform.pickFiles() on devices that don't have any compatible file manager applications installed, the method hangs indefinitely without returning a result or throwing an exception. This creates a poor user experience as the app becomes unresponsive and users have no feedback about what's happening.

### Expected Behavior
The method should either:

- Throw an exception that can be caught and handled appropriately
- Provide a way to check if file picking is available before attempting to open the picker

### Actual Behavior

- FilePicker.platform.pickFiles() never completes execution
- The Future never resolves (neither success nor error)
- No exception is thrown
- Debugger never reaches the if (result != null) clause

## Steps to Reproduce

- Use a device/emulator without any file manager apps installed
- Remove or disable system file managers (DocumentsUI, Files by Google, etc.)
- Call FilePicker.platform.pickFiles()
- Observe that the method never returns

daniJimen:
fix(android): update error code to explorer_not_found when file explorer is missing

fix: https://github.com/miguelpruivo/flutter_file_picker/issues/1850

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -252,7 +252,7 @@ object FileUtils {
                 FilePickerDelegate.TAG,
                 "Can't find a valid activity to handle the request. Make sure you've a file explorer installed."
             )
-            finishWithError("invalid_format_type", "Can't handle the provided file type.")
+            finishWithError("explorer_not_found", "Can't find a valid activity to handle the request. Make sure you've a file explorer installed.")
         }
     }
 
@@ -364,7 +364,7 @@ object FileUtils {
                 FilePickerDelegate.TAG,
                 "Can't find a valid activity to handle the request. Make sure you've a file explorer installed."
             )
-            finishWithError("invalid_format_type", "Can't handle the provided file type.")
+            finishWithError("explorer_not_found", "Can't find a valid activity to handle the request. Make sure you've a file explorer installed.")
         }
     }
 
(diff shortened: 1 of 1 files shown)


android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -252,7 +252,7 @@ object FileUtils {
                 FilePickerDelegate.TAG,
                 "Can't find a valid activity to handle the request. Make sure you've a file explorer installed."
             )
-            finishWithError("invalid_format_type", "Can't handle the provided file type.")
+            finishWithError("explorer_not_found", "Can't find a valid activity to handle the request. Make sure you've a file explorer installed.")
```
navaronbracke: "you have" instead of "you've" ?

navaronbracke: Is the error code misleading? I don't immediately see how changing the error code prevents the app from hanging in this case. Or was the bug fixed, but the issue not closed?

navaronbracke (approved):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
daniJimen: fixed

daniJimen: The error was fixed, but the error it was returning was incorrect and confusing, so I put a different error that was more descriptive of the real problem.

PR #2009, opened 2026-05-12
feat: add support for custom dialog titles in pickFileAndDirectoryPaths (#1911)

### LINKED ISSUE #1911: FilePicker.platform.pickFileAndDirectoryPaths doesn't have a dialogTitle option like FilePicker.platform.pickFiles

daniJimen:
- Add `dialogTitle` parameter to `pickFileAndDirectoryPaths` in the plugin interface and macOS implementation.
- Update macOS native handler to set the title and message on `NSOpenPanel` dialogs.
- Update the example application to allow providing a custom dialog title.

fix: https://github.com/miguelpruivo/flutter_file_picker/issues/1911

### DIFF
--- a/darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift
+++ b/darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift
@@ -61,6 +61,10 @@ final class MacOSFilePickerHandler {
            !initialDirectory.isEmpty {
             dialog.directoryURL = URL(fileURLWithPath: initialDirectory)
         }
+        if let title = args["dialogTitle"] as? String {
+            dialog.title = title
+            dialog.message = title
+        }
         dialog.showsHiddenFiles = false
         let allowMultiple = args["allowMultiple"] as? Bool ?? false
         dialog.allowsMultipleSelection = allowMultiple
@@ -109,12 +113,17 @@ final class MacOSFilePickerHandler {
             result(entitlementError)
             return
         }
+
         let dialog: NSOpenPanel = NSOpenPanel()
         let args = call.arguments as! [String: Any]
 
         if let initialDirectory = args["initialDirectory"] as? String,
            !initialDirectory.isEmpty {
             dialog.directoryURL = URL(fileURLWithPath: initialDirectory)
+        }
+        if let title = args["dialogTitle"] as? String {
+            dialog.title = title
+            dialog.message = title
         }
         dialog.showsHiddenFiles = false
         dialog.allowsMultipleSelection = true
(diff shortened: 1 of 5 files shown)


darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift
```diff
@@ -61,6 +61,10 @@ final class MacOSFilePickerHandler {
            !initialDirectory.isEmpty {
             dialog.directoryURL = URL(fileURLWithPath: initialDirectory)
         }
+        if let title = args["dialogTitle"] as? String {
+            dialog.title = title
+            dialog.message = title
```
navaronbracke, line 66: Should the message be set to the title? What does it look like then? (multiple times here)

example/lib/src/file_picker_demo.dart
```diff
@@ -138,6 +138,7 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
 
     try {
       pickedFilesAndDirectories = await FilePicker.pickFileAndDirectoryPaths(
+        dialogTitle: _dialogTitleController.text,
```
navaronbracke, line 150: I noticed we do not clean up state in the example (we have initState() but no dispose() for TextEditingControllers etc). Can you fix this too, just in case?

navaronbracke (changes requested):

darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift
daniJimen: That's the point of the issue; ultimately, the user was complaining that the variable indicating the title wasn't being included in pickFileAndDirectoryPaths, but upon including it, I realized that nothing was being done with this title, so what I did was set it as the title of the native window.

example/lib/src/file_picker_demo.dart
daniJimen: done!

darwin/file_picker/Sources/file_picker/MacOSFilePickerHandler.swift
navaronbracke: Sorry, I was talking about `dialog.message = title`. Where is `dialog.message` in the dialog, in relation to the title?

navaronbracke: Ah, found out what it is for.

navaronbracke (approved):

Issue #2013, opened 2026-05-15
12-beta.3 File type any does not list sub directories


xShuusui:
Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened inssues. Also, check if it isn't described on the [Wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki), specially on [Troubleshooting](https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting) page.

Also, sometimes a simple `flutter clean` and `flutter build` again with latest file_picker version, may end up by fixing cached issues, so I encourage you to first do so.

**Describe the bug**
When picking with `FileType.any` on the current beta, the file explorer is opening with its root path and when navigating to our specific directory structure ("Documents/XXX/XXX/.zip") there are no directories visible under ("Documents/XXX").

When using `FileType.custom` without any extensions (required as zip can not be translated to MimeTypes) it works like in v11 and we can go through our directory structure till we see all ".zip" files. 

**Platform**

- [x ] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
12.0.0-beta.3

**How are you picking?**
```
await FilePicker.pickFiles(type: FileType.any);
```

**Details to reproduce the issue**
Provide all the details to reproduce the issue.

**Error Log**
Please, post the _full_ console log of your issue, if applicable.

**Screenshots and/or video**
If applicable, add screenshots or video to help explain your problem.

**Flutter Version details**
```
[✓] Flutter (Channel stable, 3.41.9, on Fedora Linux 44 (KDE Plasma Desktop Edition) 7.0.4-200.fc44.x86_64, locale
    en_US.UTF-8) [88ms]
    • Flutter version 3.41.9 on channel stable at ~/.puro/envs/3.41.9/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 00b0c91f06 (2 weeks ago), 2026-04-29 10:03:19 -0700
    • Engine revision 42d3d75a56
    • Dart version 3.11.5
    • DevTools version 2.54.2
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop,
      enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file,
      enable-lldb-debugging, enable-uiscene-migration

[!] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [575ms]
[...]

PR #2015, opened 2026-05-15
fix(android): restore subdirectory visibility for FileType.any

xShuusui:
Should solve #2013 and #2012.

Hii @daniJimen I don't want to do anything I have no deep understanding from. And this PR is just some fast vibe coding without thoughts behind it. I think you edited this part a few days ago, so maybe you can have a look at it. After this change it was possible for me to navigate the directories and pick our .ini files again. Otherwise `any` behaves differently than `custom without extensions`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 12.0.0-beta.4
+
+### Android
+- Fixed `FileType.any` not showing subdirectories in the document picker due to `EXTRA_MIME_TYPES` being passed as a bare `String` instead of a `String[]`.
+
 ## 12.0.0-beta.3
 ### General
 - Added `onFileLoading` callback to `saveFile` and implemented status tracking via an event channel. `saveFile` now reports loading status (for example `FilePickerStatus.loading` and `FilePickerStatus.done`).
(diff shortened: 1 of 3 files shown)


vicajilau:
Hi! Thank you very much for your contribution and for identifying the issue so quickly.

While your proposal effectively works around the bug by changing the input in `FilePickerPlugin`, we've decided to go with a fix directly in `FileUtils.kt`. 

The root cause is that `Intent.EXTRA_MIME_TYPES` strictly requires a `String[]` array according to Android documentation. By fixing it inside the `startFileExplorer` method, we ensure that the `Intent` is built correctly for any future use case or file type, making the implementation more robust and resilient to empty extension lists.

Closing this now as I have created https://github.com/miguelpruivo/flutter_file_picker/pull/2016 as the fix for this. Thanks again for the help!

xShuusui:
> Hi! Thank you very much for your contribution and for identifying the issue so quickly.
> 
> While your proposal effectively works around the bug by changing the input in `FilePickerPlugin`, we've decided to go with a fix directly in `FileUtils.kt`.
> 
> The root cause is that `Intent.EXTRA_MIME_TYPES` strictly requires a `String[]` array according to Android documentation. By fixing it inside the `startFileExplorer` method, we ensure that the `Intent` is built correctly for any future use case or file type, making the implementation more robust and resilient to empty extension lists.
> 
> Closing this now as I have created #2016 as the fix for this. Thanks again for the help!

Thanks for the fast Feedback! Tested your fix and it works for us when using type any :partying_face: . Looking forward to the version 12 release with the improvements of the API.

PR #2017, opened 2026-05-16
feat: implement pickFile() API and improve PlatformFile documentation (#1469)

vicajilau:
This PR implements the suggestions from issue #1469 to improve the Developer Experience (DX). 
## Changes: 
- Added FilePicker.pickFile() convenience method for single selections. 
- Improved PlatformFile documentation regarding nullability and memory usage. 
- Updated example app and README.md.

### DIFF
--- a/example/lib/src/file_picker_demo.dart
+++ b/example/lib/src/file_picker_demo.dart
@@ -85,52 +85,26 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         final result = await FilePicker.pickFiles(
           type: _pickingType,
           allowMultiple: true,
-          onFileLoading: (FilePickerStatus status) => setState(() {
-            _isLoading = status == FilePickerStatus.picking;
-          }),
-          allowedExtensions: (_extension?.isNotEmpty ?? false)
-              ? _extension?.replaceAll(' ', '').split(',')
-              : null,
+          onFileLoading: _onFileLoading,
+          allowedExtensions: _allowedExtensionsFromInput(),
           dialogTitle: _dialogTitleController.text,
           initialDirectory: _initialDirectoryController.text,
           lockParentWindow: _lockParentWindow,
           withData: _withData,
-          androidSafOptions: (_safPersist || _safReadWrite)
-              ? AndroidSAFOptions(
-                  grant: _safPersist
-                      ? AndroidSAFGrant.lifetime
-                      : AndroidSAFGrant.transient,
-                  accessMode: _safReadWrite
-                      ? AndroidSAFAccessMode.readWrite
-                      : AndroidSAFAccessMode.readOnly,
-                )
-              : null,
+          androidSafOptions: _androidSafOptionsFromFlags(),
         );
         printInDebug("pickedFiles: $result");
         pickedFiles = result?.files;
       } else {
         final file = await FilePicker.pickFile(
           type: _pickingType,
-          onFileLoading: (FilePickerStatus status) => setState(() {
-            _isLoading = status == FilePickerStatus.picking;
-          }),
-          allowedExtensions: (_extension?.isNotEmpty ?? false)
-              ? _extension?.replaceAll(' ', '').split(',')
-              : null,
+          onFileLoading: _onFileLoading,
+          allowedExtensions: _allowedExtensionsFromInput(),
           dialogTitle: _dialogTitleController.text,
           initialDirectory: _initialDirectoryController.text,
           lockParentWindow: _lockParentWindow,
           withData: _withData,
-          androidSafOptions: (_safPersist || _safReadWrite)
-              ? AndroidSAFOptions(
-                  grant: _safPersist
-                      ? AndroidSAFGrant.lifetime
-                      : AndroidSAFGrant.transient,
-                  accessMode: _safReadWrite
-                      ? AndroidSAFAccessMode.readWrite
-                      : AndroidSAFAccessMode.readOnly,
-                )
-              : null,
+          androidSafOptions: _androidSafOptionsFromFlags(),
         );
         printInDebug("pickedFile: $file");
         pickedFiles = file != null ? [file] : null;
@@ -177,9 +151,7 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
       pickedFilesAndDirectories = await FilePicker.pickFileAndDirectoryPaths(
         dialogTitle: _dialogTitleController.text,
         type: _pickingType,
-        allowedExtensions: (_extension?.isNotEmpty ?? false)
-            ? _extension?.replaceAll(' ', '').split(',')
-            : null,
+        allowedExtensions: _allowedExtensionsFromInput(),
         initialDirectory: _initialDirectoryController.text,
       );
       hasUserAborted = pickedFilesAndDirectories == null;
@@ -252,16 +224,7 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         dialogTitle: _dialogTitleController.text,
         initialDirectory: _initialDirectoryController.text,
         lockParentWindow: _lockParentWindow,
-        androidSafOptions: (_safPersist || _safReadWrite)
-            ? AndroidSAFOptions(
-                grant: _safPersist
-                    ? AndroidSAFGrant.lifetime
-                    : AndroidSAFGrant.transient,
-                accessMode: _safReadWrite
-                    ? AndroidSAFAccessMode.readWrite
-                    : AndroidSAFAccessMode.readOnly,
-              )
-            : null,
+        androidSafOptions: _androidSafOptionsFromFlags(),
       );
       hasUserAborted = pickedDirectoryPath == null;
     } on PlatformException catch (e) {
@@ -301,9 +264,7 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
 
     try {
       pickedSaveFilePath = await FilePicker.saveFile(
-        allowedExtensions: (_extension?.isNotEmpty ?? false)
-            ? _extension?.replaceAll(' ', '').split(',')
-            : null,
+        allowedExtensions: _allowedExtensionsFromInput(),
         type: FileType.custom,
         dialogTitle: _dialogTitleController.text,
         fileName: _defaultFileNameController.text,
@@ -351,6 +312,31 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
       _isLoading = true;
       _userAborted = true;
     });
+  }
+
+  void _onFileLoading(FilePickerStatus status) {
+    setState(() {
+      _isLoading = status == FilePickerStatus.picking;
+    });
+  }
+
+  List<String>? _allowedExtensionsFromInput() {
+    return (_extension?.isNotEmpty ?? false)
+        ? _extension?.replaceAll(' ', '').split(',')
+        : null;
+  }
+
+  AndroidSAFOptions? _androidSafOptionsFromFlags() {
+    return (_safPersist || _safReadWrite)
+        ? AndroidSAFOptions(
+            grant: _safPersist
+                ? AndroidSAFGrant.lifetime
+                : AndroidSAFGrant.transient,
+            accessMode: _safReadWrite
+                ? AndroidSAFAccessMode.readWrite
+                : AndroidSAFAccessMode.readOnly,
+          )
+        : null;
   }
 
   @override
--- a/example/lib/src/file_picker_demo.dart
+++ b/example/lib/src/file_picker_demo.dart
@@ -261,15 +261,17 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
     String? pickedSaveFilePath;
     bool hasUserAborted = true;
 
-    final bytes = pickedFiles?.firstOrNull?.bytes;
+    final file = pickedFiles?.firstOrNull;
     final fileName = _defaultFileNameController.text;
 
-    if (bytes == null || fileName.isEmpty) {
+    if (file == null || fileName.isEmpty) {
       _logException(
-        'Please pick a file with data enabled first and provide a default file name.',
+        'Please pick a file first and provide a default file name.',
       );
       return;
     }
+
+    final bytes = await file.readAsBytes();
 
     _resetState();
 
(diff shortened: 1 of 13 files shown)


example/lib/src/file_picker_demo.dart
```diff
@@ -81,32 +81,60 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-              )
-            : null,
-      );
-      printInDebug("pickedFiles: $result");
-      pickedFiles = result?.files;
+      if (_multiPick) {
+        final result = await FilePicker.pickFiles(
+          type: _pickingType,
+          allowMultiple: true,
+          onFileLoading: (FilePickerStatus status) => setState(() {
```
navaronbracke: Some things are duplicated here:
- `onFileLoading: (FilePickerStatus status) => ...`
- allowed extensions computation
- android saf options computation

lib/src/file_picker.dart
```diff
@@ -87,6 +88,46 @@ abstract final class FilePicker {
...
+    AndroidSAFOptions? androidSafOptions,
+  }) async {
+    final result = await pickFiles(
+      dialogTitle: dialogTitle,
+      initialDirectory: initialDirectory,
+      type: type,
+      allowedExtensions: allowedExtensions,
+      onFileLoading: onFileLoading,
+      compressionQuality: compressionQuality,
+      allowMultiple: false,
```
navaronbracke, line 133: Can we deprecate `allowMultiple` on `pickFiles()` ?

Now that we have pickFile (single file), pickFiles implies being able to select many files, instead of just one.

lib/src/file_picker.dart
navaronbracke, line 78: One thing I would also try to deprecate with version 12.0.0 is `withData` and `withReadStream` + `readSequential `.

I think that you should be able to call `readAsBytes()` / `readAsByteStream()` on PlatformFile directly, so it is up to the user to decide when to load in the data, but we provide the means to do it.

We can do that in a new PR thogh.

navaronbracke, line 266: While I am looking at the public API a little, I think the last improvement that we can do is the following:

Currently I find it confusing with how `saveFile()` takes a nullable bytes and nullable file name.

Isn't the point of this method on all platforms that you provide a filename + bytes and we will save that file for you? Also I think we can make the documentation for the method easier to read?

```
  /// For mobile, this function will save a file with the given [fileName] and [bytes] and return the path where the file was saved.
  ///
  /// For desktop platforms, this function opens a dialog to let the user choose a location for the file and returns the selected path.
  /// If the bytes are provided, then the bytes are written to a file at the chosen path.
```

is saying twice the same?

navaronbracke: "selected a file" is a bit weird, the user chose the path where a file should be saved, no file was selected?

navaronbracke: `If the [bytes] or [fileName] are omitted, this will throw an [ArgumentError].` can be removed if we make these two argments non null?

navaronbracke: What happens if fileName is the empty string? Isn't the filename required to be non empty?

navaronbracke:
```suggestion
  /// On macOS if the [initialDirectory] is invalid, the user directory or previously valid directory
```

navaronbracke (changes requested):
I have some suggestions on cleaning up the public API, but overall this is LGTM.

Feel free to do most of the other cleanups in new PR's though.

example/lib/src/file_picker_demo.dart
vicajilau: Good point, fixed!

lib/src/file_picker.dart
vicajilau: Another good point, fixed!

lib/src/file_picker.dart
vicajilau: Agreed!! Done in this PR

vicajilau: True!! Fixed

vicajilau: Fixed

vicajilau: Fixed!

vicajilau: I've fixed it manually, I think this commit is intended to modify the existing line rather than replace the blank line 😄

vicajilau: Fixed now

vicajilau:
I really liked your ideas!! Great job @navaronbracke!!

vicajilau:
What do you think of this pull request @navaronbracke? I’ve followed your suggestions and it might be ready for version v12

lib/src/api/platform_file.dart
```diff
@@ -40,6 +40,11 @@ class PlatformFile {
   /// ```
   /// final File myFile = File(platformFile.path);
   /// ```
+  ///
+  /// This property is `null` in the following cases:
```
navaronbracke: The explanation for web is duplicated? (see lines 48-49)

lib/src/api/platform_file.dart
```diff
@@ -49,10 +54,19 @@ class PlatformFile {
 
   /// Byte data for this file. Particularly useful if you want to manipulate its data
   /// or easily upload to somewhere else.
+  ///
+  /// This property is `null` unless [FilePicker.pickFiles] (or `pickFile`) was called with
+  /// `withData: true`. Enabling this on mobile platforms for large files may lead to
+  /// Out Of Memory (OOM) issues; use [readStream] instead.
+  ///
   /// [Check here in the FAQ](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ) an example on how to use it to upload on web.
   final Uint8List? bytes;
```
navaronbracke, line 64: Per my previous comment: Can we add `@Deprecated()` to this, in favor of the new `readAsBytes()`?

We now deprecated `withData`, so it is a bit weird that this is also not deprecated now.

lib/src/api/platform_file.dart
```diff
@@ -49,10 +54,19 @@ class PlatformFile {
...
 
-  /// File content as stream
+  /// File content as a stream of bytes.
+  ///
+  /// This property is `null` unless [FilePicker.pickFiles] (or `pickFile`) was called with
+  /// `withReadStream: true`. This is the recommended way to handle large files on
+  /// mobile and desktop platforms.
   final Stream<List<int>>? readStream;
```
navaronbracke, line 72: Same remark about `@Deprecated()` for this field

lib/src/api/platform_file.dart
```diff
@@ -79,6 +97,12 @@ class PlatformFile {
     }
   }
 
+  /// Read the file content as bytes.
+  Future<Uint8List> readAsBytes() => xFile.readAsBytes();
+
+  /// Read the file content as a stream of bytes.
```
navaronbracke, line 108: Should we mention that this is preferred for large files?

lib/src/api/platform_file.dart
```diff
@@ -79,6 +97,12 @@ class PlatformFile {
...
+  Stream<Uint8List> readAsByteStream() => xFile.openRead();
```
navaronbracke, line 116: Should we also add a `length()` function?

lib/src/file_picker.dart
```diff
@@ -55,17 +56,35 @@ abstract final class FilePicker {
   /// Note: This requires the User Selected File Read entitlement on macOS.
   ///
   /// Returns `null` if aborted.
+  /// NOTE: `allowMultiple` is deprecated. Use `pickFile` for single-file
```
navaronbracke: These notes are redundant? `@Deprecated` is sufficient

lib/src/file_picker.dart
```diff
@@ -87,6 +106,55 @@ abstract final class FilePicker {
...
+    String? dialogTitle,
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+    Function(FilePickerStatus)? onFileLoading,
+    int compressionQuality = 0,
+    @Deprecated(
+      'Use PlatformFile.readAsBytes(); this parameter will be removed in a future release',
+    )
+    bool withData = kIsWeb,
```
navaronbracke: Since this is a new API, can we _not_ add immediately deprecated parameters? :)

Just remove them here. These will be removed in v13.0 anyway

navaronbracke (changes requested):
Overall this is LGTM, but I have some questions around the deprecations.

lib/src/api/platform_file.dart
vicajilau: **Line 48:** Provides specific technical information about the web—it explains that `path` points to a Blob URL and how to revoke it using `URL.revokeObjectURL()`. This is concrete, actionable technical content.

**Line 49:** This is simply a generic reference link (“Read more about it here”) that points to the FAQ for further details.

They are not the same: one is specific technical documentation, the other is just a link. However, it might seem redundant if you prefer. If you wish, we can:
- Remove line 49 (the link) if the technical explanation in line 48 is sufficient
- Or combine them into a single, more concise line
- Or keep them as they are if you feel the link adds value

Which do you prefer @navaronbracke?

lib/src/api/platform_file.dart
vicajilau: Done!!

lib/src/api/platform_file.dart
vicajilau: Done!!

lib/src/api/platform_file.dart
navaronbracke: I would combine both? Also is the "On Android, when picking a directory" bit relevant? This should not allow you to pick directories?

```
This property is `null` on Android, when using SAF without caching enabled.

On the web this may or may not point to a Blob URL, which can be cleaned up using [URL.revokeObjectURL](https://pub.dev/documentation/web/latest/web/URL/revokeObjectURL.html).
```

lib/src/api/platform_file.dart
```diff
@@ -49,10 +54,23 @@ class PlatformFile {
...
   /// Byte data for this file. Particularly useful if you want to manipulate its data
   /// or easily upload to somewhere else.
+  ///
+  /// This property is `null` unless [FilePicker.pickFiles] (or `pickFile`) was called with
+  /// `withData: true`. Enabling this on mobile platforms for large files may lead to
+  /// Out Of Memory (OOM) issues; use [readStream] instead.
+  ///
   /// [Check here in the FAQ](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ) an example on how to use it to upload on web.
+  @Deprecated(
+    'Use readAsBytes() instead to avoid OOM issues on mobile platforms',
```
navaronbracke:
```suggestion
    'Use readAsBytes() instead to avoid out-of-memory issues with large files.',
```

lib/src/api/platform_file.dart
```diff
@@ -79,6 +101,22 @@ class PlatformFile {
...
+  ///
+  /// For large files on mobile and desktop platforms, prefer [readAsByteStream]
+  /// to avoid Out Of Memory (OOM) issues.
+  Future<Uint8List> readAsBytes() => xFile.readAsBytes();
+
+  /// Read the file content as a stream of bytes.
+  ///
+  /// This is the recommended way to handle large files on mobile and desktop platforms.
+  ///
+  /// **Note on Web:** This method currently requires [FilePicker.pickFiles] or [FilePicker.pickFile]
```
navaronbracke, line 112: Could we solve this in the future by leveraging the blob capabilities of XFile?

lib/src/file_picker.dart
navaronbracke: These notes are still present?

navaronbracke (changes requested):

lib/src/api/platform_file.dart
vicajilau: Noted!!

lib/src/file_picker.dart
vicajilau: Removed!

lib/src/api/platform_file.dart
vicajilau: Done!!

lib/src/file_picker.dart
vicajilau: Makes sense!! Done!!

lib/src/api/platform_file.dart
vicajilau: Yes, I think we could solve this in the future by leveraging the Blob capabilities of XFile on web.

Right now, `readAsByteStream()` depends on `bytes`, which means it still relies on `withData: true` indirectly. Since `withData` is deprecated, that creates a mismatch: the API encourages streaming, but the web implementation still needs the deprecated path.

A better long-term approach would be to preserve the Blob URL as the file source on web and let `XFile` stream directly from that Blob, or use JS/Web APIs such as `Blob.stream()` or `FileReader` as a fallback. That would remove the dependency on `withData` and make `readAsByteStream()` consistent across platforms.

lib/src/file_picker.dart
vicajilau: Fixed!!

vicajilau:
Everything resolved @navaronbracke. Better now? 🚀

lib/src/api/platform_file.dart
```diff
@@ -73,22 +93,61 @@ class PlatformFile {
...
+  /// to have been called with `withData: true`, otherwise this will fail since `bytes` is required
+  /// to create the stream. This is a limitation that should be addressed in a future version to
+  /// support proper streaming on web without requiring `withData`.
+  Stream<Uint8List> readAsByteStream() => xFile.openRead();
+
+  /// Returns the length of the file in bytes.
+  ///
+  /// Resolution order:
```
navaronbracke: This documentation is an implementation detail?

I would remove it and only leave the "Note:" below

lib/src/api/platform_file.dart
```diff
@@ -73,22 +93,61 @@ class PlatformFile {
...
+    }
+  }
+
   @override
   bool operator ==(Object other) {
     if (identical(this, other)) {
       return true;
     }
 
+    // ignore: deprecated_member_use_from_same_package
```
navaronbracke: Put the deprecated_member_use_from_same_package as disabled in the analysis options?

file_picker is allowed to use its own deprecated members

navaronbracke (approved):
LGTM with two nits

lib/src/api/platform_file.dart
vicajilau: Fixed!!

lib/src/api/platform_file.dart
vicajilau: Fixed!

PR #2018, opened 2026-05-18
Feature/refactor code prevent freece UI

daniJimen:
Perform Android file saving on `Dispatchers.IO` using Coroutines to prevent UI blocking.
- Use Dart's `compute` function to read and write file bytes in background isolates.
- Delegate byte writing to the native layer in `saveFile` instead of performing it in Dart.
- Update `pickFiles` to read file bytes asynchronously when `withData` is enabled.
- Improve `PlatformFile.toString()` output to display byte length instead of full content.
- Ensure the native event subscription is properly cleared in the `finally` block.
- Add `kotlinx-coroutines-android` dependency to the Android project

resolves issue: https://github.com/miguelpruivo/flutter_file_picker/issues/2019

### DIFF
--- a/darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
+++ b/darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
@@ -1,0 +1,139 @@
+import Foundation
+import UniformTypeIdentifiers
+
+enum FilePickerUtils {
+    static func buildSaveFileName(from arguments: [String: Any], data: Data?, lastPickedFileExtension: String?) -> String {
+        let providedName = ((arguments["fileName"] as? String) ?? "").trimmingCharacters(
+            in: .whitespacesAndNewlines)
+        let fileType = (arguments["fileType"] as? String)?.lowercased()
+        let allowedExtensions = arguments["allowedExtensions"] as? [String] ?? []
+
+        let fallbackBaseName = "file_picker_file"
+        let baseName = providedName.isEmpty ? fallbackBaseName : providedName
+
+        // Keep user extension when provided; only infer when it is missing.
+        if !URL(fileURLWithPath: baseName).pathExtension.isEmpty {
+            return baseName
+        }
+
+        if let inferredExtension = inferSaveExtension(
+            fileType: fileType,
+            allowedExtensions: allowedExtensions,
+            lastPickedExtension: lastPickedFileExtension,
+            data: data)
+        {
+            return "\(baseName).\(inferredExtension)"
+        }
+
+        return baseName
+    }
+
+    static func inferSaveExtension(
+        fileType: String?,
+        allowedExtensions: [String],
+        lastPickedExtension: String?,
+        data: Data?
+    ) -> String? {
+        if let lastPickedExtension,
+           !lastPickedExtension.isEmpty
+        {
+            return lastPickedExtension
+        }
+
+        if let customExtension = allowedExtensions
+            .map({ $0.trimmingCharacters(in: .whitespacesAndNewlines) })
+            .map({ $0.hasPrefix(".") ? String($0.dropFirst()) : $0 })
+            .first(where: { !$0.isEmpty })
+        {
+            return customExtension
+        }
+
+        if let inferredByBytes = inferExtensionFromDataSignature(data) {
+            return inferredByBytes
+        }
+
+        switch fileType {
+        case "image":
+            return "jpg"
+        case "video":
+            return "mp4"
+        case "audio":
+            return "m4a"
+        default:
+            return nil
+        }
+    }
+
+    static func inferExtensionFromDataSignature(_ data: Data?) -> String? {
+        guard let data, !data.isEmpty else {
+            return nil
+        }
+
+        func hasPrefix(_ bytes: [UInt8]) -> Bool {
+            guard data.count >= bytes.count else {
+                return false
+            }
+            return data.prefix(bytes.count).elementsEqual(bytes)
+        }
+
+        func asciiString(from range: Range<Int>) -> String? {
+            guard data.count >= range.upperBound else {
+                return nil
+            }
+            let subdata = data.subdata(in: range)
+            return String(data: subdata, encoding: .ascii)
+        }
+
+        if hasPrefix([0x25, 0x50, 0x44, 0x46]) { // %PDF
+            return "pdf"
+        }
+        if hasPrefix([0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A]) {
+            return "png"
+        }
+        if hasPrefix([0xFF, 0xD8, 0xFF]) {
+            return "jpg"
+        }
+        if hasPrefix([0x47, 0x49, 0x46, 0x38]) {
+            return "gif"
+        }
+        if hasPrefix([0x49, 0x49, 0x2A, 0x00]) || hasPrefix([0x4D, 0x4D, 0x00, 0x2A]) {
+            return "tiff"
+        }
+        if hasPrefix([0x42, 0x4D]) {
+            return "bmp"
+        }
+        if hasPrefix([0x52, 0x49, 0x46, 0x46]),
+           let riffType = asciiString(from: 8..<12)
+        {
+            if riffType == "WEBP" {
+                return "webp"
+            }
+            if riffType == "WAVE" {
+                return "wav"
+            }
+        }
+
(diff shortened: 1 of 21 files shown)


vicajilau (changes requested):
Could you add an entry to the changelog? On the other hand, I'm worried that this change might affect the other platforms. Do you need help testing it on any of the platforms?

daniJimen:
> Could you add an entry to the changelog? On the other hand, I'm worried that this change might affect the other platforms. Do you need help testing it on any of the platforms?

now, you have updated the changelog! about test in another platforms, its working fine my changes in all platforms, i only was detected one issue, but i could check actually is in master:
https://github.com/miguelpruivo/flutter_file_picker/issues/2020

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
```diff
@@ -145,6 +146,7 @@ final class IOSFilePickerHandler: NSObject,
                 else {
                     return
                 }
+                self.updateLastPickedFileExtension(from: copiedURL)
```
navaronbracke, line 149: Why do we need the last picked extension?

lib/src/file_picker_utils.dart
```diff
@@ -1,11 +1,18 @@
...
 import 'package:flutter/foundation.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:path/path.dart';
 
+/// Reads the file at [path] and returns its bytes.  Runs in a worker isolate.
+Future<Uint8List> _readBytesFromPath(String path) => File(path).readAsBytes();
+
+/// Writes [args[1]] (Uint8List) to the path [args[0]] (String).
+Future<void> _writeBytesToPath(List<Object> args) async {
+  final path = args[0] as String;
```
navaronbracke: Use an if-case, instead of casts?

lib/src/platform/file_picker_method_channel.dart
```diff
@@ -132,22 +132,33 @@ class MethodChannelFilePicker extends FilePickerPlatform {
...
         return null;
       }
 
-      final List<PlatformFile> platformFiles = <PlatformFile>[];
+      await _eventSubscription?.cancel();
+      _eventSubscription = null;
 
+      final List<PlatformFile> platformFiles = [];
       for (final Map platformFileMap in result) {
+        final String? path = platformFileMap['path'] as String?;
```
navaronbracke: Use an if case to get path/bytes in the for loop?

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
```diff
@@ -279,6 +282,147 @@ final class IOSFilePickerHandler: NSObject,
         topViewController()?.present(picker, animated: true)
     }
 
+    private func buildSaveFileName(from arguments: [String: Any], data: Data?) -> String {
```
navaronbracke: Should we move these new functions to a utils class?

navaronbracke (changes requested):

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
daniJimen: Because in iOS we actually save the files in their raw format, and with my changes, I save the extension of the original file to assign it to the new file when it is saved.

lib/src/file_picker_utils.dart
daniJimen: done!

lib/src/platform/file_picker_method_channel.dart
daniJimen: done!

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
daniJimen: done!

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
```diff
@@ -0,0 +1,139 @@
...
+        lastPickedExtension: String?,
+        data: Data?
+    ) -> String? {
+        if let lastPickedExtension,
+           !lastPickedExtension.isEmpty
+        {
+            return lastPickedExtension
+        }
+
+        if let customExtension = allowedExtensions
```
navaronbracke, line 43: What happens if allowedExtensions is out of order? I.e. `.pdf, .docx`, but I pick a `.docx` without extension for some reason? Does this then infer `.pdf`?

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
```diff
@@ -0,0 +1,139 @@
...
+            .first(where: { !$0.isEmpty })
+        {
+            return customExtension
+        }
+
+        if let inferredByBytes = inferExtensionFromDataSignature(data) {
+            return inferredByBytes
+        }
+
+        switch fileType {
```
navaronbracke, line 55: Is this switch case accurate enough?

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
```diff
@@ -279,6 +282,16 @@ final class IOSFilePickerHandler: NSObject,
         topViewController()?.present(picker, animated: true)
     }
 
+    // buildSaveFileName, inferSaveExtension and inferExtensionFromDataSignature
```
navaronbracke, line 285: This comment is a bit redundant in code

vicajilau:
To me, and with all due respect, these changes don't make sense, since, as we discussed [here]( https://github.com/miguelpruivo/flutter_file_picker/pull/2017), `withData` is deprecated and will be phased out in favor of `readAsBytes`. Are you agree @navaronbracke and @daniJimen?

navaronbracke:
1) Use Dart's compute function to read and write file bytes in background isolates
-> can we move the read file on Isolate behavior to the new readAsBytes() method?

2) Delegate byte writing to the native layer in saveFile instead of performing it in Dart.
-> Why is this needed?

3) Update pickFiles to read file bytes asynchronously when withData is enabled.
-> I suppose this is fine, but I do agree, withData is deprecated, so I don't think this will be needed for long? (we will remove withData in the next major release)

4) Improve PlatformFile.toString() output to display byte length instead of full content.
-> same remakr with withData being deprecated, but this is fine as a stopgap fix

5) Ensure the native event subscription is properly cleared in the finally block
-> I did mention this before, so thanks for addressing this

Add kotlinx-coroutines-android dependency to the Android project
-> The coroutine fix is something we should keep

vicajilau:
> 1. Use Dart's compute function to read and write file bytes in background isolates
>    -> can we move the read file on Isolate behavior to the new readAsBytes() method?

`readAsBytes()` is already asynchronous, but moving it to an isolate is not a universal solution. For large files it still requires loading the whole file into memory, and passing large byte arrays across isolates adds overhead. It also does not fit web or URI-based platforms like Android SAF. For that reason, I’d keep `readAsBytes()` as the API and rely on `readAsByteStream()` for large files.

> 2\. Delegate byte writing to the native layer in saveFile instead of performing it in Dart.
>     -> Why is this needed?

Because `saveFile` is not just “write these bytes to a local path”. The platform layer is already responsible for choosing the destination, and on some platforms that destination is not a plain filesystem path at all.

The main reasons are:

- On Android, the save target is often a `content://` URI from the Storage Access Framework, so Dart cannot reliably write to it as a normal file.
- On iOS, saving/exporting is driven by `UIDocumentPickerViewController`, and the platform owns the export flow.
- On web, there is no writable path, only a download/blob flow.
- Keeping the write in native gives one consistent implementation per platform, instead of forcing Dart to assume a local path exists everywhere.

So the short answer is: it is needed to make `saveFile` truly cross-platform and to support save destinations that are URI-based or sandboxed, not just filesystem paths.

> 3\. Update pickFiles to read file bytes asynchronously when withData is enabled.
>     -> I suppose this is fine, but I do agree, withData is deprecated, so I don't think this will be needed for long? (we will remove withData in the next major release)

Agree.

> 4\. Improve PlatformFile.toString() output to display byte length instead of full content.
>     -> same remakr with withData being deprecated, but this is fine as a stopgap fix

Agree.

> 5\. Ensure the native event subscription is properly cleared in the finally block
>     -> I did mention this before, so thanks for addressing this

Agree.

> Add kotlinx-coroutines-android dependency to the Android project
> -> The coroutine fix is something we should keep

Agree.

PR #2026, opened 2026-05-20
fix(android): remove explicit KGP dependency to resolve warnings under AGP 9.0+

vicajilau:
This PR removes the explicit application and dependencies of the Kotlin Gradle Plugin (KGP) in the plugin module to resolve warnings about legacy KGP usage under AGP 9.0+.

Additionally, it adds a dynamic configuration block to align the Kotlin compiler's JVM target bytecode with Java (JVM 17) to prevent target compatibility mismatch errors when compiling.

It also adds `android.builtInKotlin=false` to the example app's gradle.properties to allow compatibility with other third-party dependencies (like `flutter_plugin_android_lifecycle`) that still require external KGP.

Resolves this warning:
```sh
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): file_picker
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.

Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing 
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors

If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors
```

I have tested this using AGP 8 and AGP 9.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.5.1'
+        classpath 'com.android.tools.build:gradle:8.5.2'
     }
 }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,9 +2,6 @@ group 'com.mr.flutter.plugin.filepicker'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext {
-        kotlin_version = '1.8.22'
-    }
     repositories {
         google()
         mavenCentral()
@@ -12,7 +9,6 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:8.5.1'
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -50,11 +38,15 @@ android {
         targetCompatibility JavaVersion.VERSION_17
     }
 
-    if (shouldApplyKotlinAndroidPlugin) {
-        kotlinOptions {
-            jvmTarget = JavaVersion.VERSION_17.toString()
+    tasks.configureEach { task ->
+        if (task.class.name.contains("KotlinCompile")) {
+            task.kotlinOptions {
+                jvmTarget = "17"
```
navaronbracke: Should this not use `JavaVersion.VERSION_17.toString()` ?

android/build.gradle
```diff
@@ -50,11 +38,15 @@ android {
...
+            }
         }
     }
+
```
navaronbracke, line 48: Remove these blank lines?

navaronbracke (approved):
I suppose some of this is part of the migration from https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#supporting-flutter-versions-earlier-than-3-44 ?

android/build.gradle
vicajilau: Updated!

android/build.gradle
vicajilau: Done!

vicajilau:
> I suppose some of this is part of the migration from https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#supporting-flutter-versions-earlier-than-3-44 ?

Yes!! I based my work on that

PR #2035, opened 2026-05-22
feat: move web save file and byte handling logic

daniJimen:
* Prepare savefile from web to receive the bytes of blob file picked
* Update FilePicker.saveFile and the underlying platform interface to require the bytes parameter and remove the optional path parameter.
* Enhance PlatformFile.readAsBytes() on Web to automatically fetch content from blob: and data: URLs if bytes are not already cached.
* Implement fetchBytesFromWebPath using js_interop for Web, with a corresponding stub for non-web platforms to support conditional importing.
* Improve error messaging in PlatformFile when file data is missing on Web, providing guidance on using withData: true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## NEXT
+
+### General
+- Restored `pickFile()` default data loading on web so single-file picks include bytes by default.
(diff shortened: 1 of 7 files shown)


vicajilau:
It's very interesting, although it's missing some features, such as loading the Blob without reading the file. I'm closing this PR to open another one based on it.

PR #2044, opened 2026-05-26
[IOS][MACOS] - Feature: save file with selected file extension

### LINKED ISSUE #1740: [macOS] Hide/Unselect file extension in Save-as dialog
When saving a file on macOS, the Save-as dialog shows and automatically selects the entire filename with the extension included. This is undesirable because users rename the file name before saving and unknowingly delete the extension at the end.

Some other macOS apps automatically select the filename without the extension in the Save-As dialog so it looks like it's possible. Can you please add a setting to enable this behavior?

daniJimen:
* save files with correct extension
fixed: https://github.com/miguelpruivo/flutter_file_picker/issues/1740


lib/src/file_picker_utils.dart
```diff
@@ -90,25 +102,24 @@ class FilePickerUtils {
...
-        path,
-        transferable,
-      ]);
-
-      final result = await receivePort.first;
-      receivePort.close();
-      if (result is Exception) {
-        throw result;
-      }
+      await compute(_writeBytesToPath, [path, bytes]);
```
vicajilau: I like the replacement of `Isolate.spawn` with `compute`

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
```diff
@@ -0,0 +1,107 @@
+import Foundation
+import UniformTypeIdentifiers
+#if canImport(CoreServices)
+import CoreServices
```
vicajilau: Can you split this into two separate files for iOS and macOS?

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
```diff
@@ -0,0 +1,107 @@
...
+        let tempURL = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
+        defer { try? FileManager.default.removeItem(at: tempURL) }
+
+        do {
+            try data.prefix(1024).write(to: tempURL)
+        } catch {
+            return nil
+        }
+
+        if #available(macOS 11.0, iOS 14.0, *) {
```
vicajilau: Does that “if available” make sense? Could it be prevented?

vicajilau (changes requested):

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
daniJimen: solved

darwin/file_picker/Sources/file_picker/FilePickerUtils.swift
daniJimen: solved

vicajilau:
For me happens the same here @navaronbracke, this is completely outdated!!

PR #2045, opened 2026-05-26
[Android] - Feature: add coroutine to save file process

daniJimen:
* Added a corutine process in android to prevent freeze UI

### DIFF
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerDelegate.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerDelegate.kt
@@ -12,6 +12,9 @@ import io.flutter.plugin.common.EventChannel.EventSink
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.PluginRegistry.ActivityResultListener
 import java.io.IOException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 
 class FilePickerDelegate(
     val activity: Activity,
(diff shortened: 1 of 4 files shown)


lib/src/platform/file_picker_method_channel.dart
```diff
@@ -189,6 +190,10 @@ class MethodChannelFilePicker extends FilePickerPlatform {
             "bytes": bytes,
           });
 
+      if (!Platform.isAndroid) {
```
vicajilau: This code duplicates the file-writing logic in both native code and Dart.

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerDelegate.kt
```diff
@@ -71,21 +74,26 @@ class FilePickerDelegate(
...
+                    finishWithSuccess(renamedUri.path)
+                }
+            } catch (e: IOException) {
+                Log.e(TAG, "Error while saving file", e)
+                Handler(Looper.getMainLooper()).post {
+                    finishWithError("Error while saving file", e.message)
+                }
+            }
         }
+        return true
```
vicajilau, line 96: Does it make sense to return `true` if you don't know how the write operation went until the coroutine completes?

vicajilau (changes requested):

lib/src/platform/file_picker_method_channel.dart
daniJimen: removed

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FilePickerDelegate.kt
daniJimen: the boolean are not evaluated in any moment, its only needed becasue you are inside of an activityResult, and it needs return a boolean

vicajilau (approved):

vicajilau:
LGTM

PR #2047, opened 2026-05-27
[All platforms][IOS] - feature: Refactor code in example & ios image order

### LINKED ISSUE #1866: Not preserving file names or selection order on iOS
**Describe the bug**

On iOS only, the FilePicker.platform.pickFiles function with {type: FileType.video} returns the selected files out of order, and without the source file names. For my app, I either need the selection order or the correct file names, as that will allow the app to sequence the videos in the correct order. The file names, last modified date, etc all seem to be generated when the files are being cached and the file names seem to be random. The last version that I could find that preserves the source file name is 8.0.7, so I know that it's possible.

**Platform**

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Desktop

**Platform OS version**
17.5.1

**How are you picking?**
```
FilePicker.platform.pickFiles(
    type: FileType.video,
    allowMultiple: true,
    compressionQuality: 0
 )
```

Tested with all of the latest versions, including 10.3.0. Select video files with the above configuration, and then log the file paths to the console. You will notice a random order with random file names every time.

**Flutter Version details**
[✓] Flutter (Channel stable, 3.32.6, on macOS 15.6 24G84 darwin-arm64, locale en-US) [459ms]
    • Flutter version 3.32.6 on channel stable at /Users/jonathankerbelis/Development/flutter
    • Framework revision 077b4a4ce1 (5 weeks ago), 2025-07-08 13:31:08 -0700
    • Engine revision 72f2b18bb0
    • Dart version 3.8.1
    • DevTools version 2.45.1

daniJimen:
Disable save file in multi-pick mode and maintain selection order on iOS

fixed: https://github.com/miguelpruivo/flutter_file_picker/issues/1866

### DIFF
--- a/darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
+++ b/darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
@@ -132,9 +132,10 @@ final class IOSFilePickerHandler: NSObject,
 
         eventSink?(true)
         let group = DispatchGroup()
-        var resolved: [[String: Any]] = []
+        var resolved = Array<[String: Any]?>(repeating: nil, count: results.count)
+        let resolvedLock = NSLock()
 
-        for item in results {
+        for (index, item) in results.enumerated() {
             group.enter()
             item.itemProvider.loadFileRepresentation(
                 forTypeIdentifier: UTType.item.identifier
@@ -146,7 +147,9 @@ final class IOSFilePickerHandler: NSObject,
                     return
                 }
                 if let fileInfo = self.makeFileInfo(from: copiedURL) {
-                    resolved.append(fileInfo)
+                    resolvedLock.lock()
+                    resolved[index] = fileInfo
+                    resolvedLock.unlock()
                 }
             }
         }
@@ -156,7 +159,8 @@ final class IOSFilePickerHandler: NSObject,
                 return
             }
             eventSink?(false)
-            currentResult(resolved.isEmpty ? nil : resolved)
+            let orderedResolved = resolved.compactMap { $0 }
+            currentResult(orderedResolved.isEmpty ? nil : orderedResolved)
             result = nil
         }
     }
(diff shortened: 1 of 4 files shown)


daniJimen:
@vicajilau @navaronbracke The order feature is available only on devices running iOS 15 or later. On devices with earlier iOS versions (fewer than 1% of iOS devices), the user's selection order cannot be preserved.

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
```diff
@@ -215,6 +219,9 @@ final class IOSFilePickerHandler: NSObject,
     private func presentMediaPicker(type: String, allowsMultipleSelection: Bool) {
         var configuration = PHPickerConfiguration(photoLibrary: .shared())
         configuration.selectionLimit = allowsMultipleSelection ? 0 : 1
+        if #available(iOS 15.0, *) {
```
vicajilau, line 222: Since the list of supported devices is the same in [iOS 14](https://support.apple.com/en-by/guide/iphone/iphe3fa5df43/14.0/ios/14.0) as in [iOS 15](https://support.apple.com/en-by/guide/iphone/iphe3fa5df43/15.0/ios/15.0), I don't see any problem with upgrading support to iOS 15 to avoid using `if #available` structures in the code. Are you agree @navaronbracke?

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
```diff
@@ -146,7 +147,9 @@ final class IOSFilePickerHandler: NSObject,
                     return
                 }
                 if let fileInfo = self.makeFileInfo(from: copiedURL) {
-                    resolved.append(fileInfo)
+                    resolvedLock.lock()
```
navaronbracke, line 150: This is needed because `loadFileRepresentation()` can take a whle?

navaronbracke (approved):

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
navaronbracke: I would keep the `#available` ? That keeps support for iOS 14 and I think we can drop support for iOS 14 when Flutter does that itself as well.

darwin/file_picker/Sources/file_picker/IOSFilePickerHandler.swift
daniJimen: The NSLock around resolved[index] = fileInfo is required to prevent data races when multiple loadFileRepresentation callbacks run concurrently

PR #2052, opened 2026-05-29
Release 12.0.0

vicajilau:
Promote file_picker to 12.0.0 and add a stable changelog entry above the beta series.


navaronbracke:
@vicajilau Getting the federated plugin refactor in, is that more work than expected to safely include in 12.0.0?

vicajilau:
This will be done after federate migration

PR #2057, opened 2026-06-01
Fix readAsBytes fallback for path-based files

vicajilau:
This PR makes `PlatformFile.readAsBytes()` work even when a file was picked without `withData` and only has a local `path` or a read stream.

Impact:
- Desktop callers can now safely use `readAsBytes()` after picking files without preloading bytes.
- Keeps the web blob/data URL fallback intact.
- Adds a regression test to cover the path-only case.
- Updates the changelog and version to `12.0.0-beta.6`.

Before this change there was an issue on Desktop platform using `readAsBytes()`:

```sh
Bad state: PlatformFile.readAsBytes(): file data is not available. Consume the file via PlatformFile.readAsByteStream(), or on Web ensure the file path is a fetchable blob/data URL that can be retrieved.
```

After this change:

https://github.com/user-attachments/assets/c49d9a44-0449-4873-b826-6c7712114f47

### DIFF
--- a/lib/src/api/platform_file.dart
+++ b/lib/src/api/platform_file.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:typed_data';
 
 import 'package:cross_file/cross_file.dart';
 import 'package:flutter/foundation.dart';
@@ -110,11 +111,11 @@ class PlatformFile {
     if (bytes != null) return bytes!;
 
     if (readStream != null) {
-      final buffer = <int>[];
+      final builder = BytesBuilder();
       await for (final chunk in readStream!) {
-        buffer.addAll(chunk);
+        builder.add(chunk);
       }
-      return Uint8List.fromList(buffer);
+      return builder.takeBytes();
     }
 
     if (kIsWeb) {
--- a/lib/src/api/platform_file.dart
+++ b/lib/src/api/platform_file.dart
@@ -109,9 +109,19 @@ class PlatformFile {
   Future<Uint8List> readAsBytes() async {
     if (bytes != null) return bytes!;
 
+    if (readStream != null) {
+      final buffer = <int>[];
+      await for (final chunk in readStream!) {
+        buffer.addAll(chunk);
+      }
+      return Uint8List.fromList(buffer);
+    }
+
     if (kIsWeb) {
       final fetchedBytes = await fetchBytesFromWebPath(path);
       if (fetchedBytes != null) return fetchedBytes;
+    } else if (path != null) {
+      return xFile.readAsBytes();
     }
 
     throw StateError(
--- a/lib/src/platform/web/file_picker_web.dart
+++ b/lib/src/platform/web/file_picker_web.dart
@@ -7,6 +7,7 @@ import 'package:file_picker/src/api/file_picker_result.dart';
 import 'package:file_picker/src/api/platform_file.dart';
 import 'package:file_picker/src/api/android_saf_options.dart';
 import 'package:file_picker/src/platform/file_picker_platform_interface.dart';
+import 'package:file_picker/src/file_picker_utils.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:path/path.dart' as p;
 import 'package:web/web.dart';
@@ -65,11 +66,7 @@ class FilePickerWeb extends FilePickerPlatform {
     int compressionQuality = 0,
     AndroidSAFOptions? androidSafOptions,
   }) async {
-    if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
-      throw Exception(
-        'You are setting a type [$type]. Custom extension filters are only allowed with FileType.custom, please change it or remove filters.',
-      );
-    }
+    FilePickerUtils.validateAllowedExtensions(type, allowedExtensions);
 
     Completer<List<PlatformFile>?>? filesCompleter =
         Completer<List<PlatformFile>?>();
--- a/lib/src/platform/web/file_picker_web.dart
+++ b/lib/src/platform/web/file_picker_web.dart
@@ -7,7 +7,7 @@ import 'package:file_picker/src/api/file_picker_result.dart';
 import 'package:file_picker/src/api/platform_file.dart';
 import 'package:file_picker/src/api/android_saf_options.dart';
 import 'package:file_picker/src/platform/file_picker_platform_interface.dart';
-import 'package:file_picker/src/file_picker_utils.dart';
+import 'package:file_picker/src/utils/file_picker_utils.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:path/path.dart' as p;
 import 'package:web/web.dart';
--- a/lib/src/utils/_file_utils_io.dart
+++ b/lib/src/utils/_file_utils_io.dart
@@ -6,10 +6,10 @@ import 'package:file_picker/file_picker.dart';
 import 'package:path/path.dart';
 
 Future<List<PlatformFile>> filePathsToPlatformFiles(
-  List<String> filePaths,
-  bool withReadStream,
-  bool withData,
-) {
+  List<String> filePaths, {
+  bool withReadStream = false,
+  bool withData = false,
+}) {
   return Future.wait(
     filePaths.where((String filePath) => filePath.isNotEmpty).map((
       String filePath,
--- a/lib/src/utils/_file_utils_io.dart
+++ b/lib/src/utils/_file_utils_io.dart
@@ -69,21 +69,23 @@ Future<String> isExecutableOnPath(String executable) async {
 }
 
 Future<void> saveBytesToFile(Uint8List? bytes, String? path) async {
-  if (path != null && bytes != null && bytes.isNotEmpty) {
-    final receivePort = ReceivePort();
-    final transferable = TransferableTypedData.fromList([bytes]);
+  if (path == null || bytes == null || bytes.isEmpty) {
+    return;
+  }
 
-    await Isolate.spawn(_saveBytesIsolateEntry, [
-      receivePort.sendPort,
-      path,
-      transferable,
-    ]);
+  final receivePort = ReceivePort();
+  final transferable = TransferableTypedData.fromList([bytes]);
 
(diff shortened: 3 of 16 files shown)


lib/src/api/platform_file.dart
```diff
@@ -109,9 +109,19 @@ class PlatformFile {
   Future<Uint8List> readAsBytes() async {
     if (bytes != null) return bytes!;
 
+    if (readStream != null) {
+      final buffer = <int>[];
```
navaronbracke: Use BytesBuilder? `<int>[]` for byte buffers is just asking for trouble. Always prefer `Uint8List` (which implements List<int>), but here the BytesBuilder is even better.

lib/src/platform/web/file_picker_web.dart
```diff
@@ -65,11 +66,7 @@ class FilePickerWeb extends FilePickerPlatform {
     int compressionQuality = 0,
     AndroidSAFOptions? androidSafOptions,
   }) async {
-    if (type != FileType.custom && (allowedExtensions?.isNotEmpty ?? false)) {
-      throw Exception(
-        'You are setting a type [$type]. Custom extension filters are only allowed with FileType.custom, please change it or remove filters.',
-      );
-    }
+    FilePickerUtils.validateAllowedExtensions(type, allowedExtensions);
```
navaronbracke, line 69: `FilePickerUtils` import dart:io and `dart:isolate`, so this will break.

I think we should have the following:
- `_file_utils_web.dart`
- `_file_utils_io.dart`
- conditional import in FilePickerUtils, for the native only things

navaronbracke (changes requested):

lib/src/api/platform_file.dart
vicajilau: Good point!! Fixed

lib/src/platform/web/file_picker_web.dart
vicajilau: Fixed!

vicajilau:
Everything resolved @navaronbracke 🚀

lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      final bytes = await file.readAsBytes();
+      return createPlatformFile(file, bytes, null);
+    }).toList(),
+  );
+}
+
+Future<PlatformFile> createPlatformFile(
+  dynamic file,
```
navaronbracke: Can this be `Object?`

lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,107 @@
...
+  Uint8List? bytes,
+  Stream<List<int>>? readStream,
+) async {
+  final f = file as File;
```
navaronbracke: Use an if-case?

```
if (file case final File nativeFile) {
  // ...
}

throw ArgumentError('Expected file to be a File.');
```

lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,107 @@
+import 'dart:io';
+import 'dart:isolate';
+import 'dart:typed_data';
+
+import 'package:file_picker/file_picker.dart';
+import 'package:path/path.dart';
+
+Future<List<PlatformFile>> filePathsToPlatformFiles(
+  List<String> filePaths,
+  bool withReadStream,
```
navaronbracke: make these named bool arguments?

lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,107 @@
...
+Future<String> isExecutableOnPath(String executable) async {
+  final path = await runExecutableWithArguments('which', [executable]);
+  if (path == null) {
+    throw Exception('Couldn\'t find the executable $executable in the path.');
+  }
+  return path;
+}
+
+Future<void> saveBytesToFile(Uint8List? bytes, String? path) async {
+  if (path != null && bytes != null && bytes.isNotEmpty) {
```
navaronbracke: Use an early return?

lib/src/utils/_file_utils_web.dart
```diff
@@ -0,0 +1,34 @@
...
+Future<List<PlatformFile>> filePathsToPlatformFiles(
+  List<String> filePaths,
+  bool withReadStream,
+  bool withData,
+) => throw UnsupportedError(
+  'filePathsToPlatformFiles is only supported on native platforms',
+);
+
+Future<PlatformFile> createPlatformFile(
+  dynamic file,
```
navaronbracke:
```suggestion
  Object? file,
```

lib/src/utils/_file_utils_web.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:typed_data';
+import 'package:file_picker/file_picker.dart';
+
+Future<List<PlatformFile>> filePathsToPlatformFiles(
+  List<String> filePaths,
+  bool withReadStream,
```
navaronbracke: required named arguments, to match the other implementation

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+/// Utility class for [FilePicker] that provides common helper methods
+/// used across different platform implementations.
+class FilePickerUtils {
+  /// The default title for the file picker dialog.
+  static const String defaultDialogTitle = 'File Picker';
+
+  /// Converts a list of file paths into a list of [PlatformFile]s.
+  ///
+  /// This method is useful when the platform file picker returns a list of paths
+  /// and we need to convert them into the plugin's internal representation.
```
navaronbracke: Avoid "we", also I think the docstring can just be "Converts a list of file paths into a list of [PlatformFile]s."

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+  ) => impl.filePathsToPlatformFiles(filePaths, withReadStream, withData);
+
+  /// Creates a [PlatformFile] instance from a [File] object.
+  ///
+  /// [file] is the source file.
+  /// [bytes] are the file bytes (optional).
+  /// [readStream] is a stream of the file content (optional).
+  @visibleForTesting
+  static Future<PlatformFile> createPlatformFile(
+    dynamic file,
```
navaronbracke:
```suggestion
    Object? file,
```

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+  ///
+  /// This method is useful when the platform file picker returns a list of paths
+  /// and we need to convert them into the plugin's internal representation.
+  ///
+  /// [filePaths] is the list of absolute paths to the selected files.
+  /// [withReadStream] if true, the [PlatformFile] will contain a read stream.
+  /// [withData] if true, the [PlatformFile] will contain the file bytes (use carefully with large files).
+  static Future<List<PlatformFile>> filePathsToPlatformFiles(
+    List<String> filePaths,
+    bool withReadStream,
```
navaronbracke: required named arguments for the bools

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+    bool withData,
+  ) => impl.filePathsToPlatformFiles(filePaths, withReadStream, withData);
+
+  /// Creates a [PlatformFile] instance from a [File] object.
+  ///
+  /// [file] is the source file.
+  /// [bytes] are the file bytes (optional).
```
navaronbracke: "(optional)" is inferred from the nullability, so can be removed from the docstring?

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+  static Future<PlatformFile> createPlatformFile(
+    dynamic file,
+    Uint8List? bytes,
+    Stream<List<int>>? readStream,
+  ) => impl.createPlatformFile(file, bytes, readStream);
+
+  /// Runs an executable with the given arguments and returns the output.
+  ///
+  /// Returns the trimmed stdout as a [String], or null if the process fails
+  /// (exit code != 0) or produces no output.
```
navaronbracke: "(exit code != 0)" is a bit redundant?

lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,99 @@
...
+    if (x.isEmpty) return false;
+    final int codeUnit = x.codeUnitAt(0);
+    return (codeUnit >= 65 && codeUnit <= 90) || // A-Z
+        (codeUnit >= 97 && codeUnit <= 122); // a-z
+  }
+
+  /// Validates the `allowedExtensions` parameter against the provided [type].
+  ///
+  /// Throws an [ArgumentError] if extension filters are provided while the
+  /// `type` is not `FileType.custom`. Uses a consistent error message across
```
navaronbracke: "Uses a consistent error message across platforms" is a bit redundant, since it is an internal helper function?

navaronbracke (changes requested):
Some minor nits still

lib/src/utils/_file_utils_io.dart
vicajilau: True!!

lib/src/utils/_file_utils_io.dart
vicajilau: Fixed!

lib/src/utils/_file_utils_io.dart
vicajilau: YESS!! Fixed

lib/src/utils/file_picker_utils.dart
vicajilau: Done!

lib/src/utils/_file_utils_io.dart
vicajilau: Simplier... I love it!

lib/src/utils/file_picker_utils.dart
vicajilau: True

vicajilau:
All ready @navaronbracke !!

navaronbracke (approved):

navaronbracke:
Fixed up some small typos in the utils file. Once the tests are green this can be merged

PR #2059, opened 2026-06-02
[GENERAL] Feature/implement writing by path

### LINKED ISSUE #1903: `allowedExtensions` does not recognize `jsonl` files
**Describe the bug**
`allowedExtensions` doesn't work if the input is `jsonl` while it works fine for `json`

**Platform**

- [x] Android
- [x] iOS
- [ ] Web        (not tested)
- [ ] Desktop (not tested)

**Platform OS version**
- iPhone 16 Pro Max with iOS 18.1
- Pixel 9 with Android 15

**How are you picking?**
```dart
    final fileContent = await _fileSystemRepository.pickFile(
      extensions: ['jsonl'], // FilePicker doesn't seems to support jsonl files
    );



  Future<File> pickFile({List<String>? extensions}) async {
      allowedExtensions: extensions,
      type: extensions != null ? FileType.custom : FileType.any,
    );
    if (result != null) return File(result.files.single.path!);

    throw FileStorageException('File not selected');
  }
```

**Screenshots and/or video**

Image

**Flutter Version details**
flutter `3.29.3` & file_picker `10.3.3`

**Additional context**
`saveFile` function, the extension is specified in the filename, it works fine with `json` but not `jsonl`
```dart
  Future<void> saveFile(String content, String filename) async {
    final bytes = utf8.encode(content);
    final result = await FilePicker.platform.saveFile(
      bytes: bytes,
      fileName: filename,
    );

    if (result == null) throw FileStorageException('File not saved');
  }
```

### LINKED ISSUE #2036: Allow to save files without using byte data
Currently the `FilePicker.saveFile(..)` API requires passing the `bytes` parameter containing fill file data to save. For large files (200..400mb video files aren't uncommon) it puts a lot of pressure to device memory.

It would be great if the `.saveFile(..)` offered alternative ways to provide the file data: stream, a file path (the temp path may already be used by the platform code) or maybe XFile. Also see `share_plus` and `file_saver` plugins.

### LINKED ISSUE #2037: Allow to select files with a "persistent" link/path without creating a temporary copy
When working with large files, especially media files from user's gallery it is not desirable to make copies for large media files.

It would be really great to have an option for selecting files that would to return a "persistent" link/path that could use used at any time later to access selected file content.

### On MacOS, Windows, Linux

Should be okay to return the full file path and have the original file:// url in the `PlatformFile.identifier`
Though 

### On iOS

If a user picks a file outside app's sandbox, the OS grants a temporary security token. To access it later without copying it, the native file URL need to be converted to into a Security-Scoped Bookmark:

* https://www.avanderlee.com/swift/security-scoped-bookmarks-for-url-access/
* https://developer.apple.com/documentation/professional-video-applications/using-bookmark-data
* https://pub.dev/packages/ns_file_coordinator_util
* https://pub.dev/packages/fast_file_picker

1. Create a Bookmark (Save Access). Convert selected file URL.

```
func createSecurityScopedBookmark(from fileURL: URL) -> String? {
    // MUST explicitly declare intent to access the file resource
    guard fileURL.startAccessingSecurityScopedResource() else {
        print("Failed to start accessing security-scoped resource.")
        return nil
    }
    
    // Ensure resource access is released when the function completes
    defer {
        fileURL.stopAccessingSecurityScopedResource()
    }
    
    do {
        // Generate security-scoped data bytes
        let bookmarkData = try fileURL.bookmarkData(
            options: .withSecurityScope, // Required for sandboxed apps
            includingResourceValuesForKeys: nil,
            relativeTo: nil
        )
        
        // convert to a Base64 string for storing it
        return bookmarkData.base64EncodedString()
        
    } catch {
        print("Error creating bookmark: \(error.localizedDescription)")
        return nil
    }
}
```

2. Resolve a Bookmark (Regain Access Later). Call this on a future app launch using the base64-encoded bookmark string.

```
func resolveSecurityScopedBookmark(from base64String: String) -> URL? {
    guard let bookmarkData = Data(base64Encoded: base64String) else {
        print("Invalid base64 string.")
        return nil
    }
    
    do {
        var isStale = false
        
        // Resolve data back into a valid security-scoped URL
        let resolvedURL = try URL(
            resolvingBookmarkData: bookmarkData,
            options: .withSecurityScope,
            relativeTo: nil,
            bookmarkDataIsStale: &isStale
        )
        
        if isStale {
            print("Bookmark data is stale. You may need to ask the user to pick the file again.")
        }
        
        return resolvedURL
        
    } catch {
        print("Error resolving bookmark: \(error.localizedDescription)")
        return nil
    }
}

// How to read the file contents using the resolved URL
func printFileContents(at url: URL) {
    if url.startAccessingSecurityScopedResource() {
        defer { url.stopAccessingSecurityScopedResource() }
        
        // Do your file reading operations safely here
        if let data = try? Data(contentsOf: url) {
            print("Successfully read \(data.count) bytes directly from the original file source!")
        }
    } else {
        print("Permission denied to access the resolved URL.")
    }
}
```

### On Android

You cannot use raw file paths (/storage/emulated/0/...). Instead, can use the native storage access framework to capture the content:// URI and request long-term persistence permissions. See `shared_storage` or more recently updated `saf_util` and `saf_stream` packages for Android.

The `PlatformFile.identifier` property already containing the content://  URI. Though there is no API to read data or create a `PlatformFile` for a previously saved URI.

Also it seem like there is no option to not create a temp file when there is `content://` resource selected. Perhaps it could be added to `AndroidSAFOptions`.

### LINKED ISSUE #2055: Android compilation bug all versions - not serious
exception: c:\Users\...\AppData\Local\Pub\Cache\hosted\pub.dev\file_picker-12.0.0-beta.5\android\src\main\kotlin\com\mr\flutter\plugin\filepicker\FilePickerPlugin.kt:167:70: warning: unchecked cast of 'Any?' to 'ArrayList<String>?'.
exception:                     getMimeTypes(arguments?.get("allowedExtensions") as ArrayList<String>?)

copied from PowerShell

the same bug was from mart - when I started using it.
Despite this bug, after compiling alwais worked good.

### LINKED ISSUE #2056: `FilePickerResult? result = await FilePicker.pickFiles();` freezes the app for a really short time on macOS.
`FilePickerResult? result = await FilePicker.pickFiles();` freezes the app for a really short time on macOS.

This can be clearly observed if a loading spinner is spinning on the app or on the animation of the button you just clicked.

This makes the app feel less native than it could.

daniJimen:
* Implement write data when save file by path or bytes, and prevent to send bytes to native platform in mobile & desktop
* Prevent Freece in macos when you pick a file

Fixes: https://github.com/miguelpruivo/flutter_file_picker/issues/2037,
Fixes: https://github.com/miguelpruivo/flutter_file_picker/issues/2036, 
Fixes: https://github.com/miguelpruivo/flutter_file_picker/issues/2055,
Fixes: https://github.com/miguelpruivo/flutter_file_picker/issues/2056
Fixes: https://github.com/miguelpruivo/flutter_file_picker/issues/1903


lib/src/api/platform_file.dart
```diff
@@ -79,28 +67,27 @@ class PlatformFile {
...
-  /// to a [NSURL](https://developer.apple.com/documentation/foundation/nsurl) on iOS.
-  /// Is set to `null` on all other platforms since those are all already referencing the original file content.
-  ///
-  /// This property is `null` in the following cases:
-  /// - On Web, where local file paths are not available (it may point to a Blob URL instead).
-  /// - On Android, when picking a directory or using SAF without caching enabled.
-  ///
-  /// Note: You can't use this to create a Dart `File` instance since this is a safe-reference for the original platform files, for
-  /// that the [path] property should be used instead.
-  final String? identifier;
```
ekuleshov, line 92: what would be the replacement for the identifier? 

* on Android selected file path and content:// urls aren't anywhere similar
* on iOS to have permanent urls for selected files they has to be converted to [scoped bookmarks](https://www.avanderlee.com/swift/security-scoped-bookmarks-for-url-access/) also see [ns_file_coordinator_util](https://pub.dev/packages/ns_file_coordinator_util)

ekuleshov:
@daniJimen I'd like to confirm about #2037 issue. The request is to be able to get a persistent resource identifier on all platforms, which could be read without copying data.

E.g. a `content://` url on Android could be selected from a gallery app or other source and passed directly to `video_player` and other consumers that could read it directly from the source. There is a similar NSURL API on iOS as well.

daniJimen:
> @daniJimen I'd like to confirm about #2037 issue. The request is to be able to get a persistent resource identifier on all platforms, which could be read without copying data.
> 
> E.g. a `content://` url on Android could be selected from a gallery app or other source and passed directly to `video_player` and other consumers that could read it directly from the source. There is a similar NSURL API on iOS as well.

Good day, @ekuleshov.
This is actually the exact behavior we're using. We keep the original URI (for example, on Android, a content:// URI). When we need to save the file to a specific location, we derive the target path from that content:// URI and pass only the file path to the saveFile function, and the savefile can access to file, and write bytes in selected folder.
You always have access to the file reference when picking a file, because the returned object (PlatformFile) contains the URI in its path property. On Android, this is typically a content:// URI. Using that URI, you can access the file and perform any additional operations you may need beyond simply calling saveFile.
the point its that actually you only access to bytes in savefile method, and you don,t need to write file in a temporal path

ekuleshov:
> @daniJimen ...returned object (PlatformFile) contains the URI in its path property

If I'm following your explanation this is a significant behavior change. The path now has the url (maybe it shouldn't be called path then?) and consumer app would have to use some special APIs in order to read content of selected file (as far as I know the content:// urls aren't directly readable in Dart). Ideally we need separate APIs to read full file and to open a stream for a given url.

In my case I need that url to be persistent for life. So, the app can access and work with selected large media files without copying them in the app's own space. However the `AndroidSAFOptions` is being removed in this PR. How will we control the permissions grant for select url?

Also what about supporting persistent urls on other platforms, such as iOS?

PR #2078, opened 2026-06-19
Update build.gradle.kts

morisedigitalsolution:


vicajilau:
Hi @morisedigitalsolution, thank you for opening this PR!

However, we cannot accept this change in its current form. The conditional check `if (shouldApplyKotlinAndroidPlugin)` was specifically introduced to maintain compatibility with Android Gradle Plugin (AGP) 9.0+, where Kotlin support is built into AGP directly. 

Unconditionally applying `apply(plugin = "org.jetbrains.kotlin.android")` causes plugin collisions and build failures on AGP 9+. You can refer to the official Android documentation regarding built-in Kotlin support in AGP for more details:
 [Android Gradle Plugin Release Notes & Migration Guide](https://developer.android.com/build/releases/gradle-plugin)

If you are experiencing a specific build failure or issue with the current logic in your environment, please feel free to share your Gradle/AGP version details and build logs so we can investigate further.

Closing for now. Thanks again for your contribution!

PR #2079, opened 2026-06-23
feat(windows): add parentWindowHandle parameter to override window handle lookup

aexei:
**Description:**

Adds an optional `int? parentWindowHandle` parameter to pickFiles, pickFile, saveFile, and getDirectoryPath. When provided alongside lockParentWindow: true, the given handle is used directly as hwndOwner instead of calling FindWindowA internally.

This is useful when the app manages multiple windows or the automatic lookup returns the wrong handle.

**Usage:**

``` dart
await FilePicker.pickFiles(
  lockParentWindow: true,
  parentWindowHandle: myHwnd, // raw HWND as int
);
```

If parentWindowHandle is null, behavior is unchanged — _getWindowHandle() is used as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 - Decodes images with `inSampleSize` in `compressImage` to prevent excessive memory usage and `OutOfMemoryError` on large images. [#2083](https://github.com/miguelpruivo/flutter_file_picker/pull/2083)
 
 ### Desktop & Web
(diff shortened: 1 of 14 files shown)


vicajilau (approved):

vicajilau:
Thank you so much for this contribution and for your work on this PR, @aexei! 👏

Your code served as the foundation for a broader refactoring where we have encapsulated platform-specific settings into dedicated configuration classes (`WindowsOptions`, `LinuxOptions`, and `WebOptions`). 

Wrapping custom parameters like `parentWindowHandle` inside these platform options allows us to keep the public API clean, flexible, and scalable while significantly improving the overall Developer Experience (DX).

Your feature and improvements are now pushed and merged into master, and will be included in the upcoming `12.0.0-beta.8` release. Thank you again for your valuable contribution! 🚀

vicajilau (approved):

PR #2088, opened 2026-07-20
fix(macos): pass dialogTitle in getDirectoryPath (#2087)

### LINKED ISSUE #2087: regression dialogTitle ignored on MacOS
I found a regression: file_picker 11.0.2 accepts dialogTitle in Dart but drops it before calling the macOS directory panel.

vicajilau:
Fixes #2087 by including 'dialogTitle' in the method channel arguments map for getDirectoryPath on macOS.

### DIFF
--- a/test/file_picker_macos_test.dart
+++ b/test/file_picker_macos_test.dart
@@ -160,21 +160,27 @@ void main() {
       log.clear();
     });
 
-    test('should pass dialogTitle and initialDirectory to method channel', () async {
-      final picker = FilePickerMacOS();
+    test(
+      'should pass dialogTitle and initialDirectory to method channel',
+      () async {
+        final picker = FilePickerMacOS();
 
-      final result = await picker.getDirectoryPath(
-        dialogTitle: 'Select Directory',
-        initialDirectory: '~/Documents',
-      );
+        final result = await picker.getDirectoryPath(
+          dialogTitle: 'Select Directory',
+          initialDirectory: '~/Documents',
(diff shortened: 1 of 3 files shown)


vicajilau:
@navaronbracke I have removed the obsolete `escapeDialogTitle` helper from the macOS platform picker. Manual escaping is a legacy code from the old `osascript` CLI implementation. Since we now use standard Flutter `MethodChannel` and Swift's native APIs, strings are marshaled safely without escaping. Leaving it in actually displays unwanted backslashes in the native dialog.

test/file_picker_macos_test.dart
```diff
@@ -141,4 +106,45 @@ void main() {
...
+    final MethodChannel channel = const MethodChannel(
+      'miguelruivo.flutter.plugins.filepicker',
+    );
+    final List<MethodCall> log = <MethodCall>[];
+
+    setUp(() {
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(channel, (MethodCall methodCall) async {
+            log.add(methodCall);
+            return '/Users/test/Documents';
```
navaronbracke: Should this not check the method name, for clarity reasons?

navaronbracke (approved):
LGTM with nit

test/file_picker_macos_test.dart
vicajilau: Good point! Fixed

PR #2089, opened 2026-07-21
feat: add file_picker_platform_interface package and setup melos monorepo

vicajilau:
### Summary
  Initial step for plugin federation:
  - Added `melos.yaml` for managing the monorepo.
  - Extracted `file_picker_platform_interface` into `/packages/file_picker_platform_interface`.
  - Defined `FilePickerPlatform` interface, shared data models (`PlatformFile`, `FilePickerResult`, options, etc.), and default `MethodChannelFilePicker`.
  - Added unit tests for `FilePickerPlatform`.

  ### Test Plan
  - [x] Executed `flutter test` in `packages/file_picker_platform_interface` (All tests passing).
  - [x] Executed `flutter analyze` in `packages/file_picker_platform_interface` (0 issues found).
  - [x] Executed `flutter run`in example.


packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
```
navaronbracke: Should this be part of the Android implementation, instead of the platform implementation?

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
```diff
@@ -0,0 +1,28 @@
...
+
+/// An opaque handle to an Android Storage Access Framework [uri].
+final class AndroidSAFHandle {
+  const AndroidSAFHandle({required this.accessMode, required this.uri});
+
+  factory AndroidSAFHandle.fromMap(Map map) {
```
navaronbracke:
```suggestion
  factory AndroidSAFHandle.fromMap(Map<Object?, Object?> map) {
```

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
```diff
@@ -0,0 +1,28 @@
...
+  final AndroidSAFAccessMode accessMode;
+
+  /// The URI that this handle represents.
+  ///
+  /// Typically a `content://` URI.
+  final Uri uri;
+
+  /// Release the grant on the given [uri].
+  Future<void> releaseGrant() async {
+    await FilePickerPlatform.instance.releaseSAFGrant(uri.toString());
```
navaronbracke: Could this do the following, to avoid putting releaseSAFGrant() on the general platform interface, since it is Android specific?

```
if (FilePickerPlatform.instance case final FilePickerAndroid implementation) {
  await implementation.releaseSAFGrant(uri.toString());
}
```

packages/file_picker_platform_interface/lib/src/api/android_options.dart
```diff
@@ -0,0 +1,50 @@
+/// The access mode that is coupled to an [AndroidSAFGrant].
```
navaronbracke, line 1: This should also live in the Android implementation?

packages/file_picker_platform_interface/lib/src/api/exceptions.dart
```diff
@@ -0,0 +1,6 @@
+class IllegalCharacterInFileNameException implements Exception {
```
navaronbracke, line 3:
```suggestion
class IllegalCharacterInFileNameException implements Exception {
  const IllegalCharacterInFileNameException(this.message);
  
  final String message;
  
  @override
  String toString() => 'IllegalCharacterInFileNameException: $message';
}
```

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:cross_file/cross_file.dart';
+import 'package:file_picker_platform_interface/src/api/platform_file.dart';
+import 'package:flutter/foundation.dart';
+
+class FilePickerResult {
```
navaronbracke, line 5:
```suggestion
/// The result of a file picking operation.
class FilePickerResult {
```

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
...
+  const FilePickerResult(this.files);
+
+  /// Picked files.
```
navaronbracke, line 8:
```suggestion
  /// The files that were picked, if any.
```

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
...
+  final List<PlatformFile> files;
+
+  /// If this pick contains only a single resource.
+  bool get isSinglePick => files.length == 1;
+
+  /// The length of picked files.
+  int get count => files.length;
```
navaronbracke, line 15: This is redundant? `files` is public, so consumers can check this themselves.

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:flutter/foundation.dart';
+
+class FilePickerResult {
+  const FilePickerResult(this.files);
+
+  /// Picked files.
+  final List<PlatformFile> files;
+
+  /// If this pick contains only a single resource.
+  bool get isSinglePick => files.length == 1;
```
navaronbracke, line 12: This is a bit of an anti pattern? (I think back to Matan's original idea that a single file pick should always return a single file, or null)

What if we do this:
- `sealed class FilePickerResult {...}`
- `final class SingleFilePickerResult extends FilePickerResult`
- `final class MultiFilePickerResult extends FilePickerResult`

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
...
+
+  /// The length of picked files.
+  int get count => files.length;
+
+  /// A `List<String>` containing all paths from picked files.
+  List<String?> get paths => files
```
navaronbracke, line 18: These getters feel redundant? I think that FilePickerResult should just be a dumb container for PlatformFile.

packages/file_picker_platform_interface/lib/src/api/file_picker_result.dart
```diff
@@ -0,0 +1,48 @@
...
+  bool operator ==(Object other) {
+    if (identical(this, other)) {
+      return true;
+    }
+
+    return other is FilePickerResult && listEquals(other.files, files);
+  }
+
+  @override
+  int get hashCode => files.hashCode;
```
navaronbracke, line 44: This is a wrong implementation of a hash code for list elements. collection can do that too for us

packages/file_picker_platform_interface/lib/src/api/windows_options.dart
```diff
@@ -0,0 +1,16 @@
...
+    this.parentWindowHandle,
+    this.lockParentWindow = false,
+  });
+
+  /// An optional Win32 window handle (`HWND`) to override the default parent
+  /// window lookup when `lockParentWindow` is enabled.
+  final int? parentWindowHandle;
+
+  /// Whether the child window (file picker window) will stay in front of the
+  /// Flutter window until it is closed (like a modal window).
```
navaronbracke, line 14:
```suggestion
  /// Flutter window until it is closed (like a modal window).
  ///
  /// Defaults to false.
```

pubspec.yaml
```diff
@@ -3,7 +3,23 @@ description: A package that allows you to use a native file explorer to pick sin
 homepage: https://github.com/miguelpruivo/plugins_flutter_file_picker
 repository: https://github.com/miguelpruivo/flutter_file_picker
 issue_tracker: https://github.com/miguelpruivo/flutter_file_picker/issues
-version: 12.0.0-beta.8
+version: 12.0.0
+
+workspace:
+  - packages/file_picker_platform_interface
+
+melos:
```
navaronbracke, line 11: This section is defined in melos.yml as well?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 12.0.0
+### General
```
navaronbracke, line 2: I think we should relist the changelog from the beta releases? (or at least mention them somehow)

packages/file_picker_platform_interface/lib/src/platform_file_web_fetch_stub.dart
```diff
@@ -0,0 +1,9 @@
+import 'dart:typed_data';
```
navaronbracke, line 1: This should not be here, but in the web implementation, which makes this file redundant in general?

packages/file_picker_platform_interface/lib/src/platform_file_web_fetch.dart
```diff
@@ -0,0 +1,94 @@
+import 'dart:js_interop';
```
navaronbracke, line 1: Since this imports `dart:js_interop` this needs to be in the web implementation only.

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+
+import 'package:cross_file/cross_file.dart';
+import 'package:flutter/foundation.dart';
+
+import 'android_saf_handle.dart';
+import '../platform_file_web_fetch_stub.dart'
+    if (dart.library.js_interop) '../platform_file_web_fetch.dart';
+
+/// Represents a file returned by the file picker.
+class PlatformFile {
```
navaronbracke, line 11: I think PlatformFile should be an interface, instead of a class. (maybe an abstract class)

Then we can:
- have a `PlatformFileWeb` in the web implementation, which can call the web fetch utility directly, without having to do conditional imports
- have a PlatformFileAndroid, which knows how to decode SAF handles from the underlying Map

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+        file: file,
+        safHandle: AndroidSAFHandle.fromMap(safHandle),
+      );
+    }
+
+    return file;
+  }
+
+  /// The absolute path for a cached copy of this file.
+  final String? path;
```
navaronbracke, line 33: I think path should only be available on non-web version of this interface and we should not define it here

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  /// File name including its extension.
+  final String name;
+
+  /// Byte data for this file.
+  @Deprecated(
+    'Use readAsBytes() instead to avoid out-of-memory issues with large files.',
+  )
+  final Uint8List? bytes;
```
navaronbracke, line 42: Was this deprecated before? I would remove it in v12.0

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  /// File content as a stream of bytes.
+  @Deprecated('Use readAsByteStream() instead')
+  final Stream<List<int>>? readStream;
```
navaronbracke, line 46: Same here

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+  final int size;
+
+  /// The platform identifier for the original file.
+  final String? identifier;
+
+  /// File extension for this file.
+  String? get extension => name.split('.').last;
+
+  /// Retrieves this as a XFile
+  XFile get xFile {
```
navaronbracke, line 58: Make this an abstract getter in the interface

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+    if (kIsWeb && bytes != null) {
+      return XFile.fromData(bytes!, name: name, length: size);
+    } else {
+      return XFile(path!, name: name, bytes: bytes, length: size);
+    }
+  }
+
+  /// Read the file content as bytes.
+  Future<Uint8List> readAsBytes() async {
```
navaronbracke, line 67: Make this an abstract function and reimplement in the subclasses

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+
+    throw StateError(
+      'PlatformFile.readAsBytes(): file data is not available. '
+      'Consume the file via PlatformFile.readAsByteStream(), or on Web ensure '
+      'the file path is a fetchable blob/data URL that can be retrieved.',
+    );
+  }
+
+  /// Read the file content as a stream of bytes.
+  Stream<Uint8List> readAsByteStream() async* {
```
navaronbracke, line 93: Make this an abstract function and reimplement in the subclasses

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+
+      yield await readAsBytes();
+      return;
+    }
+
+    yield* xFile.openRead();
+  }
+
+  /// Returns the length of the file in bytes.
+  Future<int> length() async {
```
navaronbracke, line 109: Make this an abstract function and reimplement in the subclasses

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+        other.bytes == bytes &&
+        other.readStream == readStream &&
+        other.identifier == identifier &&
+        other.size == size;
+  }
+
+  @override
+  int get hashCode {
+    return kIsWeb
+        ? 0
```
navaronbracke, line 136: 0 on web seems like a bug?

packages/file_picker_platform_interface/lib/src/api/platform_file.dart
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  @override
+  String toString() {
+    return 'PlatformFile(${kIsWeb ? '' : 'path $path'}, name: $name, bytesLength: ${bytes?.lengthInBytes}, readStream: ${readStream != null}, size: $size)';
+  }
+}
+
+/// A [PlatformFile] implementation that includes a handle to an Android Storage Access Framework URI.
+class AndroidPlatformFile extends PlatformFile {
```
navaronbracke: This should be moved to the Android implementation

packages/file_picker_platform_interface/lib/src/method_channel/file_picker_method_channel.dart
```diff
@@ -0,0 +1,202 @@
...
+  @override
+  Future<bool?> clearTemporaryFiles() async =>
+      methodChannel.invokeMethod<bool>('clear');
+
+  @override
+  Future<String?> getDirectoryPath({
+    String? dialogTitle,
+    bool lockParentWindow = false,
+    String? initialDirectory,
+    AndroidSAFOptions? androidSafOptions,
```
navaronbracke: Now that I think of it, should it just be `AndroidOptions` ? (which the Android implementation then provides an implementation for with `AndroidSAFOptions`, for access to the handle)

packages/file_picker_platform_interface/lib/src/method_channel/file_picker_method_channel.dart
```diff
@@ -0,0 +1,202 @@
...
+    allowedExtensions,
+    onFileLoading,
+    withData,
+    withReadStream,
+    compressionQuality,
+    androidSafOptions,
+  );
+
+  @override
+  Future<void> releaseSAFGrant(String uri) async {
```
navaronbracke: Only Android should implement this, not every method channel implementation

packages/file_picker_platform_interface/lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,112 @@
+import 'dart:io';
+import 'dart:isolate';
+import 'dart:typed_data';
+
+import 'package:file_picker_platform_interface/src/api/platform_file.dart';
+import 'package:path/path.dart';
+
+Future<List<PlatformFile>> filePathsToPlatformFiles(
```
navaronbracke, line 8: I think we should remove this, since it is up to the individual implementations to implement PlatformFile creation? withReadStream/withData should be removed anyway

packages/file_picker_platform_interface/lib/src/utils/_file_utils_io.dart
```diff
@@ -0,0 +1,112 @@
...
+      path: nativeFile.path,
+      readStream: readStream,
+      size: nativeFile.existsSync() ? nativeFile.lengthSync() : 0,
+    );
+  }
+
+  throw ArgumentError('Expected file to be a File.');
+}
+
+Future<String?> runExecutableWithArguments(
```
navaronbracke, line 51: Should `runExecutableWithArguments` + `isExecutableOnPath` be in a different file? I.e. `runnable_executable_utils` ? Since that is only for desktop. Just for organizational purposes

packages/file_picker_platform_interface/lib/src/utils/_file_utils_web.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:typed_data';
+import 'package:file_picker_platform_interface/src/api/platform_file.dart';
+
+Future<List<PlatformFile>> filePathsToPlatformFiles(
```
navaronbracke, line 4: Can we investigate removing this entire file, when we have the web implementation and a native implementation?

packages/file_picker_platform_interface/lib/src/utils/file_picker_utils.dart
```diff
@@ -0,0 +1,79 @@
...
+    as impl;
+
+/// Utility class for file picker that provides common helper methods
+/// used across different platform implementations.
+class FilePickerUtils {
+  /// The default title for the file picker dialog.
+  static const String defaultDialogTitle = 'File Picker';
+
+  /// Converts a list of file paths into a list of [PlatformFile]s.
+  static Future<List<PlatformFile>> filePathsToPlatformFiles(
```
navaronbracke, line 16: Like with the other file, I think that `filePathsToPlatformFiles`, `createPlatformFile`, `runExecutableWithArguments`, `isExecutableOnPath`, `saveBytesToFile` should not exist in this file

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  static FilePickerPlatform get instance => _instance;
+
+  static set instance(FilePickerPlatform instance) {
+    PlatformInterface.verifyToken(instance, _token);
+    _instance = instance;
+  }
+
+  /// Retrieves the file(s) from the underlying platform
+  Future<FilePickerResult?> pickFiles({
```
navaronbracke, line 41: Because this is a major version, I would make a hard breaking change here:

- pickFiles specifically returns a `MultiFilePickerResult`
- if no files were picked, the list of files is empty, instead of null
- remove `allowMultiple`, `withData`, `withReadStream`
- can we remove `readSequential` ? Or was that specifically for web? If it was, `readSequential` should go in `WebOptions`
- is `cancelUploadOnWindowBlur` specific to web or desktop? Can it be moved to the options classes?

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+    List<String>? allowedExtensions,
+    Function(FilePickerStatus)? onFileLoading,
+    int compressionQuality = 0,
+    bool allowMultiple = false,
+    bool withData = false,
+    bool withReadStream = false,
+    bool lockParentWindow = false,
+    bool readSequential = false,
+    bool cancelUploadOnWindowBlur = true,
+    AndroidSAFOptions? androidSafOptions,
```
navaronbracke: Why is this one nullable and the other options not?

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+  }) async {
+    throw UnimplementedError(
+      'pickFileAndDirectoryPaths() has not been implemented.',
+    );
+  }
+
+  /// Releases the given SAF grant.
+  Future<void> releaseSAFGrant(String uri) async {
```
navaronbracke: This method should not be here, but in the Android implementation

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+  Future<bool?> clearTemporaryFiles() async {
+    throw UnimplementedError('clearTemporaryFiles() has not been implemented.');
+  }
+
+  /// Selects a directory and returns its absolute path.
+  Future<String?> getDirectoryPath({
+    String? dialogTitle,
+    bool lockParentWindow = false,
+    String? initialDirectory,
+    AndroidSAFOptions? androidSafOptions,
```
navaronbracke: Do directories on Android ever return SAF handles? If not then I think this should just be `AndroidOptions`, in line with the other method

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+    String? dialogTitle,
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+    Function(FilePickerStatus)? onFileLoading,
+    int compressionQuality = 0,
+    bool allowMultiple = false,
+    bool withData = false,
+    bool withReadStream = false,
+    bool lockParentWindow = false,
```
navaronbracke, line 54: Should lockParentWindow be part of the options classes instead? I don't think it applies for Android/iOS?

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Asks the underlying platform to remove any temporary files created by this plugin.
+  Future<bool?> clearTemporaryFiles() async {
+    throw UnimplementedError('clearTemporaryFiles() has not been implemented.');
+  }
+
+  /// Selects a directory and returns its absolute path.
+  Future<String?> getDirectoryPath({
+    String? dialogTitle,
+    bool lockParentWindow = false,
```
navaronbracke, line 90: Same here for lockParentWindow

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+  /// name to save a file.
+  Future<String?> saveFile({
+    String? dialogTitle,
+    required String fileName,
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+    required Uint8List bytes,
+    Function(FilePickerStatus)? onFileLoading,
+    bool lockParentWindow = false,
```
navaronbracke, line 113: Same here for `lockParentWindow`

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
```diff
@@ -0,0 +1,121 @@
...
+    WindowsOptions windowsOptions = const WindowsOptions(),
+    LinuxOptions linuxOptions = const LinuxOptions(),
+    WebOptions webOptions = const WebOptions(),
+  }) async {
+    throw UnimplementedError('saveFile() has not been implemented.');
+  }
+
+  Future<void> skipEntitlementsChecks() async {
```
navaronbracke: This should only be present in the MacOS implementation, not in the platform interface

navaronbracke (changes requested):

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
vicajilau: It needs to stay in `file_picker_platform_interface` because `FilePickerPlatform` itself references Android SAF types across several places:

1. `FilePickerPlatform.pickFiles()` and `getDirectoryPath()` accept `AndroidSAFOptions` as parameters.
2. `FilePickerPlatform` defines `releaseSAFGrant(String uri)`.
3. `PlatformFile.fromMap()` constructs an `AndroidPlatformFile` containing `AndroidSAFHandle`.

Keeping these types in `platform_interface` ensures that the base interface and shared models remain self-contained without introducing circular dependencies with `file_picker_android`.

packages/file_picker_platform_interface/lib/src/api/android_options.dart
vicajilau: Same reason as `AndroidSAFHandle`. `AndroidSAFOptions` is used directly in `FilePickerPlatform.pickFiles()` and `FilePickerPlatform.getDirectoryPath()` parameter signatures. 

Having platform-specific options classes (`AndroidSAFOptions`, `WindowsOptions`, `LinuxOptions`, `WebOptions`) defined in `platform_interface` is the standard pattern used across official Flutter federated plugins so callers can pass options to `FilePickerPlatform.instance` without needing to import platform-specific implementation packages.

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: Good point! `AndroidSAFOptions` was only introduced in `v12.0.0-beta.1` and isn't part of any stable release yet. 

We can rename it directly to `AndroidOptions` without needing a `@Deprecated` typedef, keeping the API clean and consistent for the 12.0.0 release.

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: Exactly! `lockParentWindow` has already been moved to `WindowsOptions.lockParentWindow` and `LinuxOptions.lockParentWindow` (and similarly `cancelUploadOnWindowBlur` to `WebOptions.cancelUploadOnWindowBlur`).

The top-level `lockParentWindow` and `cancelUploadOnWindowBlur` parameters are only kept on `pickFiles()` as deprecated parameters for backwards compatibility during the v12 migration, and will be removed in a future release.

I have added explicit `@Deprecated` annotations to those top-level parameters in `file_picker_platform_interface.dart` as well.

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: Fixed!

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: Fixed!

vicajilau:
Everyrthing ready for a new review @navaronbracke 🚀

melos.yaml
```diff
@@ -0,0 +1,13 @@
+name: file_picker
```
navaronbracke: This file can be deleted, per https://pub.dev/packages/melos#migrate-to-melos-7xx ?

pubspec.yaml
```diff
@@ -3,7 +3,23 @@ description: A package that allows you to use a native file explorer to pick sin
...
+  packages:
+    - .
```
navaronbracke, line 13: What do we need the root for under melos?

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
navaronbracke: @vicajilau I think we disagree here. I am referring to the official pattern in https://pub.dev/packages/webview_flutter#platform-specific-features

So methods that are specific to one platform (the releaseGrant() and skipEntitlements() methods for example), should not be in the general interface.

Likewise with the options, I think there should be a stub class `AndroidOptions` in the `file_picker_platform_interface`, but `file_picker_android` has

```
class FilePickerAndroidOptions extends AndroidOptions {}
```

which provides the default options, with things for the SAF handle

navaronbracke (changes requested):
@vicajilau I provided some extra guidance on what I think is needed according to platform specific things in the plugin.

Also, I see that not all the remaining feedback was addressed yet, fyi.

melos.yaml
vicajilau: Good point, fixed!

pubspec.yaml
vicajilau: Currently, the main `file_picker` package is still located at the root (`.`), while `file_picker_platform_interface` is under `packages/`.

Including `- .` ensures Melos continues to run `analyze` and `test` against the root `file_picker` package during this transition phase. 

Once we complete the refactor and move `file_picker` into `packages/file_picker` (or finish the workspace migration), `- .` won't be needed anymore and can be removed.

vicajilau: @navaronbracke , I can move it now but I would prefer to do it in a separate PR.

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
vicajilau: Ohh @navaronbracke , I think I understand what you're saying... I hadn't looked at it that way before. Thanks and fixed!

packages/file_picker_platform_interface/lib/src/api/android_saf_handle.dart
vicajilau: True, fixed!

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: True, fixed!

vicajilau:
One step closer @navaronbracke! Thank you very much for your patience. I hope it's more polished now

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
navaronbracke: @vicajilau We can do these in a follow-up?

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: True, fixed!

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: This has been removed

packages/file_picker_platform_interface/lib/file_picker_platform_interface.dart
vicajilau: Of course @navaronbracke. How do you suggest we proceed?

navaronbracke:
@vicajilau let's keep this open until the other bits have been added as well

navaronbracke:
This PR is now obsolete

PR #2094, opened 2026-07-28
File picker platform interface setup

navaronbracke:
This PR adds a package for `file_picker_platform_interface`, which will be the base for the federated plugin.

Note: This is only the platform interface itself.

@vicajilau Provide as much feedback on the API itself. There are some notable changes to the interface:

1) There is a `pickFile()` method that returns a PlatformFile, or null if canceled
2) `pickFiles()` returns a list of PlatformFile, instead of a wrapper class around the list
3) I removed the deprecated options for now (we could add those back later for the transition? I'd rather not, since there was a beta where it was deprecated, so people had time) We can write a migration guid on the Wiki and link it in the changelog.
I would rather bite the bullet and clean it up for good.
4) `pickFileAndDirectoryPaths` no longer returns null. I think that an empty list here is good enough for the same result. Saves a null check.
5) `clearTemporaryFiles()` now returns `Future<void>`. I don't know if the nullable boolean was ever useful.
6) `saveFile()` now no longer accepts allowed extensions / type. Instead I added a `mimeType` for the type hint and I would think that we make the `fileName` include the extension.
7) Since `saveFile()`'s return type was actually somewhat ambiguous / undocumented in places, I changed the return type to `Uri?`. Then we can do file paths, blob/data urls, and Android content Uris. Null is for when the operation is cancelled.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,18 +43,20 @@ jobs:
 
       - run: flutter --version
 
-      - name: Install dependencies
-        run: flutter pub get
+      - name: Install dependencies & Bootstrap
+        run: |
+          flutter pub get
+          dart run melos bootstrap
 
       - name: Verify formatting
         if: matrix.flutter == 'stable'
         run: dart format --output=none --set-exit-if-changed .
 
       - name: Analyze project source
-        run: dart analyze .
+        run: dart run melos run analyze
 
       - name: Run tests
-        run: flutter test
+        run: dart run melos run test
 
   test_os_matrix:
     needs: test_flutter_versions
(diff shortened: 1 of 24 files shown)


navaronbracke:
@vicajilau This is now ready for review.

Some notable omissions:
- `file_picker_android` should implement `AndroidOptions`, with the SAF handle things, including providing a `releaseGrant()` method in its own interface
- `file_picker_darwin` should implement `skipEntitlementChecks()` in its own package (not in the platform interface)
- `file_picker_web` should implement the fetching on the web for the blobs (your original utility was in the wrong place)
- `file_picker_linux/_windows/_web` should implement their options classes by extending the stubs

And I noticed that the executable finder method was unused, so I deleted that

vicajilau (approved):

PR #2099, opened 2026-07-29
feat(darwin): extract file_picker_darwin package for iOS and macOS

vicajilau:
## Description
This PR extracts the combined Apple platforms (iOS and macOS) implementation of `file_picker` into an independent federated package under `packages/file_picker_darwin` using `sharedDarwinSource: true`.

## Changes Included
- Created `packages/file_picker_darwin/` package configured for iOS and macOS platforms with `sharedDarwinSource: true`.
- Set up both CocoaPods (`file_picker_darwin.podspec`) and Swift Package Manager (`Package.swift`) configurations.
- Moved native Swift source code (`FilePickerPlugin.swift`, `IOSFilePickerHandler.swift`, `MacOSFilePickerHandler.swift`, `PrivacyInfo.xcprivacy`).
- Implemented `FilePickerDarwin` extending `FilePickerPlatform`.
- Implemented `DarwinPlatformFile extends PlatformFile`.
- Added unit tests in `packages/file_picker_darwin/test/file_picker_darwin_test.dart`.
- Added `packages/file_picker_darwin` to workspace in root `pubspec.yaml`.

## Verification
- `dart run melos bootstrap`: 3 packages bootstrapped successfully.
- `dart run melos run analyze`: 0 issues found across all packages.
- `dart run melos run test`: All unit tests passed across all packages.


packages/file_picker_darwin/darwin/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
navaronbracke, line 1: I think this should go under `Resources` and not `Classes`

packages/file_picker_darwin/darwin/Classes/MacOSFilePickerHandler.swift
```diff
@@ -0,0 +1,312 @@
...
+        case "pickFileAndDirectoryPaths":
+            handleFileAndDirectorySelection(call, result: result)
+
+        case "saveFile":
+            handleSaveFile(call, result: result)
+
+        case "skipEntitlementsChecks":
+            skipEntitlementsChecks = true
+            result(nil)
+
```
navaronbracke, line 43: iOS does implement `clear`, so should MacOS just no-op? Or maybe we do that at the file_picker_platform_interface level for that method? I don't think that web will implement it either.

WDYT?

packages/file_picker_darwin/lib/src/file_picker_darwin.dart
```diff
@@ -0,0 +1,191 @@
...
+          data,
+        ) {
+          if (data is! bool) return;
+          onFileLoading(
+            data ? FilePickerStatus.picking : FilePickerStatus.done,
+          );
+        }, onError: (error) => throw Exception(error));
+      }
+
+      final List<Map>? result = await methodChannel.invokeListMethod(typeName, {
```
navaronbracke: I noticed that this Map is no typed on K,V. Do we add a lint for this at some point?

navaronbracke (approved):

packages/file_picker_darwin/darwin/Resources/PrivacyInfo.xcprivacy
vicajilau: I initially placed it there following the structure in the official Flutter SPM guide (https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors), which places `PrivacyInfo.xcprivacy` inside the target sources directory.

However, separating non-code resources into a `Resources/` folder is indeed more typical in native iOS/macOS conventions and has no functional impact as long as `Package.swift` and `.podspec` are updated. 

I've moved it to `Resources/PrivacyInfo.xcprivacy` and updated both `Package.swift` and `file_picker_darwin.podspec`. Thanks!

packages/file_picker_darwin/darwin/Classes/MacOSFilePickerHandler.swift
vicajilau: Agree! Providing a default no-op implementation for `clearTemporaryFiles()` in `file_picker_platform_interface` makes a lot of sense:

```dart
Future<void> clearTemporaryFiles() async {}
```

Since macOS and Web access files in-place and don't create temporary copies in a plugin cache (unlike iOS and Android), platforms that don't need cleanup can just inherit the default no-op without needing boilerplate in their native channel handlers.

What do you think about updating `clearTemporaryFiles()` in `file_picker_platform_interface` to be a no-op by default?

navaronbracke: That is indeed my suggestion!

packages/file_picker_darwin/lib/src/file_picker_darwin.dart
vicajilau: Good spot! I've updated `invokeListMethod` to explicitly specify the generic types `<Map<Object?, Object?`.


Enabling strict linter rules like `strict-raw-types` in `analysis_options.yaml` across the monorepo sounds like a great idea to catch un-typed raw generics in future PRs!

navaronbracke: Also: side note: `strict-raw-types` will be deprecated in favor of a lint in the Dart SDK :P

vicajilau: giphy

Issue #2050, opened 2026-05-28
Migrate file_picker to a federated plugin architecture


vicajilau:
### Summary

This proposal suggests migrating the `file_picker` plugin to a **federated plugin architecture**. As the plugin grows and supports more platforms (Android, iOS, Web, macOS, Windows, and Linux), a federated approach will separate the platform-specific implementations into distinct packages. This will significantly improve maintainability, reduce the main package's dependency footprint, and allow platform experts to contribute more efficiently.

---

### Motivation

Currently, maintaining all platform implementations within a single package introduces several challenges:

* **Monolithic Maintenance:** A change or bug fix in the Windows implementation requires testing and releasing the entire package, potentially impacting users on other platforms.
* **Heavy Dependency Tree:** Users on mobile-only projects still inherit configuration or potential conflicts related to desktop or web implementations.
* **Contribution Barriers:** External developers who want to improve a single platform (e.g., Linux) have to navigate the entire codebase, making review and testing more cumbersome.

By moving to a federated model, we align with the current Flutter team best practices for multi-platform plugins.

---

### Proposed Architecture

The refactored plugin would be split into the following structure:

* **`file_picker` (App-facing package):** The main package that developers import. It defines the public API and forwards calls to the platform interface.
* **`file_picker_platform_interface` (Platform interface package):** Declares the common interface that all platform-specific packages must implement.
* **Platform-specific packages:**
* `file_picker_android`
* `file_picker_ios`
* `file_picker_web`
* `file_picker_windows`
* `file_picker_macos`
* `file_picker_linux`

---

### Implementation Steps

1. Create the `file_picker_platform_interface` package and define the abstract base classes.
2. Update the main `file_picker` package to depend on the platform interface.
3. Extract each platform's implementation into its respective sub-package.
4. Update the main package's `pubspec.yaml` to register the federated implementations using the `flutter.plugin.platforms` key.
5. Ensure backward compatibility so that existing users can upgrade without breaking changes to the public API.

PR #2100, opened 2026-07-29
feat(linux): extract Linux implementation into file_picker_linux package

vicajilau:
## Description

This PR extracts the Linux implementation of `file_picker` into a separate federated package (`packages/file_picker_linux`) as part of the ongoing package federation refactoring ([#2050](https://github.com/miguelpruivo/flutter_file_picker/issues/2050)).

### Key Changes

- **New Package**: Created `packages/file_picker_linux` implementing the Linux platform using the XDG Desktop Portal DB

### DIFF
--- a/packages/file_picker_linux/lib/src/file_picker_linux.dart
+++ b/packages/file_picker_linux/lib/src/file_picker_linux.dart
@@ -1,0 +1,293 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:dbus/dbus.dart';
+import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
+import 'package:flutter/foundation.dart';
+
+import 'file_picker_linux_options.dart';
+import 'filters.dart';
+import 'linux_platform_file.dart';
+import 'xdp_filechooser.dart';
+import 'xdp_request.dart';
+
+/// An implementation of [FilePickerPlatform] for Linux using XDG Desktop Portal.
+class FilePickerLinux extends FilePickerPlatform {
+  /// Registers this class as the default instance of [FilePickerPlatform].
+  static void registerWith() {
+    FilePickerPlatform.instance = FilePickerLinux();
+  }
+
+  final destination = "org.freedesktop.portal.Desktop";
+  DBusClient? _client;
+  OrgFreedesktopPortalFileChooser? _xdpChooser;
+
+  FilePickerLinux({
+    DBusClient? client,
+    OrgFreedesktopPortalFileChooser? xdpChooser,
+  })  : _client = client,
+        _xdpChooser = xdpChooser;
+
+  /// Gets the [DBusClient] instance, lazy-initializing it if needed.
+  DBusClient get client => _client ??= DBusClient.session();
+
+  /// Gets the [OrgFreedesktopPortalFileChooser] instance.
+  OrgFreedesktopPortalFileChooser get xdpChooser =>
+      _xdpChooser ??= OrgFreedesktopPortalFileChooser(
+        client,
+        destination,
+        path: DBusObjectPath("/org/freedesktop/portal/desktop"),
+      );
+
+  @override
+  Future<PlatformFile?> pickFile({
+    String? dialogTitle,
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+    Function(FilePickerStatus)? onFileLoading,
+    int compressionQuality = 0,
+    AndroidOptions androidOptions = const AndroidOptions(),
+    WindowsOptions windowsOptions = const WindowsOptions(),
+    LinuxOptions linuxOptions = const FilePickerLinuxOptions(),
+    WebOptions webOptions = const WebOptions(),
+  }) async {
+    final files = await pickFiles(
+      dialogTitle: dialogTitle,
+      initialDirectory: initialDirectory,
+      type: type,
+      allowedExtensions: allowedExtensions,
+      onFileLoading: onFileLoading,
+      compressionQuality: compressionQuality,
+      linuxOptions: linuxOptions,
+    );
+    return files.firstOrNull;
+  }
+
+  @override
+  Future<List<PlatformFile>> pickFiles({
+    String? dialogTitle,
+    String? initialDirectory,
+    FileType type = FileType.any,
+    List<String>? allowedExtensions,
+    Function(FilePickerStatus)? onFileLoading,
+    int compressionQuality = 0,
+    AndroidOptions androidOptions = const AndroidOptions(),
+    WindowsOptions windowsOptions = const WindowsOptions(),
+    LinuxOptions linuxOptions = const FilePickerLinuxOptions(),
+    WebOptions webOptions = const WebOptions(),
+  }) async {
+    final FilePickerLinuxOptions options = switch (linuxOptions) {
+      FilePickerLinuxOptions opts => opts,
+      _ => const FilePickerLinuxOptions(),
+    };
+
+    final filter = Filter(type, allowedExtensions);
+    Map<String, DBusValue> xdpOption = {
+      'handle_token': DBusString('flutter_picker'),
+      'multiple': DBusBoolean(true),
+      'modal': DBusBoolean(options.lockParentWindow),
+      'filters': filter.toDBusArray(),
+    };
+    if (initialDirectory != null) {
+      final Uint8List tmp = _encodeDirectory(initialDirectory);
+      DBusArray directory = DBusArray.byte(tmp);
+      xdpOption["current_folder"] = directory;
+    }
+
+    final replyPath = await xdpChooser.callOpenFile(
+      _formatParentWindow(options.parentWindow),
+      dialogTitle ?? "flutter picker",
+      xdpOption,
+    );
+
+    List<Uri> uriPaths = [];
+
+    final request = OrgFreedesktopPortalRequest(
+      client,
+      destination,
+      path: replyPath,
+    );
+
+    await for (var response in request.response) {
+      final status = response.response;
+      if (status != 0) {
+        return [];
+      }
+      final result = response.results;
+      uriPaths =
+          result["uris"]
+              ?.asArray()
+              .map((data) => Uri.parse(data.asString()))
+              .toList() ??
+          [];
+      break;
+    }
+
+    final filePaths = uriPaths.map((uri) => uri.toFilePath()).toList();
+
+    return filePaths.map((path) => LinuxPlatformFile.fromPath(path)).toList();
+  }
+
+  @override
+  Future<List<String>> pickFileAndDirectoryPaths({
+    String? dialogTitle,
+    String? initialDirectory,
+    FileType type = FileType.any,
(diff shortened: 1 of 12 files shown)


packages/file_picker_linux/lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,293 @@
...
+import 'xdp_request.dart';
+
+/// An implementation of [FilePickerPlatform] for Linux using XDG Desktop Portal.
+class FilePickerLinux extends FilePickerPlatform {
+  /// Registers this class as the default instance of [FilePickerPlatform].
+  static void registerWith() {
+    FilePickerPlatform.instance = FilePickerLinux();
+  }
+
+  final destination = "org.freedesktop.portal.Desktop";
```
navaronbracke: static const? This is also lacking `String`

packages/file_picker_linux/lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,293 @@
...
+      uriPaths =
+          result["uris"]
+              ?.asArray()
+              .map((data) => Uri.parse(data.asString()))
+              .toList() ??
+          [];
+      break;
+    }
+
+    final filePaths = uriPaths.map((uri) => uri.toFilePath()).toList();
```
navaronbracke: This could be a for literal and do the 2 maps in one go? Saves an iteration

packages/file_picker_linux/lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,293 @@
...
+    return filePaths.firstOrNull;
```
navaronbracke: I do not think you need to call toList() if we do firstOrNull on iterables?

```
return uriPaths.map((uri) => uri.toFilePath()).firstOrNull;
```

packages/file_picker_linux/lib/src/file_picker_linux.dart
```diff
@@ -0,0 +1,293 @@
...
+      final file = File(savedFilePath);
+      await file.writeAsBytes(bytes);
+      return Uri.file(savedFilePath);
+    }
+
+    return null;
+  }
+
+  Uint8List _encodeDirectory(String initialDirectory) {
+    return Uint8List.fromList([...utf8.encode(initialDirectory), 0]);
```
navaronbracke, line 272: We need a null byte at the back?

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+        (allowedExtensions == null || allowedExtensions.isEmpty)) {
+      throw ArgumentError(
+        'If type is set to FileType.custom, allowedExtensions cannot be null or empty.',
+      );
+    }
+    switch (type) {
+      case FileType.any:
+        return;
+      case FileType.audio:
+        final audio = ["*.aac", "*.midi", "*.mp3", "*.ogg", "*.wav"];
```
navaronbracke: I wonder if these should be private static const lists?

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+        List<(int, String)> audioList = [];
```
navaronbracke: this can be final and a for literal

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+        for (var filter in audio) {
+          audioList.add((0, filter));
+        }
+        info["Audio"] = audioList;
+      case FileType.custom:
+        final custom = [...?allowedExtensions];
```
navaronbracke: same here for the for literal

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+        List<(int, String)> customList = [];
+        for (var filter in custom) {
+          customList.add((0, "*.$filter"));
+        }
+        info["Custom"] = customList;
+      case FileType.image:
+        final image = ["*.bmp", "*.gif", "*.jpeg", "*.jpg", "*.png", "*.webp"];
+        List<(int, String)> imageList = [];
```
navaronbracke: same here for the for literal

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+          "*.webm",
+          "*.wmv",
+          "*.bmp",
+          "*.gif",
+          "*.jpeg",
+          "*.jpg",
+          "*.png",
+        ];
+        List<(int, String)> mediaList = [];
+        for (var filter in media) {
```
navaronbracke: same here for the for literal

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+          "*.flv",
+          "*.mkv",
+          "*.mov",
+          "*.mp4",
+          "*.m4v",
+          "*.mpeg",
+          "*.webm",
+          "*.wmv",
+        ];
+        List<(int, String)> videoList = [];
```
navaronbracke: same here for the for literal

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+        for (var filter in video) {
+          videoList.add((0, filter));
+        }
+        info["Video"] = videoList;
+    }
+  }
+  DBusArray toDBusArray() {
```
navaronbracke, line 78:
```suggestion

  DBusArray toDBusArray() {
```

packages/file_picker_linux/lib/src/filters.dart
```diff
@@ -0,0 +1,106 @@
...
+    List<DBusValue> dataList = [];
+    info.forEach((var key, var values) {
```
navaronbracke: Use a regular for loop? I dislike forEach, unless using a tear off (and I recall there is a lint against it too)

packages/file_picker_linux/lib/src/xdp_request.dart
```diff
@@ -0,0 +1,55 @@
+// This file was generated using the following command and may be overwritten.
+// dart-dbus generate-remote-object request.xml
```
navaronbracke, line 2: Same remark for request.xml

packages/file_picker_linux/lib/src/xdp_filechooser.dart
```diff
@@ -0,0 +1,91 @@
+// This file was generated using the following command and may be overwritten.
```
navaronbracke, line 1: Should we document how to install dart-dbus? (I assume it is installed globally via Pub)

Also, what is `org.freedesktop.portal.FileChooser.xml` ? Should that be checked in somewhere?

navaronbracke (changes requested):
Just some nits, but this was easier than expected to review :)

packages/file_picker_linux/lib/src/file_picker_linux.dart
vicajilau: Good point!

packages/file_picker_linux/lib/src/file_picker_linux.dart
vicajilau: Good catch!

packages/file_picker_linux/lib/src/file_picker_linux.dart
vicajilau: Done!!

packages/file_picker_linux/lib/src/file_picker_linux.dart
vicajilau: Yes, the trailing null byte (`0`) is required here. The XDG Desktop Portal specification expects `current_folder` as a null-terminated C-string byte array (`ay` DBus signature) for the underlying GLib/C portal daemon.

packages/file_picker_linux/lib/src/filters.dart
vicajilau: Agree!!

packages/file_picker_linux/lib/src/filters.dart
vicajilau: Done!

packages/file_picker_linux/lib/src/filters.dart
vicajilau: Updated `toDBusArray` to use a standard `for (final entry in info.entries)` loop and collection `for` instead of `.forEach` with a function literal. Thanks for pointing that out!

packages/file_picker_linux/lib/src/xdp_filechooser.dart
vicajilau: > Should we document how to install `dart-dbus`?

Yes! I've updated `packages/file_picker_linux/README.md` to document how to install `dart-dbus` via Pub global (`dart pub global activate dbus`) and how to run the `generate-remote-object` command.

> Also, what is `org.freedesktop.portal.FileChooser.xml`? Should that be checked in somewhere?

That header comment is produced automatically by the `dart-dbus` CLI tool when generating remote object bindings against the DBus interface specification.  

Following standard Flutter plugin conventions (like `url_launcher_linux`), we don't need to check in raw `.xml` introspection files into the repository—keeping the generated `.dart` files and documenting the `dart-dbus` generation steps in `README.md` is sufficient.

packages/file_picker_linux/lib/src/xdp_request.dart
vicajilau: Review https://github.com/miguelpruivo/flutter_file_picker/pull/2100#discussion_r3680093491

vicajilau:
Ready for another review @navaronbracke!!

packages/file_picker_linux/lib/src/linux_platform_file.dart
```diff
@@ -0,0 +1,82 @@
...
+  @override
+  Stream<Uint8List> readAsByteStream() {
+    return xFile.openRead();
+  }
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other is! LinuxPlatformFile) return false;
+    return super == other && other.name == name && other.uri == uri;
```
navaronbracke: What is `super == other` for specifically? PlatformFile is abstract, so this is always going to be false? Nothing can be equal to `PlatformFile`

navaronbracke (approved):
LGTM, with a question (technically applies to all implementations of PlatformFile)

packages/file_picker_linux/lib/src/linux_platform_file.dart
navaronbracke: @vicajilau I think we should remove the `super == other` since the super class delegates to de default Object `==` implementation, which is the same as the identical() above.

vicajilau: Good catch! Since `PlatformFile` inherits `Object.==`, calling `super == other` was redundant with `identical(this, other)`. Removed `super == other` and updated `hashCode`!!

PR #2103, opened 2026-07-30
feat(web): extract file_picker_web package

vicajilau:
## Description
Extracts the Web platform implementation of `file_picker` into its own federated package `packages/file_picker_web`.

## Changes
- Created `packages/file_picker_web` package using modern `package:web` (`web: ^1.1.1`).
- Implemented `FilePickerWeb` and `WebPlatformFile` in the federated package.
- Updated `WebOptions` in `file_picker_platform_interface`.
- Registered `packages/file_picker_web` in root `pubspec.yaml` workspace.
- Added unit tests in `packages/file_picker_web/test/`.

## Verification
- `melos run analyze` passes with 0 issues across all monorepo packages.
- `melos run test` passes 100% cleanly.


packages/file_picker_web/lib/src/web_file_input_session.dart
```diff
@@ -0,0 +1,107 @@
...
+  Future<List<PlatformFile>?> start() async {
+    final uploadInput = HTMLInputElement()
+      ..type = 'file'
+      ..draggable = true
+      ..multiple = webOptions.allowMultiple
+      ..accept = accept
+      ..style.display = 'none';
+
+    if (onFileLoading != null) {
+      onFileLoading!(FilePickerStatus.picking);
```
navaronbracke: Nit: `onFileLoading?.call(FilePickerStatus.picking);`

packages/file_picker_web/lib/src/web_file_input_session.dart
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    uploadInput.onChange.listen(_onFileSelection);
+    uploadInput.addEventListener('change', _onFileSelection.toJS);
```
navaronbracke, line 65: What is the difference withe the line above?

packages/file_picker_web/lib/src/web_file_input_session.dart
```diff
@@ -0,0 +1,107 @@
...
+    if (onFileLoading != null) {
+      onFileLoading!(FilePickerStatus.done);
+    }
+    _completer.complete(pickedFiles);
+  }
+
+  void _onCancel(Event _) {
+    window.removeEventListener('focus', _onCancel.toJS);
+
+    Future.delayed(const Duration(seconds: 1)).then((_) {
```
navaronbracke: Is 1 second needed? What about 500ms?

packages/file_picker_web/lib/src/web_file_input_session.dart
navaronbracke, line 65: Are the listeners removed when a file is picked? (not just on cancel)

packages/file_picker_platform_interface/lib/src/file_picker_options/web_options.dart
```diff
@@ -1,5 +1,64 @@
 /// The options for the web file picker.
 class WebOptions {
```
navaronbracke, line 2: Why is this different, compared to the other implementations? Should this be `FilePickerWebOptions extends WebOptions` (with FilePickerWebOptions declared in file_picker_web)?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+import 'web_platform_file.dart';
+
+/// An implementation of [FilePickerPlatform] for the Web platform.
+///
+/// Uses standard HTML5 `<input type="file">` element interop via `package:web`
+/// to provide single and multiple file picking, as well as file saving capabilities
+/// in browser environments.
+class FilePickerWeb extends FilePickerPlatform {
+  late Element _target;
+  final String _kFilePickerInputsDomId = '__file_picker_web-file-input';
```
navaronbracke: This can be static const?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+
+  final int _readStreamChunkSize = 1000 * 1000; // 1 MB
```
navaronbracke: This can be static const?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+  @override
+  Future<String?> getDirectoryPath({
+    String? dialogTitle,
+    String? initialDirectory,
+    AndroidOptions androidOptions = const AndroidOptions(),
+    WindowsOptions windowsOptions = const WindowsOptions(),
+    LinuxOptions linuxOptions = const LinuxOptions(),
+    WebOptions webOptions = const WebOptions(),
+  }) async {
+    throw UnimplementedError('getDirectoryPath() has not been implemented.');
```
navaronbracke: Use `UnsupportedError` instead, with an error message that indicates that web does not support it?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  /// Opens an HTML file input dialog to pick a single file.
```
navaronbracke: I would remove this doc comment (it will be outdated when we use a private helper method anyway)

packages/file_picker_platform_interface/lib/src/file_picker_options/web_options.dart
```diff
@@ -1,5 +1,64 @@
...
+  const WebOptions({
+    this.allowMultiple = false,
+    this.withData = true,
+    this.withReadStream = false,
+    this.readSequential = false,
+    this.cancelUploadOnWindowBlur = true,
+  });
+
+  /// Creates a copy of this [WebOptions] with the given fields replaced.
+  WebOptions copyWith({
```
navaronbracke: Do we really need a copyWith() ?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+      compressionQuality: compressionQuality,
+      androidOptions: androidOptions,
+      windowsOptions: windowsOptions,
+      linuxOptions: linuxOptions,
+      webOptions: webOptions.copyWith(allowMultiple: false),
+    );
+    return files.firstOrNull;
+  }
+
+  /// Opens an HTML file input dialog to pick one or more files.
```
navaronbracke: I would remove this comment (the platform interface already provides docs and the options you mention here will change anyway)

packages/file_picker_web/lib/src/web_file_input_session.dart
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  void _onFileSelection(Event e) async {
+    if (_eventTriggered) return;
+    _eventTriggered = true;
+
+    final targetInput = e.target as HTMLInputElement?;
+    final files = targetInput?.files;
+    if (files == null) {
+      _completer.complete(null);
```
navaronbracke: Should we guard against 2x completing the completer?

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  /// Reads an HTML [File] content into a [Uint8List] using [FileReader].
+  Future<Uint8List?> _readSingleFileBytes(File file) async {
+    final completer = Completer<Uint8List?>();
+    final reader = FileReader();
+
+    reader.onLoadEnd.listen((_) {
+      final byteBuffer = (reader.result as JSArrayBuffer?)?.toDart;
+      completer.complete(byteBuffer?.asUint8List());
```
navaronbracke: Should we guard against 2x completing the completer?

packages/file_picker_web/lib/src/web_platform_file.dart
```diff
@@ -0,0 +1,115 @@
...
+  final String name;
+
+  /// The web URI representing this file (typically a `blob:` or `data:` URL).
+  @override
+  final Uri uri;
+
+  final XFile? _xFile;
+  final int? _bytesLength;
+  final Uint8List? _bytes;
+  final Stream<List<int>>? _readStream;
```
navaronbracke: `Stream<List<int>>` should be replaced with `Stream<Uint8List>` everywhere. `Uint8List` is an optimized version of `List<int>` (also check other usages of the type elsewhere)

packages/file_picker_web/test/file_picker_web_test.dart
```diff
@@ -0,0 +1,17 @@
...
+library;
+
+import 'package:file_picker_web/file_picker_web.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('WebPlatformFile properties work correctly', () {
+    final file = WebPlatformFile(
+      name: 'test.txt',
+      uri: Uri.parse('blob:http://localhost/123'),
```
navaronbracke, line 11: Is `blob:http:` a valid Uri type? I would expect blob, data or http, but not blob + http?

packages/file_picker_web/README.md
```diff
@@ -0,0 +1,7 @@
+# file_picker_web
+
+The Web implementation of [`file_picker`](https://pub.dev/packages/file_picker).
```
navaronbracke, line 3: This README format is not consistent with the other packages? (altho I do like the idea of linking to Pub like this)

navaronbracke (changes requested):

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: Done! Replaced `if (onFileLoading != null)` with `onFileLoading?.call(...)` across the session handler.

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: Good catch! `uploadInput.onChange.listen` was a leftover duplicate from the original implementation before adding `addEventListener('change', ...)`. 

I've removed the redundant `onChange` listener.

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: The delay is necessary because when the file picker dialog closes, browsers (especially Chrome/Safari) fire the window `focus` event a few milliseconds before the input `change` event. 

Without a small delay, the `focus` listener would resolve the picker with `null` before `change` has a chance to execute. 

Reduced the delay from 1s to 500ms, which is plenty of time for the event loop to process `change` while feeling significantly faster for cancelled dialogs.

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: Good catch! Added a `_cleanupListeners` helper that removes the `focus` listener on `window` and the `change`/`cancel` listeners on the `uploadInput` as soon as a file selection or cancellation occurs.

packages/file_picker_platform_interface/lib/src/file_picker_options/web_options.dart
vicajilau: I agree... Updated `WebOptions` in `file_picker_platform_interface` to be a base options class (`const WebOptions();`).

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Good catch! Updated both `_kFilePickerInputsDomId` and `_readStreamChunkSize` to `static const` fields at the top of `FilePickerWeb`.

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Done!

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Makes complete sense! Updated `getDirectoryPath()` to throw `UnsupportedError('getDirectoryPath() is not supported on Web.')` instead of `UnimplementedError`.

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Done! Removed the doc comment from `pickFile`.

packages/file_picker_platform_interface/lib/src/file_picker_options/web_options.dart
vicajilau: I agree

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: Good safety check!

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Done!

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Added `if (!completer.isCompleted)` guards for both `loadend` and `error` event listeners on `FileReader` in `_readSingleFileBytes()`.

packages/file_picker_web/lib/src/web_platform_file.dart
vicajilau: Great catch!

packages/file_picker_web/README.md
vicajilau: True! Updated.

packages/file_picker_web/test/file_picker_web_test.dart
vicajilau: I ran the following code snippet directly in Chrome using `flutter test --platform chrome`:

```dart
final blob = Blob(['hello world'.toJS].toJS);
final blobUrl = URL.createObjectURL(blob);
print('REAL BROWSER BLOB URL: $blobUrl');
```

And Chrome outputted the following URL:

```text
REAL BROWSER BLOB URL: blob:http://localhost:50290/c8a19e49-bd9c-46f3-bf7b-196925075520
```

According to the W3C File API specification, Blob URLs follow the structure `blob:<origin>/<uuid>`. When running on an `http://localhost` origin, the browser produces `blob:http://localhost:...`, where `blob:` is the scheme and `http://...` represents the origin path.

vicajilau:
Everything ready for another review @navaronbracke 🚀

packages/file_picker_web/lib/src/file_picker_web.dart
```diff
@@ -0,0 +1,325 @@
...
+    final blob = Blob([bytes.toJS].toJS, BlobPropertyBag(type: mimeType));
+    final url = URL.createObjectURL(blob);
+
+    HTMLAnchorElement()
+      ..href = url
+      ..target = 'blank'
+      ..download = fileName
+      ..click();
+
+    URL.revokeObjectURL(url);
```
navaronbracke, line 276: If we revoke the object URL here, then it is invalid anyway for Uri.parse? Should this return null instead?

packages/file_picker_web/lib/src/file_picker_web_options.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
+
+/// Configuration options specific to the Web platform.
+final class FilePickerWebOptions extends WebOptions {
+  /// Whether to return bytes (`PlatformFile.bytes`) in memory upon file picking.
+  final bool withData;
+
+  /// Whether to create a read stream (`PlatformFile.readStream`) for each picked file.
```
navaronbracke:
```suggestion
  /// Whether to create a read stream for each picked file.
```

packages/file_picker_web/lib/src/web_file_input_session.dart
```diff
@@ -0,0 +1,125 @@
...
+  final bool allowMultiple;
+
+  /// Web-specific configuration options.
+  final FilePickerWebOptions webOptions;
+
+  /// Optional callback triggered during file picking status transitions.
+  final Function(FilePickerStatus)? onFileLoading;
+
+  /// Callback delegate used to process and convert the native [FileList] to [PlatformFile]s.
+  final Future<List<PlatformFile>> Function(FileList, FilePickerWebOptions)
```
navaronbracke: Nit: Private typedef, for formatting purposes?

navaronbracke (changes requested):

packages/file_picker_web/lib/src/file_picker_web.dart
vicajilau: Spot on! Since `URL.revokeObjectURL(url)` immediately invalidates the Blob URL in browser memory (and browsers trigger the file download directly to the user's Downloads folder without exposing a persistent URI), returning `null` makes total sense. Updated `saveFile()` to return `null`.

packages/file_picker_web/lib/src/web_file_input_session.dart
vicajilau: Added `ProcessFilesCallback` private typedef for `processFiles` in `web_file_input_session.dart` formatted with `///` doc comment.

navaronbracke (approved):

PR #2104, opened 2026-07-31
fix(android): remove Apache Tika dependency for MIME type detection

### LINKED ISSUE #2101: Suggestion: Remove Apache Tika dependency on Android to reduce app size
Hi, first of all, thank you for maintaining this plugin.

We're using `file_picker` in a production app with millions of daily active users. At this scale, every metric matters, including binary size. Recently, while upgrading our Flutter SDK, we also bumped several plugin versions, including `file_picker`. After the upgrade, we noticed a significant increase in our Android app size, with approximately **300 KB** attributed to the addition of the Apache Tika dependency.

From what I understand, Tika is currently used to determine the MIME type when calling `saveFile()`. In our case, we don't use `saveFile()` at all—we only use the file picking APIs.

Since the primary purpose of this package is file picking, I was wondering if introducing a heavy dependency like Apache Tika for this use case is the best trade-off, especially considering the impact on all Android consumers regardless of whether they use `saveFile()`.

We forked the plugin internally and experimented with replacing the Tika-based MIME detection with the native Android APIs (`ContentResolver`/`MimeTypeMap`) for the `saveFile()` implementation. For our testing, the plugin behaved identically while completely removing the Tika dependency, and we recovered around **300 KB** in APK size.

Would you be open to incorporating this approach into the plugin? I'd be happy to open a PR if it aligns with the project's direction. Alternatively, if there are edge cases or reasons Tika was chosen that I'm not aware of, I'd love to hear your thoughts. Perhaps there's another approach that preserves the functionality while avoiding the additional dependency.

Thanks again for all your work on this plugin.

ryanaidilp:
## Summary
- Removes the Apache Tika dependency (~300 KB) from the legacy `android/` implementation, used only to detect a MIME type for `saveFile()`.
- MIME types are now resolved from the file name extension via `MimeTypeMap` (with the existing CSV special-case preserved), falling back to `URLConnection.guessContentTypeFromStream` content sniffing only when the file name has no usable extension.
- Removes the now-unused Tika proguard rules and the `tika-core` dependency from `build.gradle.kts`.

Companion PRs:
- Backport of this same fix targeting the `11.0.x` line (based on `v11.0.2`).
- A separate PR applying the same fix to `file_picker_android` specifically.

Fixes #2101

## Test plan
- [x] `flutter build apk --debug` in `example/` compiles cleanly against this change.
- Java/Kotlin code in this plugin has no existing unit test harness (per `CONTRIBUTING.md`, native platform code is not currently unit tested), so verification was via compilation and manual review of the MIME-mapping logic against the previous Tika-based behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Android
 - Decodes images with `inSampleSize` in `compressImage` to prevent excessive memory usage and `OutOfMemoryError` on large images. [#2083](https://github.com/miguelpruivo/flutter_file_picker/pull/2083)
+- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
 
 ### Desktop & Web
 - Added `WindowsOptions` (with `parentWindowHandle` and `lockParentWindow`), `LinuxOptions` (with `parentWindow` and `lockParentWindow`), and `WebOptions` (with `cancelUploadOnWindowBlur`). [#2079](https://github.com/miguelpruivo/flutter_file_picker/pull/2079)
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -48,7 +48,7 @@ object FileUtils {
     private const val CSV_MIME_TYPE = "text/csv"
     // Fallback MIME type used when neither the file name extension nor
     // content sniffing can determine a type, matching the value used by
-    // most content-detection libraries (and previously Apache Tika) for
+    // most content-detection libraries for
     // unrecognized binary content.
     private const val DEFAULT_MIME_TYPE = "application/octet-stream"
     // Maximum dimension (width or height) allowed when decoding an image for
@@ -332,7 +332,7 @@ object FileUtils {
 
     // Sniffs [bytes] for a MIME type using URLConnection's built-in magic-number
     // detection. This is a best-effort fallback for when the file name has no
-    // usable extension, without pulling in a full content-detection dependency.
+    // usable extension.
     private fun guessMimeTypeFromBytes(bytes: ByteArray?): String? {
         if (bytes == null || bytes.isEmpty()) {
             return null
--- a/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -22,18 +22,16 @@ import io.flutter.plugin.common.MethodChannel
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import org.apache.tika.Tika
-import org.apache.tika.io.TikaInputStream
-import org.apache.tika.metadata.Metadata
-import org.apache.tika.metadata.TikaCoreProperties
 import java.io.BufferedInputStream
 import java.io.BufferedOutputStream
+import java.io.ByteArrayInputStream
 import java.io.File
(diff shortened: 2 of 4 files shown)


android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -48,6 +46,11 @@ object FileUtils {
     // (see https://android.googlesource.com/platform/frameworks/base/+/61ae88e/core/java/android/webkit/MimeTypeMap.java#439)
     private const val CSV_EXTENSION = "csv"
     private const val CSV_MIME_TYPE = "text/csv"
+    // Fallback MIME type used when neither the file name extension nor
+    // content sniffing can determine a type, matching the value used by
+    // most content-detection libraries (and previously Apache Tika) for
```
navaronbracke: Since Tika will be removed, I would remove the reference to it in docs. I don't think that years into the future, when someone reads this code, they will have the original context.

```suggestion
    // most content-detection libraries for
```

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -303,31 +306,44 @@ object FileUtils {
...
+        return if (detectedType == null || detectedType == "text/plain") {
             "*/*"
         } else {
             detectedType
         }
     }
 
+    // Sniffs [bytes] for a MIME type using URLConnection's built-in magic-number
+    // detection. This is a best-effort fallback for when the file name has no
+    // usable extension, without pulling in a full content-detection dependency.
```
navaronbracke: Same reason as above

```suggestion
    // usable extension.
```

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -303,31 +306,44 @@ object FileUtils {
...
+                return CSV_MIME_TYPE
+            }
+            MimeTypeMap.getSingleton()
+                .getMimeTypeFromExtension(extension.lowercase(Locale.getDefault()))
+                ?.let { return it }
         }
-        return if (detectedType == "text/plain") {
+
+        val detectedType = guessMimeTypeFromBytes(bytes)
+        return if (detectedType == null || detectedType == "text/plain") {
```
navaronbracke, line 326: Just for clarification, `text/plain` is a legitimate catch all here?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Android
 - Decodes images with `inSampleSize` in `compressImage` to prevent excessive memory usage and `OutOfMemoryError` on large images. [#2083](https://github.com/miguelpruivo/flutter_file_picker/pull/2083)
+- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
```
navaronbracke, line 7: I was under the impression that this should go in a hotfix for v11?

Or should we skip that and include it only in v12 (in a new beta or not)

navaronbracke (changes requested):

android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
ryanaidilp: This is preserved from the original Tika-based implementation, not something new — the old code already mapped a detected `text/plain` to `*/*` before this change. It reads as a deliberate choice to avoid over-restricting the `ACTION_CREATE_DOCUMENT` intent when the content is generic/ambiguous plain text (e.g. an unknown-extension text file), so I kept the same fallback rather than changing it as part of this PR.

CHANGELOG.md
ryanaidilp: Agreed — this should go out as a v11.0.x hotfix too, in addition to master. Given Tika predates v11, the same functional fix applies there; happy to prepare that patch whenever there's a release branch to target it against. This PR stands on its own as the master-branch fix.

navaronbracke (approved):

PR #2105, opened 2026-07-31
fix(file_picker_android): remove Apache Tika dependency for MIME type detection

### LINKED ISSUE #2101: Suggestion: Remove Apache Tika dependency on Android to reduce app size

ryanaidilp:
## Summary
- Removes the Apache Tika dependency (~300 KB) from the federated `file_picker_android` package, used only to detect a MIME type for `saveFile()`.
- MIME types are now resolved from the file name extension via `MimeTypeMap` (with the existing CSV special-case preserved), falling back to `URLConnection.guessContentTypeFromStream` content sniffing only when the file name has no usable extension.
- Removes the now-unused Tika proguard rules and the `tika-core` dependency from `build.gradle.kts`.
- Bumps `file_picker_android` to `1.0.1` with a CHANGELOG entry.

This is the `file_picker_android`-specific PR requested in #2101, as a companion to:
- The equivalent fix for the still-bundled legacy `android/` implementation on `master`.
- A backport of the same fix targeting the `11.0.x` line (based on `v11.0.2`).

Fixes #2101

## Test plan
- [x] The Kotlin change mirrors the identical fix already verified to compile cleanly in the legacy `android/` implementation (same imports, same dependency versions).
- This package has no example app of its own in the repo, so the change wasn't independently exercised through a full Gradle build here — it's textually identical to the already-verified fix.

### DIFF
--- a/packages/file_picker_android/CHANGELOG.md
+++ b/packages/file_picker_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.1
+
+- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
+
 ## 1.0.0
 
 - Initial release of Android implementation package for `file_picker`.
--- a/packages/file_picker_android/CHANGELOG.md
+++ b/packages/file_picker_android/CHANGELOG.md
@@ -1,7 +1,4 @@
-## 1.0.1
-
-- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
-
 ## 1.0.0
 
 - Initial release of Android implementation package for `file_picker`.
+- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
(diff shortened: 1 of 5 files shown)


packages/file_picker_android/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: file_picker_android
 description: Android implementation of the file_picker plugin.
-version: 1.0.0
+version: 1.0.1
```
navaronbracke: This should be reverted, file_picked_android is not yet published iirc (Although there is a different naming issue that we need to look at, apparently there's a naming conflict with the plugin name)

packages/file_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1
+
+- Removed the Apache Tika dependency (~300 KB) used for MIME type detection in `saveFile()`. MIME types are now resolved from the file name extension via `MimeTypeMap`, with `URLConnection.guessContentTypeFromStream` as a content-sniffing fallback when no extension is available. [#2101](https://github.com/miguelpruivo/flutter_file_picker/issues/2101)
```
navaronbracke: This should be moved under `1.0.0`

navaronbracke (approved):
LGTM, with nits

navaronbracke (approved):

navaronbracke:
@ryanaidilp Feel free to open a pull request that we can hotfix to v11.x now :)

PR #2106, opened 2026-07-31
docs: attribute Tika removal to the 11.0.3 hotfix

ryanaidilp:
## Summary
Follow-up to #2104, which merged before this correction landed on that branch.

The Tika-removal fix targets the v11.x hotfix release (per discussion on #2104), so its CHANGELOG entry should sit under its own `11.0.3` heading rather than the unreleased `12.0.0-beta.8` bucket. This moves it there, positioned above `11.0.2` to match release chronology.

No code changes — `CHANGELOG.md` only.


navaronbracke (approved):

navaronbracke:
Yeah, my bad for landing them out of order :/

PR #2107, opened 2026-07-31
feat!: complete root package federation and legacy code cleanup

vicajilau:
## Description
This PR completes the **federation of the root `file_picker` package** (Phase 7 of the monorepo migration plan). The root package is now fully decoupled from legacy monolithic platform files and serves as the primary facing package, delegating all operations to the endorsed platform implementations.

## Key Changes
- **Default Package Endorsements**: Configured `pubspec.yaml` with `flutter.plugin.platforms` default endorsements for Android (`file_picker_android`), iOS/macOS (`file_picker_darwin`), Linux (`file_picker_linux`), Windows (`file_picker_windows`), and Web (`file_picker_web`).
- **Legacy Cleanup**: Removed legacy monolithic native directories (`android/`, `darwin/`) and obsolete implementation files in `lib/src/` (`lib/src/api/`, `lib/src/platform/`, `lib/src/utils/`).
- **Clean Conditional Exports**: Configured conditional exports in `lib/file_picker.dart` using `src/empty_stub.dart` to prevent `package:web`, `dart:ffi`, or `dbus` leakage across non-target environments.
- **PlatformFile Abstraction**: Declared `String? get path` in `PlatformFile` interface and implemented explicit platform overrides across all 5 platform packages (`WebPlatformFile.path` explicitly returns `null`).
- **Backward Compatibility**: Re-exported `FilePickerResult` and added `@Deprecated` parameter fallbacks (`androidSafOptions`, `lockParentWindow`, `withData`, etc.) in `FilePicker` facade class.
- **Unit & Integration Tests**: Added root package unit tests in `test/file_picker_test.dart` and updated `example/` demo app.

## Verification & Health
- `flutter analyze` (root package): **No issues found!**
- `melos run analyze`: **0 errors / 0 warnings** across all 7 monorepo packages.
- `melos run test`: **100% SUCCESS** across all unit and widget tests.

## Checklist
- [x] Code formatted via `dart format .`
- [x] All analyzer warnings and errors resolved
- [x] All unit and widget tests passing across monorepo
- [x] Example app dependencies and UI tests verified


android/gradle.properties
navaronbracke, line 1: Should this file be deleted? The entire root android directory should be gone?

android/proguard-rules.pro
navaronbracke, line 1: Same here

lib/src/empty_stub.dart
```diff
@@ -0,0 +1 @@
+// Intentionally empty stub file for conditional exports.
```
navaronbracke: I think this stub should be removed?

lib/file_picker.dart
```diff
@@ -1,19 +1,16 @@
...
-// Platform-specific implementations are exported only for plugin registration.
-// These exports are hidden on Web to avoid dart:ffi and dart:io compatibility issues.
-export 'src/platform/linux/file_picker_linux.dart'
-    if (dart.library.js_interop) 'src/platform/web/file_picker_web.dart';
-export 'src/platform/macos/file_picker_macos.dart'
-    if (dart.library.js_interop) 'src/platform/web/file_picker_web.dart';
-export 'src/platform/windows/file_picker_windows.dart'
-    if (dart.library.js_interop) 'src/platform/web/file_picker_web.dart';
+
+// Conditionally export platform-specific packages
```
navaronbracke: The app facing package should _never_ export the sub packages.

This is wrong, per the setup for federated plugins as a whole.

1) The app facing package depends on all of the sub packages internally.
2) If you require specific functionality for a specific platform, you add the sub package to your own pubspec.yaml explicitly (there is even a lint for this iirc)

For example, webview_flutter only exports its own API, not anything from the sub packages: https://github.com/flutter/packages/blob/main/packages/webview_flutter/webview_flutter/lib/webview_flutter.dart

The reason we had to deal with the stub in the past is exactly because this was misconfigured (and federation is the only solution, but you have to do it correctly!)

packages/file_picker_darwin/darwin/file_picker_darwin/Package.swift
```diff
@@ -17,10 +17,8 @@ let package = Package(
         .target(
             name: "file_picker_darwin",
             dependencies: [],
-            path: ".",
-            sources: ["Classes"],
```
navaronbracke, line 21: Why is `sources` removed here?

packages/file_picker_darwin/darwin/file_picker_darwin.podspec
```diff
@@ -13,9 +13,9 @@ iOS and macOS implementation of file_picker plugin
   s.license          = { :type => 'MIT', :file => '../LICENSE' }
   s.author           = 'Miguel Ruivo, Dominik Roszkowski'
   s.source           = { :path => '.' }
-  s.source_files     = 'Classes/**/*'
```
navaronbracke, line 16: Why does this need an update?

packages/file_picker_platform_interface/lib/src/method_channel_file_picker.dart
navaronbracke, line 7: @vicajilau Since each sub package provides its own implementation, should `MethodChannelFilePicker` be renamed to `NoOpFilePicker` ? Since it is essentially only required for the default value in the instance getter?

packages/file_picker_platform_interface/lib/src/file_picker_result.dart
```diff
@@ -0,0 +1,27 @@
+import 'platform_file.dart';
+
+/// The result of a file picking operation.
+class FilePickerResult {
```
navaronbracke, line 4: Per the breaking change, we don't need this class anymore?

It is replaced by `List<PlatformFile>`

example/lib/src/picked_directory_result.dart
```diff
@@ -1,4 +1,3 @@
-import 'package:file_picker/file_picker.dart';
```
navaronbracke, line 1: This should be restored? You will need to use an import for `file_picker_android` though

example/lib/src/file_picker_demo.dart
```diff
@@ -45,14 +45,6 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
   FileType _pickingType = FileType.any;
   List<PlatformFile>? pickedFiles;
 
-  String? get _customParentWindow =>
```
navaronbracke, line 48: Most of these changes should be restored?

The example can import `file_picker_android` and the like, to preserve behavior / API usage. (The example should depend on all the sub packages explicitly anyway, since it needs to demonstrate things like SAF handles)

lib/src/file_picker.dart
```diff
@@ -1,75 +1,26 @@
...
-import 'package:file_picker/src/api/platform_file.dart';
-import 'package:file_picker/src/api/file_picker_types.dart';
-import 'package:file_picker/src/api/android_saf_options.dart';
-import 'package:file_picker/src/api/windows_options.dart';
-import 'package:file_picker/src/api/linux_options.dart';
-import 'package:file_picker/src/api/web_options.dart';
 
+/// The primary entry point for picking files and directories across platforms.
 abstract final class FilePicker {
+  static AndroidOptions _resolveAndroidOptions(
```
navaronbracke, line 9: This should be removed. `file_picker` should be unaware of `file_picker_android`, except for passing through implementations of `AndroidOptions`

navaronbracke (changes requested):
I'm going to give some preliminary feedback.

This is not 100% ready to merge, but we will get there.

@vicajilau To reduce the diff from this PR I propose we do the following:

1) Delete the old platform code in one or more PRs (these are obsolete now)
- The entire /android folder
- The entire /darwin folder
- The entire /test folder
- The old Dart sources under /lib/src/api/**
- The old Dart sources under /lib/src/platform/**
- The old Dart sources under /lib/src/utils/**

2) Do the fixes for the platform implementations in a new PR
- file_picker_android
- file_picker_darwin
- file_picker_linux
- file_picker_web
- file_picker_windows
- file_picker_platform_interface

3) Update file_picker to use file_picker_platform_interface + update the example app in a last PR

lib/src/empty_stub.dart
vicajilau: This has been deleted in the latest commit

lib/file_picker.dart
vicajilau: This has been deleted in the latest commit

navaronbracke:
@vicajilau If you rebase this one, then the diff will probably be a lot smaller now

PR #2110, opened 2026-07-31
refactor: update root file_picker package and example app for federated architecture

vicajilau:
### Description
This is Step 3 (and final step) of the federated plugin migration roadmap. It updates the root `file_picker` package to endorse platform packages and use `file_picker_platform_interface`, and updates the `example` app.

### Changes Included
- **Root `file_picker` Package**:
  - Configured `pubspec.yaml` with endorsed platform packages under `flutter.plugin.platforms`.
  - Re-exported `file_picker_platform_interface` and clean platform-agnostic façade (`lib/file_picker.dart` and `lib/src/file_picker.dart`).
- **`file_picker_platform_interface`**:
  - Re-created and exported `FilePickerResult` (`lib/src/file_picker_result.dart`).
  - Added `String? get path` default getter implementation in `PlatformFile`.
  - Added `lockParentWindow` property to `WindowsOptions` and `LinuxOptions`.
- **Platform Packages**:
  - Updated `FilePickerWindowsOptions` and `FilePickerLinuxOptions` to inherit `lockParentWindow` from platform interface.
- **Example App (`example/`)**:
  - Updated `example/pubspec.yaml` with `resolution: workspace` and dependencies.
  - Updated `file_picker_demo.dart`, `picked_files_results.dart`, and `picked_directory_result.dart` to use new platform interface models and strong static typing.
  - Fixed integration widget test (`example/test/file_picker_demo_test.dart`).

### Verification
- `melos run analyze`: 0 issues / 0 warnings across all 7 packages.
- `melos run test`: 100% tests passing across all packages.
- `cd example && flutter build web`: Passed cleanly (`✓ Built build/web`).
- `cd example && flutter build macos --debug`: Passed cleanly (`✓ Built app bundle`).


packages/file_picker_platform_interface/lib/src/platform_file.dart
```diff
@@ -3,27 +3,23 @@ import 'dart:typed_data';
 import 'package:cross_file/cross_file.dart';
 
 /// The abstract representation of a picked file on the current platform.
-///
```
navaronbracke, line 6: Why are these doc lines removed?

packages/file_picker_platform_interface/lib/src/platform_file.dart
```diff
@@ -3,27 +3,23 @@ import 'dart:typed_data';
...
-/// Depending on the current platform,
-/// an implementation may provide additional properties or methods specific to that platform.
 abstract base class PlatformFile {
   /// The name of the underlying file, including the extension.
   String get name;
 
   /// An [Uri] to the underlying file.
-  ///
-  /// Depending on the platform, this may point to a local file, a blob, a data URI, or a network resource.
```
navaronbracke, line 15: Same here

packages/file_picker_platform_interface/lib/src/platform_file.dart
```diff
@@ -3,27 +3,23 @@ import 'dart:typed_data';
...
+
   /// Get this file as an [XFile].
   XFile get xFile;
 
   /// Get the length of the file in bytes.
   Future<int> length();
 
   /// Read the bytes of the file as a single chunk.
-  ///
-  /// Prefer [readAsByteStream] for large files.
```
navaronbracke, line 26: Same here

packages/file_picker_platform_interface/lib/src/file_picker_result.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:flutter/foundation.dart';
+import 'platform_file.dart';
+
+/// The result of a file picking operation containing a list of [PlatformFile].
+@immutable
+final class FilePickerResult {
```
navaronbracke: @vicajilau Did we need to reintroduce this class for something specific?

I don't see a use for it? It is just an alias for `List<PlatformFile>` really

packages/file_picker_platform_interface/lib/src/file_picker_options/linux_options.dart
```diff
@@ -1,5 +1,8 @@
 /// The options for the Linux file picker.
 class LinuxOptions {
   /// Creates an instance of [LinuxOptions].
-  const LinuxOptions();
```
navaronbracke, line 4: Should there be a `DesktopWindowOptions` mixin/superclass, that has a getter for `bool lockParentWindow` ?
Then Linux/Windows can inherit that

example/lib/src/file_picker_demo.dart
```diff
@@ -133,14 +118,8 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
           initialDirectory: _initialDirectoryController.text,
-          windowsOptions: WindowsOptions(
-            lockParentWindow: _lockParentWindow,
-            parentWindowHandle: _customWindowsHwnd,
-          ),
-          linuxOptions: LinuxOptions(
-            lockParentWindow: _lockParentWindow,
-            parentWindow: _customParentWindow,
-          ),
+          windowsOptions: WindowsOptions(lockParentWindow: _lockParentWindow),
```
navaronbracke, line 121: Were the custom parent window handles redundant for the example?

example/lib/src/file_picker_demo.dart
```diff
@@ -161,17 +140,18 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
...
-                androidPlatformFile.safHandle.releaseGrant();
-                _scaffoldMessengerKey.currentState?.showSnackBar(
-                  const SnackBar(content: Text("SAF Permission Released!")),
-                );
-                setState(() {
-                  pickedFiles!.removeAt(index);
-                  updateResults();
-                });
-              },
+          onRemoveAndroidFile: (int index, dynamic androidPlatformFile) {
```
navaronbracke: Can we avoid `dynamic` ?

What if `onRemoveAndroidFile` is a `Function(int, PlatformFile)` and we do an `platformFile is AndroidPlatformFile`

example/lib/src/file_picker_demo.dart
```diff
@@ -525,17 +493,8 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
         : null;
   }
 
-  AndroidSAFOptions? _androidSafOptionsFromFlags() {
```
navaronbracke, line 528: Restore this part? (but use an import of file_picker_android, for the SAF handle options)

example/pubspec.yaml
```diff
@@ -11,14 +13,14 @@ formatter:
 dependencies:
   flutter:
     sdk: flutter
-  file_picker:
-    path: ../
+  file_picker: ^12.0.0-beta.8
+  file_picker_android: ^1.0.0
```
navaronbracke, line 17: Add the other packages as well? That would make the example future proof if we need to add other things

lib/src/file_picker.dart
```diff
@@ -1,75 +1,26 @@
...
-import 'package:file_picker/src/api/platform_file.dart';
-import 'package:file_picker/src/api/file_picker_types.dart';
-import 'package:file_picker/src/api/android_saf_options.dart';
-import 'package:file_picker/src/api/windows_options.dart';
-import 'package:file_picker/src/api/linux_options.dart';
-import 'package:file_picker/src/api/web_options.dart';
 
+/// The primary entry point for picking files and directories across platforms.
 abstract final class FilePicker {
+  static AndroidOptions _resolveAndroidOptions(
```
navaronbracke, line 11: This is a helper function to ease the transition?

Can we document it as such?

lib/src/file_picker.dart
```diff
@@ -1,75 +1,26 @@
...
+    dynamic androidSafOptions,
```
navaronbracke:
```suggestion
   Object? androidSafOptions,
```

lib/src/file_picker.dart
```diff
@@ -1,75 +1,26 @@
...
+        return FilePickerAndroidOptions(safOptions: androidSafOptions);
+      } else if (androidSafOptions is AndroidOptions) {
+        return androidSafOptions;
+      }
+    }
+    return androidOptions;
+  }
+
   /// Retrieves the file(s) from the underlying platform
-  ///
```
navaronbracke, line 15: We should not remove the public facing documentation!

It should be updated to match the new API though

lib/src/file_picker.dart
navaronbracke, line 45: Since this is version 12.0, it should probably be removed now? (or, well in a new PR)

The same applies for the other now deprecated properties?

lib/src/file_picker.dart
```diff
@@ -101,38 +52,30 @@ abstract final class FilePicker {
       'Use WebOptions.cancelUploadOnWindowBlur instead; this parameter will be removed in a future release.',
     )
     bool cancelUploadOnWindowBlur = true,
-    AndroidSAFOptions? androidSafOptions,
+    @Deprecated('Use androidOptions instead.') dynamic androidSafOptions,
```
navaronbracke:
```suggestion
    @Deprecated('Use androidOptions instead.') Object? androidSafOptions,
```

@vicajilau Did we ever ship `androidSafOptions` in v11? If not, then I would just remove this deprecated property. I think that the betas can remove things at will, since it is not a stable release really

lib/src/file_picker.dart
```diff
@@ -1,75 +1,26 @@
...
-  ///
-  /// Note: This requires the User Selected File Read entitlement on macOS.
-  ///
-  /// Returns `null` if aborted.
-  /// selection; `pickFiles` now implies multiple selection by default.
-  /// NOTE: `withData`, `withReadStream` and `readSequential` are deprecated.
-  /// Call `PlatformFile.readAsBytes()` or `PlatformFile.readAsByteStream()` on
-  /// the returned `PlatformFile` to load data on demand. These parameters
-  /// will be removed in a future release.
   static Future<FilePickerResult?> pickFiles({
```
navaronbracke: Per what the platform_interface does, this should be `Future<List<PlatformFile>>` ?

Now it is divergent from the interface we agreed on :/ (which also brings me back to why we don't need FilePickerResult anymore)

lib/src/file_picker.dart
```diff
@@ -148,130 +91,70 @@ abstract final class FilePicker {
...
-  /// [initialDirectory] is invalid the user directory or previously valid
-  /// directory will be used.
-  ///
-  /// The file type filter [type] defaults to [FileType.any]. Optionally,
-  /// [allowedExtensions] might be provided (e.g. `["pdf", "svg", "jpg"]`).
-  ///
-  /// Returns a [Future<List<String>?>] that resolves to a list of absolute
-  /// paths for the selected files and directories. If the user cancels the
-  /// dialog or if the paths cannot be resolved, the method returns `null`.
   static Future<List<String>?> pickFileAndDirectoryPaths({
```
navaronbracke, line 134: This is not the same as the platform interface?

```suggestion
  static Future<List<String>> pickFileAndDirectoryPaths({
```

lib/src/file_picker.dart
```diff
@@ -148,130 +91,70 @@ abstract final class FilePicker {
...
-  /// This typically relates to cached files that are stored in the cache directory of
-  /// each platform and it isn't required to invoke this as the system should take care
-  /// of it whenever needed. However, this will force the cleanup if you want to manage those on your own.
-  ///
-  /// This method is only available on mobile platforms (Android & iOS).
-  ///
-  /// Returns `true` if the files were removed with success, `false` otherwise.
-  static Future<bool?> clearTemporaryFiles() {
-    return FilePickerPlatform.instance.clearTemporaryFiles();
+  static Future<bool?> clearTemporaryFiles() async {
```
navaronbracke: This is not the same as the platform interface

```suggestion
  static Future<void> clearTemporaryFiles() async {
```

lib/src/file_picker.dart
```diff
@@ -280,82 +163,42 @@ abstract final class FilePicker {
...
-  /// [LinuxOptions.lockParentWindow] instead.
-  ///
-  /// [windowsOptions] can be optionally set to configure Windows-specific options.
-  ///
-  /// [linuxOptions] can be optionally set to configure Linux-specific options.
-  ///
-  /// [webOptions] can be optionally set to configure Web-specific options.
-  ///
-  /// Returns `null` if aborted.
   static Future<String?> saveFile({
```
navaronbracke: This is not the same as the platform interface

```suggestion
  static Future<Uri?> saveFile({
```

navaronbracke (changes requested):
There is still some work left here, mainly stuff that is out of sync with `file_picker_platform_interface` itself.

packages/file_picker_platform_interface/lib/src/platform_file.dart
vicajilau: Good catch, thank you! I've restored all the missing doc comments in `PlatformFile`.

packages/file_picker_platform_interface/lib/src/platform_file.dart
vicajilau: Same

packages/file_picker_platform_interface/lib/src/platform_file.dart
vicajilau: Same

packages/file_picker_platform_interface/lib/src/file_picker_result.dart
vicajilau: True!

packages/file_picker_platform_interface/lib/src/file_picker_options/linux_options.dart
vicajilau: Great idea! I've created an abstract `DesktopWindowOptions` class with `final bool lockParentWindow` in `file_picker_platform_interface`, and made both `WindowsOptions` and `LinuxOptions` extend it.

example/lib/src/file_picker_demo.dart
vicajilau: Restored!

example/lib/src/file_picker_demo.dart
vicajilau: They were simplified to use the platform-agnostic `WindowsOptions` and `LinuxOptions` from `file_picker_platform_interface`, so the main example stays clean and platform-agnostic without importing platform-specific packages directly. 

If you'd prefer to keep showcasing custom native window handles (like HWND or X11/Wayland strings) in the example, I can use `FilePickerWindowsOptions` and `FilePickerLinuxOptions` instead. Let me know what you prefer!

example/lib/src/file_picker_demo.dart
vicajilau: Done! I've removed `dynamic` and updated `onRemoveAndroidFile` to take `(int index, AndroidPlatformFile androidPlatformFile)` strongly typed.

example/pubspec.yaml
vicajilau: Done!

lib/src/file_picker.dart
vicajilau: That is! Added documentation explaining that `_resolveAndroidOptions` is an internal helper function to handle backward compatibility between deprecated `androidSafOptions` and `androidOptions`. Also I have updated README's file.

lib/src/file_picker.dart
vicajilau: Restored and updated all public-facing Dartdoc comments on `FilePicker` methods (`pickFiles()`, `pickFile()`, `saveFile()`, `getDirectoryPath()`, and `clearTemporaryFiles()`) to accurately reflect the updated API.

lib/src/file_picker.dart
vicajilau: Yes, it's better to put it in a separate PR. I agree we should do that before releasing v12.

lib/src/file_picker.dart
vicajilau: Yes, you are right!

vicajilau: Good catch! Updated `FilePicker.pickFiles()` return type to `Future<List<PlatformFile>>` (non-nullable list, returning `[]` on cancel) to align 100% with `FilePickerPlatform.instance.pickFiles()`. Also updated the README and example app.

example/lib/src/file_picker_demo.dart
```diff
@@ -45,15 +46,8 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
   FileType _pickingType = FileType.any;
   List<PlatformFile>? pickedFiles;
```
navaronbracke: Should `pickedFiles` be `List<PlatformFile>` ?

vicajilau:
Everything ready for a new loop @navaronbracke 🚀

lib/src/file_picker.dart
```diff
@@ -1,76 +1,38 @@
...
+    return androidOptions;
+  }
+
+  /// Retrieves the file(s) from the underlying platform.
   ///
-  /// For more information, check the [API documentation](https://github.com/miguelpruivo/flutter_file_picker/wiki/api).
+  /// Opens a native file explorer and lets the user select one or multiple files.
   ///
-  /// Note: This requires the User Selected File Read entitlement on macOS.
+  /// The [type] parameter defines the type of files that can be selected (e.g. [FileType.image], [FileType.video], etc.).
```
navaronbracke, line 29: Not every parameter is documented? We can do this in a new PR for all the methods where this is the case

lib/src/file_picker.dart
navaronbracke: @vicajilau This is not resolved? The point here was to avoid users having to handle `null` as value. (which is what the platform interface does as well, since it does not allow null as return value)

lib/src/file_picker.dart
```diff
@@ -148,130 +106,73 @@ abstract final class FilePicker {
...
-  /// This typically relates to cached files that are stored in the cache directory of
-  /// each platform and it isn't required to invoke this as the system should take care
-  /// of it whenever needed. However, this will force the cleanup if you want to manage those on your own.
-  ///
-  /// This method is only available on mobile platforms (Android & iOS).
-  ///
-  /// Returns `true` if the files were removed with success, `false` otherwise.
-  static Future<bool?> clearTemporaryFiles() {
-    return FilePickerPlatform.instance.clearTemporaryFiles();
+  /// Returns `true` if the operation was initiated on a supported platform.
```
navaronbracke: It returns `Future<void>`, not `Future<bool>`

```suggestion
```

lib/src/file_picker.dart
```diff
@@ -281,81 +182,47 @@ abstract final class FilePicker {
...
-  /// [linuxOptions] can be optionally set to configure Linux-specific options.
-  ///
-  /// [webOptions] can be optionally set to configure Web-specific options.
+  /// The [fileName] parameter specifies the default file name for saving.
+  /// The [bytes] parameter contains the raw byte data to write.
+  /// The [mimeType] specifies the MIME type of the file.
   ///
-  /// Returns `null` if aborted.
+  /// Returns the saved file path or URI string, or `null` if the user canceled the operation.
   static Future<String?> saveFile({
```
navaronbracke: This is incorrect? This should return `Future<Uri?>`, as using the `Uri` class makes the result easier for consumers

README.md
```diff
@@ -18,20 +18,20 @@
 </p>
 
 # File Picker
-A package that allows you to use the native file explorer to pick single or multiple files, with extensions filtering support.
+A federated plugin that allows you to use the native file explorer to pick single or multiple files, with extensions filtering support.
```
navaronbracke: file_picker being federated is an implementation detail

```suggestion
A plugin that allows you to use the native file explorer to pick single or multiple files, with extensions filtering support.
```

README.md
```diff
@@ -18,20 +18,20 @@
...
 
 ## Currently supported features
 * Uses OS default native pickers
-* Supports multiple platforms (Mobile, Web, Desktop)
+* Supports multiple platforms (Mobile, Web, Desktop) via federated architecture
```
navaronbracke:
```suggestion
* Supports multiple platforms (Mobile, Web, Desktop)
```

README.md
```diff
@@ -51,76 +51,73 @@ See the [API section of the File Picker Wiki](https://github.com/miguelpruivo/fl
...
-   * [Filters](https://github.com/miguelpruivo/plugins_flutter_file_picker/wiki/API#filters)
-   * [Parameters](https://github.com/miguelpruivo/flutter_file_picker/wiki/API#parameters)
-   * [Methods](https://github.com/miguelpruivo/plugins_flutter_file_picker/wiki/API#methods)
-4. [FAQ](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ)
-5. [Troubleshooting](https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting)
+### Key Breaking Changes & Migration Steps
 
-## Usage
-Quick simple usage example:
+1. **`FilePicker.pickFiles()` Returns `List<PlatformFile>`**:
```
navaronbracke, line 62: Should this be moved to the Wiki?

README.md
```diff
@@ -51,76 +51,73 @@ See the [API section of the File Picker Wiki](https://github.com/miguelpruivo/fl
...
 
-```dart
-FilePickerResult? result = await FilePicker.pickFiles(
-  allowMultiple: true,
-  withData: false,
-  withReadStream: true,
-);
-```
+4. **Platform Options**:
+   - Platform-specific parameters are grouped into options objects:
```
navaronbracke:
```suggestion
   - Platform-specific parameters are grouped into configuration options, with implementations per platform:
```

navaronbracke (approved):
LGTM, although I would redo the documentation and double check the file_picker_platform_interface -> file_picker delegation in a different PR.

The documentation in `file_picker` does not match the documentation from the platform interface, and `file_picker` does not follow the `file_picker_platform_interface` in every method (the platform interface is the source of truth)

example/lib/src/file_picker_demo.dart
vicajilau: Great catch!

lib/src/file_picker.dart
vicajilau: Sounds good to me! Let's do a dedicated PR to thoroughly document all method parameters across `file_picker` and `file_picker_platform_interface`

lib/src/file_picker.dart
vicajilau: True. Updated `FilePicker.saveFile()` return type to `Future<Uri?>` directly matching `FilePickerPlatform.instance.saveFile()`, and updated the example app accordingly.

README.md
vicajilau: I agree that many developers overlook the GitHub Wiki and land straight on `pub.dev` or the `README.md` when upgrading packages (pub.dev doesn't surface the Wiki at all). 

Having a concise migration section directly in `README.md` provides immediate visibility for developers upgrading to v12. Long-term, building a dedicated documentation site (via GitHub Pages/Docusaurus) would be awesome, but for now, keeping this summary in the README ensures nobody misses these breaking changes when upgrading.

PR #2112, opened 2026-08-03
Refactor federated plugin

vicajilau:
## Description

This PR addresses [all items noted](https://github.com/miguelpruivo/flutter_file_picker/issues/1932#issuecomment-5128017014) in @navaronbracke's checklist for the federated plugin refactoring:

### 1. `file_picker_platform_interface` & `PlatformFile` Subclasses
- Made `FilePickerPlatform.clearTemporaryFiles()` a no-op by default.
- Removed dead parameters (`withData`, `withReadStream`) from native platform invocations.
- Removed `super == other` and `super.hashCode` in `operator ==` and `hashCode` across all `PlatformFile` subclasses (`AndroidPlatformFile`, `DarwinPlatformFile`, `LinuxPlatformFile`, `WindowsPlatformFile`, `WebPlatformFile`).
- Added validation throwing `ArgumentError` when constructing `PlatformFile` subclasses with empty `name`/`path`.
- Used `package:path`'s `basename()` to derive `name` when `name` is omitted but `path` is present.

### 2. Isolated `allowMultiple` Handling
- Updated `pickFile()` across all platform implementations (`FilePickerAndroid`, `FilePickerDarwin`, `FilePickerLinux`, `FilePickerWindows`, `FilePickerWeb`) so it no longer implicitly passes `allowMultiple: true` via delegation to `pickFiles()`. Instead, a private helper is reused with explicit `allowMultiple` flags.

### 3. `file_picker_android` (Kotlin)
- Cleaned up `Intent` creation logic for `allowMultiple`.
- Updated `clear` method in `FilePickerPlugin.kt` to return `result.success(null)`.

### 4. `file_picker_darwin` (Swift)
- Updated `clear` method in `IOSFilePickerHandler.swift` and `MacOSFilePickerHandler.swift` to invoke `result(nil)`.

### DIFF
--- a/packages/file_picker_android/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
+++ b/packages/file_picker_android/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
@@ -204,55 +204,43 @@ object FileUtils {
         if (type == "dir") {
             intent = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
         } else {
-            if (type == "image/*") {
-                intent = Intent(Intent.ACTION_GET_CONTENT).apply {
-                    type = this@startFileExplorer.type
-                    putExtra(Intent.EXTRA_ALLOW_MULTIPLE, this@startFileExplorer.isMultipleSelection)
-                    putExtra("multi-pick", this@startFileExplorer.isMultipleSelection)
+            intent = when (type) {
+                "image/*" -> Intent(Intent.ACTION_GET_CONTENT).apply {
+                    this.type = this@startFileExplorer.type
                     if (!allowedExtensions.isNullOrEmpty()) {
                         putExtra(Intent.EXTRA_MIME_TYPES, allowedExtensions!!.toTypedArray())
                     }
                 }
-            }
-            else if (type == "audio/*" ){
-                intent = Intent(Intent.ACTION_GET_CONTENT).apply {
+                "audio/*" -> Intent(Intent.ACTION_GET_CONTENT).apply {
                     this.type = this@startFileExplorer.type
                     addCategory(Intent.CATEGORY_OPENABLE)
-                    putExtra(Intent.EXTRA_ALLOW_MULTIPLE, this@startFileExplorer.isMultipleSelection)
-                    putExtra("multi-pick", this@startFileExplorer.isMultipleSelection)
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                         val authority = "com.android.providers.media.documents"
                         val audioRootUri = DocumentsContract.buildRootUri(authority, "audio_root")
                         putExtra(DocumentsContract.EXTRA_INITIAL_URI, audioRootUri)
                     }
                 }
-            } else if(type == "video/*"){
-                intent = Intent(Intent.ACTION_GET_CONTENT).apply {
+                "video/*" -> Intent(Intent.ACTION_GET_CONTENT).apply {
                     this.type = this@startFileExplorer.type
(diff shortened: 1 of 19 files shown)


packages/file_picker_android/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
```diff
@@ -204,55 +204,43 @@ object FileUtils {
...
             intent = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
         } else {
-            if (type == "image/*") {
-                intent = Intent(Intent.ACTION_GET_CONTENT).apply {
-                    type = this@startFileExplorer.type
-                    putExtra(Intent.EXTRA_ALLOW_MULTIPLE, this@startFileExplorer.isMultipleSelection)
-                    putExtra("multi-pick", this@startFileExplorer.isMultipleSelection)
+            intent = when (type) {
+                "image/*" -> Intent(Intent.ACTION_GET_CONTENT).apply {
+                    this.type = this@startFileExplorer.type
```
navaronbracke: Instead of `this.type = `, use `it.type = ` ?

packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/IOSFilePickerHandler.swift
```diff
@@ -32,8 +32,8 @@ final class IOSFilePickerHandler: NSObject,
         self.result = result
 
         if call.method == "clear" {
-            self.result?(clearTemporaryFiles())
-            self.result = nil
+            _ = clearTemporaryFiles()
+            result(nil)
```
navaronbracke: Also set self.result to nil? (or move `self.result = result` below his block)

packages/file_picker_web/lib/src/web_platform_file.dart
```diff
@@ -23,7 +23,8 @@ base class WebPlatformFile extends PlatformFile {
     int? bytesLength,
     Uint8List? bytes,
     Stream<Uint8List>? readStream,
-  }) : _xFile = xFile,
+  }) : assert(name.isNotEmpty, 'name cannot be empty'),
```
navaronbracke: Why is this an assert, instead of a throw?

navaronbracke (changes requested):

packages/file_picker_android/android/src/main/kotlin/com/mr/flutter/plugin/filepicker/FileUtils.kt
vicajilau: Good point! Cleaned up `this.type` to `type` inside the `.apply` configuration block.

packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/IOSFilePickerHandler.swift
vicajilau: Good catch! Moved the `call.method == "clear"` check to the very top of `IOSFilePickerHandler.handle(...)` before `self.result` is assigned, so it returns `result(nil)` immediately without locking `self.result` or triggering `multiple_request`.

packages/file_picker_web/lib/src/web_platform_file.dart
vicajilau: True, replaced `assert(name.isNotEmpty)` with throwing `ArgumentError('name cannot be empty')`.

navaronbracke (approved):

PR #2134, opened 2026-08-15
fix(darwin): write saveFile bytes to disk on macOS

### LINKED ISSUE #2132: file_picker_darwin: `saveFile` does not write bytes to the selected file path.
**Describe the bug**
saveFile does not write bytes to the selected file path
Works on Windows.

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop (MacOS)

**Platform OS version**
26.6.1 (25G76)

**Error Log**
No error logs produced

**Flutter Version details**
3.44.8, on macOS 26.6.1 25G76 darwin-arm64, locale en-NL)

vicajilau:
## Summary
- `saveFile` on macOS resolved a destination path via `NSSavePanel` but never wrote the provided bytes to it, so saved files ended up empty.
- The Dart layer now writes the bytes to the resolved path after the native panel returns, matching how the Windows and Linux implementations already handle `saveFile`.
- iOS is unaffected, its native handler still receives the bytes since it needs them before presenting the export picker.

## Test plan
- [x] `flutter analyze` on `file_picker_darwin`
- [x] `flutter test` on `file_picker_darwin`, including a new test that writes real bytes to a temp file on disk and verifies its contents

Fixes #2132


packages/file_picker_darwin/test/file_picker_darwin_test.dart
```diff
@@ -20,5 +23,45 @@ void main() {
       expect(file.name, equals('test.png'));
       expect(file.uri.path, equals('/tmp/test.png'));
     });
+
+    test('saveFile writes bytes to disk for real on macOS', () async {
```
navaronbracke: I did have a laugh on the test name :)

```suggestion
    test('saveFile writes bytes to disk on macOS', () async {
```

navaronbracke (approved):
LGTM with nit

PR #2138, opened 2026-08-15
ci: unify workflows and close platform coverage gaps

vicajilau:
The Android AGP compatibility matrix lived in its own workflow with no path filtering, so it ran on **every** pull request regardless of what the change touched. [#2133](https://github.com/miguelpruivo/flutter_file_picker/pull/2133), which only relaxes a dbus constraint in `file_picker_linux`, spent 8 minutes of runner time building Android APKs.

Fixing that surfaced a few more things that had drifted since the move to `packages/`.

## Workflows

- The AGP 8.x and 9.x matrix moves into `main.yml` as `build_android`, gated by the same `dorny/paths-filter` job as the other platform builds. `android-agp-compat.yml` is deleted. Gating with `if:` rather than a trigger-level `paths:` matters: a job skipped by `if:` counts as passing for branch protection, a filtered-out trigger stays pending.
- The path filters are reworked with a YAML anchor. They referenced `lib/**`, which no longer exists, and covered neither `packages/file_picker/**` nor `packages/file_picker_android/**`. A change to the facade package skipped every platform build.
- `timeout-minutes` on every job. The default is 360, and macOS minutes bill at 10x.
- Explicit `permissions` on all three workflows, and `concurrency` groups. Publishing uses a non-cancelling group so a release is never aborted halfway through, which would leave some packages published and others not.
- The stale bot announced 7 and 14 days in the messages it posts on issues while the configuration used 30 and 90.

## Coverage gaps

- **iOS was never compiled.** `file_picker_darwin` declares both `ios` and `macos` with `sharedDarwinSource: true`, but the job only ran `flutter build macos`. Now builds both, reusing the runner that already spins up.
- **R8 was never exercised.** `packages/file_picker_android/android/proguard-rules.pro` exists, but R8 only runs on release builds and CI only built debug. The `agp-9.x` lane now also builds a release APK. No secrets involved, the example signs release with the debug keys.
- **Web only built with `dart2wasm`.** JS interop does not behave identically under `dart2js`, which is still what most apps ship. Both are built now.
- **`tool/test` never ran.** `melos run test` covers `packages/*` and `example`, and `tool` is neither.

## Tooling

- The 56 lines of Python embedded in the workflow move to `tool/patch_android_example.dart`, covered by 19 tests in `tool/test`. Nothing formatted, linted or tested that heredoc, and it could not be run locally to reproduce a red lane. It is now one command:
  ```
  dart tool/patch_android_example.dart --agp 8.9.1 --gradle 8.14.3 --kotlin 2.2.20 --built-in-kotlin ""
  cd example && flutter build apk --debug
  ```
  Verified to produce byte identical output to the Python it replaces, on both lanes.
- `tool/fetch_versions.dart` retries the Flutter release feed and falls back to the `stable` and `beta` channels instead of exiting non-zero. Every build job hangs off that lookup through `needs`, so a transient network error skipped the entire pipeline and reported it green.
- Gradle is cached per lane. The cache step sits after the patch step on purpose, so the key reflects the toolchain the lane actually builds with. Keying it before would make both lanes share a key and thrash.

## Job names

`build_macos` becomes `build_darwin` now that it builds both Apple targets, and the Android lanes follow the same naming pattern as the rest:

```
Build Android (example app, agp-8.x)
Build Android (example app, agp-9.x)
Build Linux (example app)
Build Windows (example app)
Build Darwin (macOS + iOS example app)
Build Web (example app)
```

## Verification

`dart format` and `flutter analyze` clean, the 19 new tests plus the existing one pass, `dart tool/fetch_versions.dart` returns the same versions as before, and the fallback path was exercised against an unreachable host.


tool/test/patch_android_example_test.dart
```diff
@@ -0,0 +1,214 @@
...
+        throwsA(isA<PatchException>()),
+      );
+    });
+  });
+
+  group('patchGradleWrapper', () {
+    test('replaces the distribution version', () {
+      const source =
+          'distributionUrl=https\\://services.gradle.org/distributions/'
+          'gradle-9.5.0-all.zip\n';
```
navaronbracke, line 89: IIRC i do recall that upstream Flutter switched to the bin releases, instead of "all".

Something for later, @vicajilau  ?

tool/patch_android_example.dart
```diff
@@ -0,0 +1,186 @@
...
+  String toString() => message;
+}
+
+Future<void> main(List<String> args) async {
+  final Map<String, String> options;
+  try {
+    options = parseArgs(args);
+  } on FormatException catch (e) {
+    stderr.writeln('$e\n\n$usage');
+    exit(64);
```
navaronbracke: Is this 64 for a reason? Why not 1?

tool/patch_android_example.dart
```diff
@@ -0,0 +1,186 @@
...
+Map<String, String> parseArgs(List<String> args) {
+  final options = <String, String>{};
+  for (var i = 0; i < args.length; i++) {
+    final arg = args[i];
+    if (!arg.startsWith('--')) {
+      throw FormatException('Unexpected argument: $arg');
+    }
+    final separator = arg.indexOf('=');
+    if (separator != -1) {
+      options[arg.substring(2, separator)] = arg.substring(separator + 1);
```
navaronbracke: Why is this substring starting from 2?

tool/patch_android_example.dart
```diff
@@ -0,0 +1,186 @@
...
+        kotlinPattern,
+        'id("org.jetbrains.kotlin.android") version "$kotlinVersion" apply false',
+      );
+}
+
+/// Applies [gradleVersion] to a `gradle-wrapper.properties` source.
+///
+/// Throws a [PatchException] if the distribution URL is missing.
+String patchGradleWrapper(String source, {required String gradleVersion}) {
+  final pattern = RegExp(r'gradle-[^-]+-all\.zip');
```
navaronbracke: Should this check `distributionUrl=` ?

navaronbracke (approved):
LGTM with nits

tool/test/patch_android_example_test.dart
vicajilau: Checked against Flutter 3.47.0 and the template still writes `-all.zip`, as does `gradle_utils.dart` when it regenerates the wrapper. There is no `bin.zip` anywhere in `flutter_tools`.

That said, this was fragile for a different reason: Gradle's own `wrapper` task defaults to `bin`, so regenerating the wrapper without `--distribution-type=all` would have failed the lane with a message that reads like a bug in the script. It now captures the distribution type and writes it back unchanged, so either one works and only the version is rewritten. Covered by tests.

tool/patch_android_example.dart
vicajilau: Yes, 64 is what the Flutter CLI and `package:args` report for usage errors. Keeping it distinct from 1 separates the two ways this can fail: the workflow invoked the script wrong, or the example app drifted away from the shape the patches expect. Only the second one means anything is wrong with the repo.

It was a bare literal though, which is a fair thing to trip over. It is now a named `usageExitCode` constant with that reasoning next to it.

tool/patch_android_example.dart
vicajilau: To strip the leading `--`. It was a magic number, so it is now `arg.substring(optionPrefix.length)` against a named constant.

tool/patch_android_example.dart
vicajilau: Good catch, it should. As written it would rewrite any `gradle-<version>-all.zip` string anywhere in the file, including inside a comment. It is now anchored to `^distributionUrl=.*/` with `multiLine`, and there are two tests covering it: a comment mentioning an old distribution is left alone, and a file where only a comment mentions one throws instead of silently patching nothing.

tool/test/patch_android_example_test.dart
navaronbracke: I was referring to stuff that landed on flutter/flutter master lately, but something to keep track of

vicajilau: Good context!! thanks @navaronbracke!! Checked flutter/flutter master: the template still
writes -all.zip, but flutter_tools does treat bin as a first class format.
There is an explicit "recognizable 'bin' format returns correct version"
test in gradle_utils_test.dart, and the shared fixture in test/src/
android_common.dart builds its wrapper with -bin.zip.

So this was worth fixing rather than tracking. The patcher now detects the
distribution type and preserves it, which means whichever way upstream
goes, the AGP lanes keep working without a change on our side.

PR #2144, opened 2026-08-19
Fall back to the real home directory on sandboxed macOS

### LINKED ISSUE #1542: FilePicker should display the home directory of the user on MacOS in a sandbox environement
When using `String? selectedDirectory = await FilePicker.platform.getDirectoryPath();` on MacOS with the security sandbox enabled (necessary for shipping on the AppStore) the home directory is replaced by a `Data` folder located in `~/Library/Containers/MyAppName/Data/`

This is the home environment set by the OS for the container, allowing the app to access only files in this container and selected files by the user (using FilePicker)

Screenshot 2024-07-05 at 12 33 49

FilePicker is a dialog for the users that allows the user to access any file, it should not care if the app is in a sandbox environment and should display the user's home directory, not the sandbox home directory

vicajilau:
## Summary
- On macOS, when `initialDirectory` is not provided, `NSOpenPanel`/`NSSavePanel` default to the App Sandbox container's `Data` folder rather than the user's real home directory.
- This adds a sandbox entitlement check (`com.apple.security.app-sandbox`) and only when the app is sandboxed, falls back to `FileManager.default.homeDirectoryForCurrentUser` for `pickFiles`, `pickFileAndDirectoryPaths`, `getDirectoryPath`, and `saveFile`.
- Non-sandboxed apps are unaffected, they keep the panel's normal behavior of remembering the last visited directory.

Fixes #1542

## Test plan
- [ ] Manual verification with the example app on macOS with App Sandbox enabled and no `initialDirectory` passed

### DIFF
--- a/packages/file_picker_darwin/CHANGELOG.md
+++ b/packages/file_picker_darwin/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## 1.0.2
 
 - Fixed a crash (`FileSystemException`/`OSError: Is a directory`) when picking a Live Photo from the iOS gallery. Live Photos are saved by iOS as `.pvt` packages (directories), which `loadFileRepresentation(forTypeIdentifier: UTType.item...)` returned as-is; the still image contained in the package is now extracted instead.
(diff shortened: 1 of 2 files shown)


navaronbracke (approved):
LGTM

If I understand correctly, when sandboxed it is allowed to access the home directory, because only apps from the App Store are sandboxed?

Issue #1704, opened 2025-02-10
[MacOS] beginSheetModal API might call its result handler on a thread different from the main thread


nofacer:
**Describe the bug**
`await FilePicker.platform.getDirectoryPath()` doesn't pop up selection window

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop

**Platform OS version**
MacOS 15.3
Flutter 3.27.3
Dart 3.6.1
file_picker 8.3.3

**How are you picking?**
```
Positioned(
          right: 16.0,
          bottom: 16.0,
          child: Row(
            mainAxisSize: MainAxisSize.min,
            children: [
              if (_count > 0) ...[
                FloatingActionButton(
                  onPressed: () async {
                    print(11);
                    String? selectedDirectory =
                        await FilePicker.platform.getDirectoryPath();
                    print(selectedDirectory);
                  },
                  child: Icon(Icons.download),
                ),
                SizedBox(width: 12.0),
              ],
              FloatingActionButton(
                onPressed: () {
                  setState(() {
                    _count++;
                  });
                },
                child: Icon(Icons.refresh),
              ),
            ],
          ),
        ),
```

**Details to reproduce the issue**
You can use the code above

**Error Log**
No error but it just returns null without opening up selection window. 

**Flutter Version details**
```
[✓] Flutter (Channel stable, 3.27.3, on macOS 15.3 24D60 darwin-arm64, locale en-CN)
    • Flutter version 3.27.3 on channel stable at /Users/posinorth/opt/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision c519ee916e (3 weeks ago), 2025-01-21 10:32:23 -0800
    • Engine revision e672b006cb
    • Dart version 3.6.1
    • DevTools version 2.40.2
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn

[!] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/posinorth/Library/Android/sdk
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/to/macos-android-setup for more details.

[...]

PR #2149, opened 2026-08-19
Fix brief UI freeze on macOS by moving the entitlement check off the main thread

### LINKED ISSUE #2056: `FilePickerResult? result = await FilePicker.pickFiles();` freezes the app for a really short time on macOS.

vicajilau:
## Summary
- `checkEntitlement` performed synchronous Security framework calls (`SecTaskCreateFromSelf` and two `SecTaskCopyValueForEntitlement` calls) on the main thread at the very top of `handleFileSelection`, `handleFileAndDirectorySelection`, `handleDirectorySelection`, and `handleSaveFile`, before any dialog was created.
- This caused a brief but noticeable UI freeze right when `pickFiles`, `pickFileAndDirectoryPaths`, `getDirectoryPath`, or `saveFile` were invoked, most visible as a stutter in any running animation.
- The entitlement check now runs on a background queue (`DispatchQueue.global(qos: .userInitiated)`) first, then hops back to `DispatchQueue.main.async` to build and present the dialog, same as before.
- Also relevant to #1704: since the dialog creation and `beginSheetModal` completion still run entirely on the main thread (just reached via `DispatchQueue.main.async` now instead of synchronously), there is no risk of the panel's result landing on a background thread.

Fixes #2056

## Test plan
- [x] Manually verified with the example app using the exact freeze repro from the original issue thread (two spinning `CircularProgressIndicator`s around a button calling `pickFiles()`), no stutter observed when tapping the button after this change

### DIFF
--- a/packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/MacOSFilePickerHandler.swift
+++ b/packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/MacOSFilePickerHandler.swift
@@ -63,71 +63,86 @@ final class MacOSFilePickerHandler: NSObject, FlutterStreamHandler {
         }
     }
 
-    private func handleFileSelection(
-        _ call: FlutterMethodCall,
-        result: @escaping FlutterResult
+    /// Runs `checkEntitlement` on a background queue (Security framework calls are
+    /// synchronous and were causing a brief main-thread stall), then hops back to the
+    /// main queue: either delivering the entitlement error, or calling
+    /// `onEntitlementGranted` to build and present the dialog.
+    private func withEntitlementCheck(
+        requiredMode: EntitlementMode,
+        result: @escaping FlutterResult,
+        onEntitlementGranted: @escaping () -> Void
     ) {
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             guard let self else { return }
-            let entitlementError = self.checkEntitlement(requiredMode: .readOrWrite)
+            let entitlementError = self.checkEntitlement(requiredMode: requiredMode)
 
             DispatchQueue.main.async {
                 if let entitlementError {
                     result(entitlementError)
                     return
                 }
+                onEntitlementGranted()
+            }
+        }
+    }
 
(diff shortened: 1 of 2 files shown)


packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/MacOSFilePickerHandler.swift
```diff
@@ -67,57 +67,65 @@ final class MacOSFilePickerHandler: NSObject, FlutterStreamHandler {
...
-        dialog.canChooseDirectories = false
-        dialog.canChooseFiles = true
-        let extensions = args["allowedExtensions"] as? [String] ?? []
-        applyExtensions(dialog, extensions, method: call.method)
-
-        guard let appWindow = getFlutterWindow() else {
-            result(nil)
-            return
-        }
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
```
navaronbracke, line 75: I wonder if we can deduplicate this with a new function?

```
DispatchQueue.global(qos: .userInitiated).async { [weak self] in 
  guard let self else { return }

  let entitlementError = self.checkEntitlement(requiredMode: .readOrWrite)

  DispatchQueue.main.async {
    if let entitlementError {
      result(entitlementError)
      return
    }

    onEntitlementGranted(...)
  }
}
```

navaronbracke (changes requested):

packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/MacOSFilePickerHandler.swift
vicajilau: Good call, that pattern repeats 4 times almost verbatim. Extracting it into a shared helper now, will push an update shortly.

navaronbracke (approved):

PR #2150, opened 2026-08-19
Restore a working way to skip macOS App Sandbox entitlement checks (darwin-only)

### LINKED ISSUE #2143: macos: skipEntitlementsChecks()
would be nice to implements again skipEntitlementsChecks() as the checks are always done on mac and it impeach to use filepicker in a non sandboxed app. I had to make a fork, to always skip the checks...

vicajilau:
## Summary

Split into a darwin-only, non-breaking PR per review feedback. This PR no longer touches `file_picker` or removes anything.

- The native `skipEntitlementsChecks` mechanism in `MacOSFilePickerHandler` always existed and worked, but the only public Dart entry point, `FilePicker.skipEntitlementsChecks()`, was turned into a deprecated no-op during the v12 federation rewrite. The deprecation message assumed entitlement checks would be "handled automatically", but `checkEntitlement` has no way to detect whether the host app is actually sandboxed, so a non-sandboxed app has no way to opt out and gets incorrectly blocked with `ENTITLEMENT_NOT_FOUND`/`ENTITLEMENT_REQUIRED_WRITE`.
- Adds `FilePickerDarwin.skipEntitlementsChecks()` as a real, working entry point, for apps that specifically depend on `file_picker_darwin`:
  ```dart
  }
  ```
- The deprecated `FilePicker.skipEntitlementsChecks()` no-op on the root package is left untouched here. A follow-up PR against `file_picker` will make it delegate to this once `file_picker_darwin`'s version with this method is out, instead of removing it, so no breaking change is needed at all.

Fixes #2143

## Test plan
- [x] `dart analyze` passes on `file_picker_darwin`
- [x] `flutter test` passes on `file_picker_darwin`
- [ ] Manual verification on a non-sandboxed macOS app: confirm calling `FilePickerDarwin.skipEntitlementsChecks()` before `pickFiles()` avoids the entitlement error

### DIFF
--- a/packages/file_picker_darwin/lib/src/file_picker_darwin.dart
+++ b/packages/file_picker_darwin/lib/src/file_picker_darwin.dart
@@ -173,6 +173,29 @@ class FilePickerDarwin extends FilePickerPlatform {
     await methodChannel.invokeMethod<void>('clear');
   }
 
+  /// Skips the macOS App Sandbox entitlement checks performed before showing a dialog.
+  ///
+  /// Only relevant for non-sandboxed macOS apps, which do not declare the
+  /// `com.apple.security.files.user-selected.read-only`/`read-write` entitlements
+  /// and would otherwise be incorrectly blocked by those checks. Call this before
+  /// any other file picking method.
+  ///
+  /// This method does nothing on iOS.
+  ///
+  /// Note: skipping entitlement checks may lead to unexpected behavior if the
+  /// app is actually sandboxed. Use with caution.
+  ///
+  /// ```dart
+  /// if (FilePickerPlatform.instance case final FilePickerDarwin instance) {
+  ///   await instance.skipEntitlementsChecks();
+  /// }
+  /// ```
+  Future<void> skipEntitlementsChecks() async {
+    // Only macOS performs App Sandbox entitlement checks, this is a no-op on iOS.
+    if (!Platform.isMacOS) return;
+    await methodChannel.invokeMethod<void>('skipEntitlementsChecks');
+  }
+
   @override
   Future<Uri?> saveFile({
     required String fileName,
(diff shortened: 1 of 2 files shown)


packages/file_picker_darwin/lib/src/file_picker_darwin.dart
```diff
@@ -173,6 +173,27 @@ class FilePickerDarwin extends FilePickerPlatform {
     await methodChannel.invokeMethod<void>('clear');
   }
 
+  /// Skips the macOS App Sandbox entitlement checks performed before showing a dialog.
+  ///
+  /// Only relevant for non-sandboxed macOS apps, which do not declare the
+  /// `com.apple.security.files.user-selected.read-only`/`read-write` entitlements
+  /// and would otherwise be incorrectly blocked by those checks. Call this before
+  /// any other file picking method. No-op on iOS.
```
navaronbracke:
```suggestion
  /// any other file picking method.
  ///
  /// This method does nothing on iOS.
```

packages/file_picker/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ topics:
   - storage
   - desktop
   - web
-version: 12.0.1
+version: 12.1.0
```
navaronbracke, line 12: This should become `12.0.2`

packages/file_picker/lib/src/file_picker.dart
```diff
@@ -243,10 +243,4 @@ abstract final class FilePicker {
       webOptions: webOptions,
     );
   }
-
-  /// Deprecated entitlement check helper for legacy compatibility.
-  @Deprecated(
```
navaronbracke, line 248: I would avoid the breaking change, by doing the following:

1) Release file_picker_darwin 1.0.2
2) Depend on it in a new PR for file_picker
3) Remove the Deprecated annotation here (and copy over the docs)
4) Provide the following implementation in file_picker

```
  }
}
```

navaronbracke (changes requested):
This needs to be broken up into two PR's, to avoid the breaking change

packages/file_picker_darwin/lib/src/file_picker_darwin.dart
vicajilau: Applying this suggestion.

packages/file_picker/pubspec.yaml
vicajilau: This is going away entirely: splitting this into two PRs as you suggested, so file_picker's own version won't need a breaking bump at all here.

packages/file_picker/lib/src/file_picker.dart
vicajilau: Agreed, this is a better approach, no breaking change needed. Splitting: this PR will stay darwin-only (adding FilePickerDarwin.skipEntitlementsChecks()), and I'll open a separate PR for file_picker once file_picker_darwin's 1.0.2 is out, implementing skipEntitlementsChecks() exactly as you suggested instead of removing it.

navaronbracke (approved):

PR #2153, opened 2026-08-19
iOS picks LivePhotos as .pvt

test0terter0n:
Currently Live Photos are picked as .pvt. Before update 12.0.0, they were picked as jpeg or .heic. 
Now it's always a jpeg.

### DIFF
--- a/packages/file_picker/CHANGELOG.md
+++ b/packages/file_picker/CHANGELOG.md
@@ -1,3 +1,10 @@
+
+
+## 12.0.1
+
+### iOS
+- Fixed Live Photos being picked as `.pvt` bundles. Live Photos are now converted to JPEG, while all other image formats (JPEG, PNG, HEIC, ...) are loaded in their original format instead of being converted to JPEG. Thank you @test0terter0n!
+
 ## 12.0.0
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


vicajilau (changes requested):
Thanks for tracking down the .pvt bug on Live Photos. The root cause (forTypeIdentifier: .item resolving to the private Live Photo container) looks right.

A few concerns before merging:

1. This forces JPEG for every image pick, not just Live Photos. PNG (with transparency), HEIC and GIF assets picked through FileType.image will now be transcoded to JPEG even when the user never touched a Live Photo. That is a behavior change worth calling out explicitly, since it can silently drop alpha channels and animation.

2. The fix only covers the "image" branch. The "media" branch (mixed image and video filter, the default case) still uses UTType.item.identifier, so a Live Photo picked through FileType.media would still resolve to .pvt.

3. Could the Live Photo case be detected specifically, for example via itemProvider.hasItemConformingToTypeIdentifier(UTType.livePhoto.identifier), and only that case forced to a still image representation? That would fix the reported bug without changing the output format for every other image pick.

4. No CHANGELOG entry for what is a visible behavior change on iOS.

Requesting changes so the fix can be narrowed to Live Photos and extended to cover the media picker path too.

PR #2158, opened 2026-08-21
fix(ios): extract still image from Live Photo .pvt package instead of crashing

### LINKED ISSUE #2157: iOS: picking a Live Photo crashes with FileSystemException ("Is a directory", errno 21)
### Description

Picking a Live Photo from the gallery on iOS (`FilePicker.pickFile(type: FileType.image)` / `pickFiles`) returns a `PlatformFile` whose `path` points to a **directory** instead of a regular file. Any caller that reads the result as a file (e.g. `file.readAsBytes()`, or `File(path).readAsBytesSync()`) crashes:

```
FileSystemException: Cannot open file, path = '.../tmp/IMG_9509.pvt' (OS Error: Is a directory, errno = 21)
```

### Root cause

iOS saves Live Photos picked via `PHPickerViewController` as `.pvt` packages — which are directories containing the still image, the paired `.mov`, and metadata — not regular files.

In `IOSFilePickerHandler.swift`, `picker(_:didFinishPicking:)` loads the item via:

```swift
item.itemProvider.loadFileRepresentation(forTypeIdentifier: UTType.item.identifier) { url, _ in ... }
```

and `copyToTemporaryDirectory` copies whatever `url` resolves to verbatim — for a Live Photo, that's the whole `.pvt` package — then returns that path as the picked file's path. Nothing downstream detects that the copied path is a directory before handing it back to Dart.

### Reproduction

1. On a physical iOS device (simulators don't have Live Photos), take a photo with Live Photo enabled (default camera behavior on most iPhones).
2. Call `FilePicker.pickFile(type: FileType.image)` (or `pickFiles`) and select that Live Photo.
3. Read the returned `PlatformFile` as bytes (`file.readAsBytes()`), or inspect `file.path` — it points to a `.pvt` directory, not a file.

### Expected behavior

The picker should return a readable image file, not a directory — either the still JPEG/HEIC contained in the `.pvt` package, or some other well-defined behavior that doesn't crash callers who trust `PlatformFile.path`/`readAsBytes()` to point at a regular file.

### Environment

- file_picker: 12.0.0
- file_picker_darwin: 1.0.1
- Flutter: 3.47.0
- iOS: 26.5.2 (also reproduced on 18.5 per user reports), physical device (not simulator)

### Related issues

This is distinct from #835 and #1747, which are about picking failing outright or recovering the paired `.MOV` file. Here the pick *succeeds* but silently returns an unreadable directory.

I've opened a PR with a fix: it detects when the copied item is a directory and extracts the still image from inside the package instead of returning the package path.

TheoGermain:
Fixes #2157

### Problem

Picking a Live Photo from the gallery on iOS crashes callers who read the picked file as bytes, with:

```
FileSystemException: Cannot open file, path = '.../tmp/IMG_9509.pvt' (OS Error: Is a directory, errno = 21)
```

iOS saves Live Photos picked via `PHPickerViewController` as `.pvt` packages, which are **directories** (containing the still image, the paired `.mov`, and metadata), not regular files. `copyToTemporaryDirectory` in `IOSFilePickerHandler.swift` copies whatever `loadFileRepresentation` resolves to verbatim and hands that path back to Dart as the picked file — for a Live Photo, that's the whole package directory.

### Fix

Added `resolveActualFile(at:)`, called from `copyToTemporaryDirectory` right after the copy: if the copied path is a directory, it looks inside for the first file with an image extension (`jpg`/`jpeg`/`heic`/`heif`/`png`) and returns that instead of the package path. Regular (non-package) files are returned unchanged, so this only affects the Live Photo case.

### Why this approach

- Minimal, localized to the one function that already owns "resolve the picked item into a file we hand back to Dart" — no changes to the Dart-facing API or other platforms.
- Doesn't depend on a specific Live Photo UTType (which can vary across iOS versions) — it just handles "the picker gave us a directory" generically, which also protects against any other package-like result iOS might return.

### Testing

Verified the logic against the reported crash conditions (a `.pvt` package containing a JPEG + MOV). This package doesn't have an existing Swift/native test harness (per CONTRIBUTING.md, "Java/Objective-C code is currently not tested"), so I didn't add one — happy to add a target if you'd like one for this.

### Changes

- `packages/file_picker_darwin`: fix + patch bump to 1.0.2 + changelog entry.
- `packages/file_picker`: patch bump to 12.0.1 + changelog entry (per CONTRIBUTING.md guidance to bump the version and changelog for a bugfix PR).

### DIFF
--- a/packages/file_picker/CHANGELOG.md
+++ b/packages/file_picker/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 12.0.1
+### iOS
+- Fixed a crash (`FileSystemException`/`OSError: Is a directory`) when picking a Live Photo from the gallery. Live Photos are saved by iOS as `.pvt` packages (directories); the still image contained in the package is now extracted instead of returning the package itself.
+
 ## 12.0.0
 
 ### Breaking Changes
--- a/packages/file_picker/pubspec.yaml
+++ b/packages/file_picker/pubspec.yaml
@@ -9,7 +9,7 @@ topics:
   - storage
   - desktop
   - web
-version: 12.0.0
+version: 12.0.1
 
 resolution: workspace
 
--- a/packages/file_picker_darwin/CHANGELOG.md
+++ b/packages/file_picker_darwin/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.2
+
+- Fixed a crash (`FileSystemException`/`OSError: Is a directory`) when picking a Live Photo from the iOS gallery. Live Photos are saved by iOS as `.pvt` packages (directories), which `loadFileRepresentation(forTypeIdentifier: UTType.item...)` returned as-is; the still image contained in the package is now extracted instead.
+
 ## 1.0.1
 
 - Fixed `pickFileAndDirectoryPaths` never returning directories on iOS and macOS. It now calls the native combined file and directory picker instead of silently falling back to file only selection.
--- a/packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/IOSFilePickerHandler.swift
+++ b/packages/file_picker_darwin/darwin/file_picker_darwin/Sources/file_picker_darwin/IOSFilePickerHandler.swift
@@ -360,10 +360,36 @@ final class IOSFilePickerHandler: NSObject,
                 try FileManager.default.removeItem(at: destinationURL)
             }
             try FileManager.default.copyItem(at: sourceURL, to: destinationURL)
-            return destinationURL
+            return resolveActualFile(at: destinationURL)
         } catch {
             return nil
         }
+    }
+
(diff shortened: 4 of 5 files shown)


vicajilau (approved):

vicajilau (approved):
Clean, well-scoped fix. Handling the "loadFileRepresentation resolved to a directory" case generically at the copy step, instead of forcing a specific UTType per picker mode, is the right call here: it also covers the `media` (mixed image/video) picker, not just `image`, and it doesn't change the format of any photo that isn't actually a Live Photo package.

Verified `resolveActualFile` builds and the logic matches the reported `.pvt` structure. Closed #2153 in favor of this one, it addressed the same root cause but forced JPEG on every image pick and left the `media` branch unfixed.

Issue #2160, opened 2026-08-22
[macOS] saveFile returns a scheme-less Uri, unlike Windows and Linux


JacktheRanger:
Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened issues — checked, on 12.0.0, none found for this.

**Describe the bug**

`FilePicker.saveFile()` returns `Uri?`, documented as:

> Returns the [Uri] of the saved file, or `null` if the user canceled the operation.
> Depending on the platform, the [Uri.scheme] may be `file`, `content`, `http(s)`, `data` or `blob`.

On macOS the returned Uri has **no scheme at all**, which is not one of the documented possibilities and differs from the other desktop platforms.

In `file_picker_darwin/lib/src/file_picker_darwin.dart`:

```dart
if (savedPath == null) return null;
return Uri.tryParse(savedPath) ?? Uri.file(savedPath);
```

`savedPath` is the native `url.path` (`MacOSFilePickerHandler.swift` → `handleSaveFile` returns `result(url.path)`), i.e. a bare POSIX path like `/Users/me/My Reports/Report.txt`. `Uri.tryParse` succeeds on that string and returns a **relative, scheme-less** Uri, so the `Uri.file(savedPath)` fallback never runs.

For comparison, both other desktop implementations return a proper `file://` Uri:

- `windows_file_picker/lib/src/file_picker_windows.dart` → `return Uri.file(savedFilePath);`
- `file_picker_linux/lib/src/file_picker_linux.dart` → `return Uri.file(savedFilePath);`

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop (macOS)

**Platform OS version**

macOS 27.0 (arm64). Applies by inspection to any macOS version.

**How are you picking?**

```dart
final uri = await FilePicker.saveFile(
  fileName: 'Report.txt',
  bytes: utf8.encode('hello'),
  mimeType: 'text/plain',
);

// The natural way to show the result to a user:
final shown = uri!.isScheme('file') ? uri.toFilePath() : uri.toString();
print(shown);
```

**Details to reproduce the issue**

1. Call `saveFile()` on macOS and save into a directory whose path contains a space or non-ASCII characters.
2. Print the result as above.

On Windows/Linux this prints a clean path. On macOS `isScheme('file')` is `false`, so it falls through to `toString()` and the path comes out percent-encoded:

```
[...]

Issue #2159, opened 2026-08-22
[macOS] saveFile does not enforce the default extension (allowedContentTypes never set on NSSavePanel)


JacktheRanger:
Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened issues — checked, on 12.0.0, and no open macOS equivalent. This is the macOS counterpart of #2139, which was fixed for Windows in #2140.

**Describe the bug**

On macOS, `saveFile()` never sets `allowedContentTypes` on the `NSSavePanel`. When the user replaces the suggested `fileName` with one that has **no extension**, the file is written without any extension — even though `type: FileType.custom` and `allowedExtensions` were provided at the call site.

This is the same behaviour reported for Windows in #2139, with the same root cause that #2140 already identified: the `saveFile()` API never passes the extension list down to the platform layer. #2140 fixed the Windows side; macOS still has it.

It is also a regression from v11, where `FilePicker.saveFile()` forwarded `type`/`allowedExtensions` through the platform interface and macOS applied them to the panel.

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop (macOS)

**Platform OS version**

macOS 27.0 (arm64). Also applies by inspection to any macOS version — the code path never assigns `allowedContentTypes` regardless of OS version.

**How are you picking?**

```dart
final uri = await FilePicker.saveFile(
  dialogTitle: 'Save Report',
  fileName: 'Report.txt',
  bytes: utf8.encode('hello'),
  mimeType: 'text/plain',
  type: FileType.custom,
  allowedExtensions: ['txt'],
);
```

**Details to reproduce the issue**

1. Call `saveFile()` as above on macOS.
2. In the save panel, clear the suggested name and type `Report` (no extension).
3. Save.

Result: a file named `Report`, with no extension. Double-clicking it does not open in a text editor.
Expected (v11, and Windows after #2140): `Report.txt`.

**Where it breaks**

The extension list is dropped at three points before it can reach the panel:

1. `file_picker/lib/src/file_picker.dart` — the facade's `saveFile()` accepts `type` and `allowedExtensions`, but does not include them in the `FilePickerPlatform.instance.saveFile(...)` call.
2. `file_picker_platform_interface/lib/src/file_picker_platform_interface.dart` — `saveFile()` declares no `type`/`allowedExtensions` parameters at all, so the facade has nowhere to forward them. (`pickFiles()` in the same file declares both.)
[...]

Issue #2139, opened 2026-08-17
[Windows] saveFile does not append the default extension (lpstrDefExt never set on OPENFILENAMEW)


palmoni5:
Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened issues — checked, none found for this.

**Describe the bug**
On Windows, `saveFile()` never sets `lpstrDefExt` on the `OPENFILENAMEW` struct before calling `GetSaveFileNameW`. As a result, when the user types a file name **without an extension** in the save dialog (replacing the suggested `fileName`), Windows does not append the default extension, and the file is written without any extension — even though `type: FileType.custom` and `allowedExtensions` were provided.

The saved file is then unrecognized by the OS (double-clicking it prompts "How do you want to open this file?"). Manually renaming the file to add the extension fixes it, which is what our users end up doing (reported downstream in our app: Otzaria/otzaria#817).

The `lpstrDefExt` field already exists in the FFI struct definition (`file_picker_windows_ffi_types.dart`), it is just never assigned in `_instantiateOpenFileNameW`. Setting it to the first allowed extension would make `GetSaveFileNameW` append the extension automatically, which is the standard Win32 behavior (see [OPENFILENAMEW docs](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew), `lpstrDefExt`).

Verified present in both `12.0.0-beta.7` (hosted on pub.dev) and current `master` (`packages/file_picker_windows/lib/src/file_picker_windows.dart`, `_instantiateOpenFileNameW`).

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop (Windows)

**Platform OS version**
Windows 11 (any Windows version — this is Win32 `GetSaveFileNameW` behavior).

**How are you picking?**
```dart
final path = await FilePicker.saveFile(
  dialogTitle: 'Export file',
  fileName: 'calendar.pdf',
  type: FileType.custom,
  allowedExtensions: ['pdf'],
  bytes: bytes,
  lockParentWindow: true,
);
```

**Details to reproduce the issue**
1. Call `saveFile` as above on Windows.
2. In the save dialog, type a new file name without an extension (e.g. `my calendar`) instead of keeping the suggested one.
3. Save.
[...]

PR #2162, opened 2026-08-23
fix(darwin): enforce default extension and fix Uri scheme in macOS saveFile

vicajilau:
## Problem

Two related bugs in `saveFile()` on macOS.

### #2159: extension not enforced

`allowedExtensions` is not forwarded from the `saveFile()` API down to the native save dialog at all (`file_picker`, `file_picker_platform_interface`, and `file_picker_darwin` all drop it before it reaches `MacOSFilePickerHandler.swift`). Because of that, `applyExtensions` always falls into its empty branch for the `"save"` method, so `NSSavePanel.allowedContentTypes` is never set. If the user clears the extension from the suggested file name, the file gets saved with none.

This is the macOS counterpart of #2139, which was fixed for Windows in #2140.

### #2160: scheme-less Uri

`FilePickerDarwin.saveFile()` returned `Uri.tryParse(savedPath) ?? Uri.file(savedPath)`. `savedPath` is always a native POSIX path (`url.path` from `NSSavePanel`/`UIDocumentPickerViewController`), never a URI string, so `Uri.tryParse` always succeeds and returns a relative, scheme-less Uri, and the `Uri.file` fallback never runs. Any path containing a space or non-ASCII character comes out percent-encoded and `Uri.scheme` doesn't match any of the documented values, unlike Windows and Linux which already return `Uri.file(...)`.

## Fix

- `handleSaveFile` now falls back to the extension of the suggested `fileName` when no `allowedExtensions` reached it, on both the `UTType` (macOS 11+) and legacy (`allowedFileTypes`) paths, mirroring the `lpstrDefExt` approach already used for Windows in #2140. Full plumbing of `allowedExtensions` through the public API remains a separate, larger change.
- `FilePickerDarwin.saveFile()` now always returns `Uri.file(savedPath)`, matching Windows and Linux.

## Test plan
- [x] `dart analyze` passes on the modified file
- [ ] Manual verification with the example app on macOS: `saveFile(fileName: 'Report.txt', ...)`, clear the extension in the panel, save, and confirm it is written back as `.txt`
- [ ] Manual verification that the returned `Uri` has `scheme == 'file'` and `toFilePath()` matches the saved path for a directory containing a space


navaronbracke (approved):
LGTM, if I can assume iOS is not affected?

Issue #2161, opened 2026-08-22
When doing breaking changes, please remember to log what is removed in favor of which alternative feature


milindgoel15:
I encourage you to add new suggestions that you think that can make file_picker even a better package.

_**Note:**_: Before creating a new suggestion, make sure it isn't already listed. Filter by [suggestion] tag.

Before the recent updates, there used to be a class FilePickerResult now removed entirely and nothing mentioned in changes on what to use instead. 

Please document stuff which is being removed and what is introduced in place of it.

milindgoel15:
there used to be a .extension method available but is there any alternative available now??

milindgoel15:
migration steps in readme but would have better to include a note within changelog too.

https://pub.dev/packages/file_picker#migrating-to-v12

vicajilau:
Good point, thanks for flagging it. Opened #2163: it adds the missing migration note for `FilePickerResult` to the `12.0.0` changelog entry, and adds a rule to `CONTRIBUTING.md` requiring future breaking-change entries to state what to use instead, not just what was removed.

milindgoel15:
> Good point, thanks for flagging it. Opened #2163: it adds the missing migration note for `FilePickerResult` to the `12.0.0` changelog entry, and adds a rule to `CONTRIBUTING.md` requiring future breaking-change entries to state what to use instead, not just what was removed.
> 

Also, what is the alternative to the .extension method now? Moving to new api system should be seemless for all projects using it or atleast having alternatives available already in place while keeping the option being removed as deprecated.

Having a system marked as deprecated and not to be used is better than removing it altogether. Then with another major/minor release could be removed.

vicajilau:
`PlatformFile.extension` was not removed on purpose. The v12 release was a large refactor that federated `file_picker` into separate platform packages behind a shared interface, and this getter was simply missed while rebuilding `PlatformFile` from scratch on that new interface, it fell out by accident, not by a deliberate decision to drop it.

There is already a PR open to bring it back: #2164. It restores the getter (and fixes a bug the old one had, where a file with no extension returned its own full name instead of `null`), so once that lands you won't need a workaround at all.

Until it's released, or if you're on an older `file_picker_platform_interface`, you can derive it from `name` yourself with `package:path` (already pulled in transitively through `file_picker`, so it just needs `path: ^1.9.0` added to your own `pubspec.yaml` to import it directly):

```dart
import 'package:path/path.dart' as p;

[...]

PR #2164, opened 2026-08-23
feat(platform_interface): add PlatformFile.extension

vicajilau:
## Why

`PlatformFile.extension` existed for years before the v12 federated rewrite and was dropped without a replacement. #2161 and its comments (from milindgoel15) flagged this, asking for either the getter back or at least the migration path documented.

## What

- Added `PlatformFile.extension` back to `file_picker_platform_interface`, as a concrete getter on the base class (no platform-specific implementation needs to change).
- Delegates to `package:path`'s `extension()` for the actual parsing (already a dependency of this package), so a dotfile like `.gitignore` correctly returns `null` instead of treating the whole name as an extension, which was a real bug in the pre-12.0 getter (`name.split('.').last` returned `'README'` for a file with no extension at all).
- Strips the leading dot `package:path` includes, to match `name`'s own convention and `allowedExtensions` elsewhere in this API, which are both dot-less.
- Works unmodified on every platform, including web: it only reads `name`, never `path`/`uri`, so it is unaffected by web's `blob:`/`data:` URIs.

## Changes

- `file_picker_platform_interface`: 3.0.1 → 3.1.0 (minor, pure addition).
- `file_picker`: 12.0.1 → 12.1.0, folded together with the already-merged #2158 entry since 12.0.1 was never published. Dependency constraint bumped to `file_picker_platform_interface: ^3.1.0` so the getter is guaranteed available once this is picked up.

## Test plan
- [x] `dart analyze` passes on `file_picker_platform_interface` and `file_picker`
- [x] `flutter test` passes on both packages
- [x] New unit tests cover: normal extension, multi-dot (`archive.tar.gz` → `gz`), no extension (`README` → `null`), dotfile with no other dot (`.gitignore` → `null`), dotfile with a real extension (`.env.local` → `local`)

### DIFF
--- a/packages/file_picker/pubspec.yaml
+++ b/packages/file_picker/pubspec.yaml
@@ -9,7 +9,7 @@ topics:
   - storage
   - desktop
   - web
-version: 12.0.1
+version: 12.1.0
 
 resolution: workspace
 
@@ -21,7 +21,7 @@ dependencies:
   flutter:
     sdk: flutter
 
-  file_picker_platform_interface: ^3.0.0
+  file_picker_platform_interface: ^3.1.0
   android_file_picker: ^1.0.0
   file_picker_darwin: ^1.0.0
   file_picker_linux: ^1.0.0
--- a/packages/file_picker_platform_interface/test/platform_file_test.dart
+++ b/packages/file_picker_platform_interface/test/platform_file_test.dart
@@ -1,0 +1,51 @@
+import 'dart:typed_data';
+
+import 'package:cross_file/cross_file.dart';
+import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+base class _TestPlatformFile extends PlatformFile {
+  _TestPlatformFile(this.name);
+
+  @override
+  final String name;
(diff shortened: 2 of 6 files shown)


packages/file_picker_platform_interface/test/platform_file_test.dart
```diff
@@ -0,0 +1,51 @@
+import 'dart:typed_data';
+
+import 'package:cross_file/cross_file.dart';
+import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+base class _TestPlatformFile extends PlatformFile {
```
navaronbracke, line 7:
```suggestion
class _TestPlatformFile extends PlatformFile {
```

packages/file_picker/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ topics:
   - storage
   - desktop
   - web
-version: 12.0.1
+version: 12.1.0
```
navaronbracke, line 12: Should this be a patch version instead?

navaronbracke (approved):

PR #2166, opened 2026-08-24
fix(file_picker): make skipEntitlementsChecks delegate to FilePickerDarwin

vicajilau:
## Why

Follow-up to #2150, split out per review feedback to avoid a breaking change on `file_picker`.

`FilePicker.skipEntitlementsChecks()` was turned into a permanent no-op during the v12 federation, with `FilePickerDarwin.skipEntitlementsChecks()` (added in #2150) as the only real, working entry point, reachable only by apps that depend on `file_picker_darwin` directly.

## What

Instead of removing the deprecated no-op (which would be a breaking change), it now delegates to the real implementation:

```dart
  }
}
```

No breaking change, no major/minor bump needed beyond what #2150 already carries, this rides along as a bug fix bullet in the same unpublished `12.1.0`.

Verified that importing `file_picker_darwin` directly from `file_picker.dart` does not break Web/Wasm builds (`flutter build web` and `flutter build web --wasm` both succeed), since `file_picker_darwin`'s Dart-level `dart:io` usage is tree-shaken away when never reached on that platform.

## Depends on

Based on #2150's branch since it needs `FilePickerDarwin.skipEntitlementsChecks()` to exist. Should not be merged before #2150.

## Test plan
- [x] `dart analyze` passes on `file_picker`
- [x] `flutter test` passes on `file_picker`
- [x] `flutter build web` and `flutter build web --wasm` succeed with the example app

### DIFF
--- a/packages/file_picker/lib/src/file_picker.dart
+++ b/packages/file_picker/lib/src/file_picker.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:android_file_picker/android_file_picker.dart';
+import 'package:file_picker_darwin/file_picker_darwin.dart';
 import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
 import 'package:flutter/foundation.dart';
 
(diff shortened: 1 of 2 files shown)


packages/file_picker/lib/src/file_picker.dart
```diff
@@ -244,9 +245,20 @@ abstract final class FilePicker {
...
-  )
-  static Future<void> skipEntitlementsChecks() async {}
+  /// Skips the macOS App Sandbox entitlement checks performed before showing a dialog.
+  ///
+  /// Only relevant for non-sandboxed macOS apps, which do not declare the
+  /// `com.apple.security.files.user-selected.read-only`/`read-write` entitlements
+  /// and would otherwise be incorrectly blocked by those checks. Call this before
+  /// any other file picking method.
+  ///
+  /// This method does nothing on iOS, and on every platform other than macOS.
```
navaronbracke:
```suggestion
  /// This method does nothing on platforms other than macOS.
```

navaronbracke (changes requested):
LGTM, but let's wait until file_picker_darwin is published

packages/file_picker/lib/src/file_picker.dart
vicajilau: Applied. Agreed on waiting for file_picker_darwin's publish before merging this one.

navaronbracke (approved):

PR #2169, opened 2026-08-25
fix(ci): trigger pub.dev publishing from tag pushes instead of workflow_dispatch

vicajilau:
## Why

The last real publish attempt (after fixing OIDC auth in #2168) got past authentication and was rejected by pub.dev itself:

```
ERROR: Insufficient permissions to the resource at the https://pub.dev package repository.
The calling GitHub Action is not allowed to publish, because: publishing is not allowed from 'workflow_dispath' events.
```

Per https://dart.dev/go/publishing-from-github, this is not a config issue, it is how pub.dev's Automated Publishing works everywhere: it only accepts a publish request from a run triggered by pushing a git tag matching that package's configured tag pattern, and rejects `workflow_dispatch`-triggered runs, with no opt-out.

## What

Splits `publish.yml` into two jobs:

- **`prepare`** (`workflow_dispatch`, same `dry_run` input as before): compares each package's local version against what's on pub.dev and, unless `dry_run`, tags and pushes each unpublished one individually. Never talks to pub.dev.
- **`publish`** (triggered by the resulting `push` of a tag matching `*-v[0-9]+.[0-9]+.[0-9]+*`): the only event pub.dev accepts. Uses `bluefireteam/melos-action`'s own `publish: true` mode, which extracts the package name from the tag and publishes just that one package via OIDC (this is also what wires up `dart-lang/setup-dart`, since Flutter's bundled Dart doesn't support automated publishing).

Tags are pushed one at a time, not batched in a single `git push --tags`, since GitHub does not deliver push events for tags when more than three are pushed together. A release touching several packages produces several tags and several independent, parallel publish runs.

Also fixes a `CONTRIBUTING.md` inaccuracy: it described `melos publish` as auto-tagging on every real publish. Checked melos's own source, `--git-tag-version` defaults to off, so that was never actually happening.

## Also needed on pub.dev (not part of this PR)

Each package's Automated Publishing config on pub.dev needs a `tag_pattern` set to `<package_name>-v{{version}}` (e.g. `file_picker_darwin-v{{version}}`), matching the tag format `prepare` creates.

## Test plan
- [x] `actionlint` passes on the workflow
- [x] Tag glob pattern verified against this repo's own history: the pre-#2151 workflow used the same `v[0-9]+.[0-9]+.[0-9]+*` syntax and it successfully triggered real releases (v11.0.2, the 12.0.0 betas, v12.0.0)
- [ ] Run `prepare` with `dry_run: false`, confirm each unpublished package gets tagged and pushed one at a time, and each tag's `publish` run succeeds

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,33 +1,49 @@
 name: Publish to pub.dev
 
+# pub.dev's Automated Publishing (OIDC-based, no stored credentials) only
+# accepts publish requests from a GitHub Actions run that was triggered by a
+# `push` of a git tag matching that package's configured tag pattern. It
+# rejects any run triggered by `workflow_dispatch`, with no way to opt out of
+# that restriction: https://dart.dev/go/publishing-from-github
+#
+# So this workflow is split into two jobs:
+#   - `prepare` (workflow_dispatch): figures out which packages have an
+#     unpublished local version and, unless dry_run, tags and pushes each one
+#     individually. It never talks to pub.dev itself.
+#   - `publish` (push of a matching tag): the only event pub.dev will accept.
+#     Publishes the single package named in the tag via OIDC.
+#
+# Each pushed tag triggers its own separate `publish` run. GitHub does not
+# deliver push events for tags when more than three are pushed in the same
+# `git push`, so `prepare` pushes tags one at a time rather than batching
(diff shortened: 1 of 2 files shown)


navaronbracke (approved):
LGTM

Is publishing done entirely via pressing buttons in Github Actions?
What is the part where you change the dry run flag? Is that a different GH Action?

PR #2172, opened 2026-08-26
fix(ci): use a real token for tag pushes and stop tagging the example app

vicajilau:
## Why

First real (non-dry-run) run of the tag-based publish workflow from #2169 exposed two bugs.

### 1. Tag pushes never trigger the `publish` job

`prepare` pushed `file_picker-v12.1.1` successfully, but no `push`-triggered run ever started. GitHub does not run other workflows off a push made with the default `GITHUB_TOKEN`, this is a deliberate anti-recursion guard, not a bug on our side, so the whole tag-push trigger this design depends on silently never fired.

Confirmed by manually deleting and re-pushing the same tag with a personal token: that one did trigger `publish`, and it published `file_picker` 12.1.1 successfully. https://pub.dev/packages/file_picker/versions/12.1.1 is live.

### 2. The example app got tagged too

`melos list --json` includes every workspace package, including `example` (`publish_to: none`), which has no pub.dev entry. The previous version comparison (querying pub.dev directly) read that as "never published" and pushed `file_picker_example-v0.0.0` alongside the real release tag. Already deleted manually from the remote.

## What

- `actions/checkout` in `prepare` now authenticates with a new `RELEASE_PAT` repository secret (personal access token, `repo` scope or fine-grained "Contents: Read and write" on this repo) instead of the default token, so pushed tags are attributed to a real account and do trigger `publish`.
- Tagging now uses `melos list --no-private --no-published --json` instead of a hand-rolled pub.dev API comparison. `--no-private` drops `publish_to: none` packages (the example app), `--no-published` keeps only packages whose local version isn't published yet, using melos's own check instead of ours.
- `CONTRIBUTING.md` documents the new secret requirement.

## Needs from a repo admin

Add a `RELEASE_PAT` repository secret before the next real run, or `prepare` will fail on the checkout step (no fallback, this is intentional so a missing token fails loudly instead of quietly reverting to the bot token).

## Test plan
- [x] `actionlint` passes
- [x] `melos list --no-private --no-published --json` verified locally to return only real publishable packages, example excluded
- [x] Confirmed empirically: a tag pushed with a real token (not the bot token) does trigger the `publish` job, which successfully published `file_picker` 12.1.1


navaronbracke (approved):
LGTM

If i were to trigger the workflow, I need to setup a PAT for myself?

PR #2174, opened 2026-08-26
feat(windows): migrate pickFile/pickFiles/saveFile to IFileOpenDialog/IFileSaveDialog

### LINKED ISSUE #2173: [Windows] Migrate pickFile/pickFiles/saveFile to IFileOpenDialog/IFileSaveDialog
## Background

Surfaced while scoping #1861 (`confirmButtonText`). `pickFile()`/`pickFiles()` and `saveFile()` on Windows still use the legacy Common Dialog Box Library functions, `GetOpenFileNameW` and `GetSaveFileNameW` (`file_picker_windows.dart`), while `getDirectoryPath()` already uses the modern `IFileOpenDialog` COM interface (`FileOpenDialog`, same file).

Microsoft's own documentation for `GetOpenFileNameW` is explicit about this:

> Starting with Windows Vista, the Open and Save As common dialog boxes have been superseded by the Common Item Dialog. We recommended that you use the Common Item Dialog API instead of these dialog boxes from the Common Dialog Box Library.

https://learn.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-getopenfilenamew

## Why this matters beyond #1861

The legacy API is why #1861 can't be fully implemented on Windows, custom button text (`SetOkButtonLabel`) only exists on `IFileOpenDialog`/`IFileSaveDialog`. But that is just one symptom. The legacy dialogs are also missing everything else the modern Common Item Dialog offers: `IFileDialogEvents` callbacks, richer filter specs, custom places, and better long-path handling. Staying on `GetOpenFileNameW`/`GetSaveFileNameW` means every future Windows-specific feature request hits the same wall.

## Proposal

Migrate `pickFile()`, `pickFiles()`, and `saveFile()` to `IFileOpenDialog`/`IFileSaveDialog`, matching the pattern `getDirectoryPath()` already uses successfully in this codebase. That existing implementation is a working reference for the COM interop (interface querying, apartment threading, cleanup) this migration needs.

This is scoped as its own PR, separate from #1861. It touches the most-used entry points in the Windows implementation, so it deserves to be tested on its own rather than bundled with a smaller feature.

Related: #1861

vicajilau:
## Summary

Migrates `pickFile()`, `pickFiles()`, and `saveFile()` on Windows from the legacy `GetOpenFileNameW`/`GetSaveFileNameW` common dialogs to the modern `IFileOpenDialog`/`IFileSaveDialog` COM APIs, the same pattern `getDirectoryPath()` already used. Microsoft's own docs recommend this migration, the legacy Common Dialog Box Library has been superseded by the Common Item Dialog since Windows Vista.

This also unblocks future Windows-specific dialog features (custom button text via `SetOkButtonLabel`, richer filters, `IFileDialogEvents`) that the legacy API can't support.

Fixes #2173

## Behavior

- Multi and single file selection, filters per `FileType`, custom extensions, initial directory, default file name, default extension, overwrite confirmation, and parent window locking are all preserved.
- Filter strings are now built as `IFileDialog` filter specs (name + pattern pairs) instead of the legacy null-terminated string format.
- Removed the internal Win32 helpers tied to the legacy dialogs (`fileTypeToFileFilter`, `extractSelectedFilesFromOpenFileNameW`, `OPENFILENAMEW`, `GetOpenFileNameW`, `GetSaveFileNameW`, `ofn*` constants). These were implementation details, not part of the supported plugin API.

## Fixes from manual testing

Two issues surfaced while testing on a real Windows machine and were fixed here:

1. **`initialDirectory` empty string forced the dialog open on "This PC"** instead of falling back to Explorer's normal last-used-folder default. `""` was being treated the same as a real path; now guarded with `initialDirectory.isNotEmpty`.
2. **`lockParentWindow`/`parentWindowHandle` were silently ignored** by every caller using the public `WindowsOptions` API (including the example app), because the code only read those fields when `windowsOptions` was the internal, non-exported `FilePickerWindowsOptions` subclass. This predates this PR but affects the methods touched here, so it's fixed alongside them.

See the test checklist comment below for full manual test coverage.

## Test plan

- [x] `flutter analyze` clean
- [x] `dart format` clean
- [x] `flutter test` passes (added coverage for the new filter-spec builder)
- [x] **Manual testing on a real Windows machine** (single pick, multi pick, save with overwrite prompt, filters, cancel, initial directory, dialog title, lock parent window) — all checks pass, see comments for details


packages/file_picker_windows/lib/src/file_picker_windows.dart
```diff
@@ -289,193 +255,159 @@ class FilePickerWindows extends FilePickerPlatform {
     return null;
   }
 
-  /// Opens the Win32 file picker dialog using [GetOpenFileNameW].
+  /// Opens the file-open dialog ([IFileOpenDialog]) to select one or more files.
```
navaronbracke:
```suggestion
  /// Opens the [IFileOpenDialog] to select one or more files.
```

packages/file_picker_windows/lib/src/file_picker_windows.dart
```diff
@@ -492,32 +424,112 @@ class FilePickerWindows extends FilePickerPlatform {
...
+      throw ArgumentError(
+        'If type is set to FileType.custom, allowedExtensions cannot be null or empty.',
+      );
     }
-    calloc.free(openFileNameW);
+    return switch (type) {
+      FileType.any => const [(name: 'All Files (*.*)', pattern: '*.*')],
+      FileType.audio => const [
+        (
+          name: 'Audios (*.aac,*.midi,*.mp3,*.ogg,*.wav,*.m4a)',
```
navaronbracke: Is it "Audios" ? Or "Audio" or "Audio Files" ?

packages/file_picker_windows/test/file_picker_windows_test.dart
```diff
@@ -73,4 +73,37 @@ void main() {
...
+        () => FilePickerWindows.fileTypeFilterSpecs(FileType.custom, null),
+        throwsArgumentError,
+      );
+      expect(
+        () => FilePickerWindows.fileTypeFilterSpecs(FileType.custom, []),
+        throwsArgumentError,
+      );
+    });
+
+    test('media type produces separate video and image specs', () {
```
navaronbracke, line 122: Should we verify the full filter set for media / image / audio / etc ?

navaronbracke (approved):
LGTM with nits

packages/file_picker_windows/lib/src/file_picker_windows.dart
vicajilau: Went with `Audio`, matches the naming the Linux implementation already uses for the same filter (`info["Audio"]` in `file_picker_linux/lib/src/filters.dart`)

packages/file_picker_windows/test/file_picker_windows_test.dart
vicajilau: Added exact filter-spec assertions for FileType.image and FileType.audio, and tightened the media test to check full spec values instead of substring matches

PR #2177, opened 2026-08-26
Add configurable Darwin asset representation mode

jkerb8:
## Summary

Adds a configurable Darwin asset representation mode for media selected through the iOS photo library.

Callers can choose between:

- `DarwinAssetRepresentationMode.automatic` — preserves the existing system-default behavior.
- `DarwinAssetRepresentationMode.current` — requests the asset’s current representation to avoid transcoding when possible.
- `DarwinAssetRepresentationMode.compatible` — requests a broadly compatible representation, allowing transcoding when necessary.

## Motivation

Previous iOS implementations (version 8.0.7 is an example) allowed callers to disable compression, which configured `PHPickerConfiguration.preferredAssetRepresentationMode` as `.current`. After the Darwin implementation was rewritten, this configuration was no longer exposed and `PHPicker` defaulted to `.automatic`.

For large videos—particularly HEVC, HDR, slow-motion, or edited assets—`.automatic` may require significant preparation or transcoding before `pickFiles()` returns. Allowing callers to request `.current` can substantially reduce this delay when the original representation is suitable.

## API example

```dart
final files = await FilePicker.pickFiles(
  type: FileType.video,
  compressionQuality: 0,
  darwinOptions: const DarwinOptions(
    assetRepresentationMode: DarwinAssetRepresentationMode.current,
  ),
);
```

`automatic` remains the default, preserving existing behavior.

Non-automatic representation modes require `compressionQuality` to be `0`. An `ArgumentError` is thrown for conflicting configurations instead of silently ignoring either option.

## Implementation

- Adds `DarwinOptions` and `DarwinAssetRepresentationMode` to the platform interface.
- Passes Darwin options through the federated `pickFile()` and `pickFiles()` APIs.
- Maps the Dart enum to `PHPickerConfiguration.preferredAssetRepresentationMode`.
- Applies the option only to the iOS photo-library media picker.
- Leaves macOS, iOS document picking, and non-Darwin platforms unchanged.
- Adds API forwarding, validation, and Darwin method-channel tests.
- Documents the new option and its compression constraint.

## Validation

- Analysis passes across all federated packages.
- All federated package test suites pass.
- The iOS example compiles successfully.
- Verified in a production Flutter application with large video selection.
- Android and unsigned iOS application builds pass.


vicajilau:
Correction on the previous comment: the conflicts aren't from the Windows/publish workflow changes I mentioned, those don't touch any of the same files. The real cause is #2181, merged today, which also modifies file_picker.dart, file_picker_darwin.dart, and file_picker_platform_interface.dart (it removes the direct file_picker_darwin import from the facade and adds a skipEntitlementsChecks() hook to the platform interface). Sorry for the noise, should have checked before posting.

vicajilau (approved):

Issue #2156, opened 2026-08-20
[Proposal] provide a replacement for the old `saveFile()` method in version 11.0.0


driftwoodstudio:
[DriftwoodStudio comment: this is not an "edit" of what I wrote. **I didn't write any of this.** This is someone else's post/text with my name attached to it as the author. I didn't write this. @navaronbracke decided to completely rewrite it and leave my name on it as "author" instead of creating a new issue, and referring to this one after closing it. I am **NOT making this proposal. This is NOT my suggestion.** I support some parts of the idea here, but not all.]

Currently as it stands, file_picker 11.0.0 had a `saveFile()` implementation that had ambiguous functionality:

On Android, iOS, MacOS, Windows and Linux, it would open a save file dialog and only if the bytes were provided would it write those to that destination, which contradicts the naming of that method, which implied that it would save bytes to a file. (instead of merely choosing a destination to write bytes to)

On the web it would just download the given bytes to disk, since picking a save file destination was not implemented as such.
That is tracked in https://github.com/miguelpruivo/flutter_file_picker/issues/1732

The ambiguity around "saving bytes to a file" versus "choosing a location to write a file to" is why the nullability of the filename / bytes was updated in version 12.0.0, with the guarantee that `saveFile()` now lets the user pick a location and will save the provided bytes (which are now required) to said location.

However, this leaves us with a feature gap:
1) The original functionality, where the user could pass `bytes: null` to essentially pick a location and do the write themselves, is now gone, since `saveFile()` now always writes to the destination.
2) A side effect of `saveFile()` is now also apparent: Writing a large `Uint8List` can make the application run out of memory.

I propose that we do the following:
1) Update the `saveFile()` documentation, to mention that writing a large chunk of bytes shouldbe avoided, in which case the new alternative can be used.
2) Add a new `Future<Uri?> openSaveFileDialog()` method, that takes a filename argument, but **no bytes**, and returns the full Uri to a location on disk, where the user can write bytes to. This method will need to make sure that the Uri is writable, and returns null when the operation is cancelled.
Since it is up to the user to open the file and write to it, they can choose between a single write or a streamed write, using a mechanism of their choice.

This is essentially what https://github.com/miguelpruivo/flutter_file_picker/issues/2036 is asking as well.

_The original poster's question is kept below for posterity reasons_

----

[DriftwoodStudio: **this is what I actually wrote**.]
[...]

PR #2179, opened 2026-08-26
docs: replace dead Wiki links with README pointers, restore verified setup notes

vicajilau:
## Why

The Wiki was deleted following the discussion in #2156, but every link to it across the root README, the `file_picker` README, and the bug report template still pointed at it, all dead links now.

## What

- Replaced every Wiki link with pointers to the relevant README or the pub.dev API reference.
- Restored two setup notes that used to live only on the Wiki and have no other home, but only after verifying each against the current native implementations rather than copying the Wiki as is. The Wiki itself had gone stale (that mismatch is part of why it was removed, see the saveFile discussion in #2156), so copying it verbatim would have reintroduced the same problem:
  - **macOS**: the file access entitlements (`file_picker_darwin/README.md`). Verified these are still checked by name in `MacOSFilePickerHandler.swift`.
  - **Android**: the note about calling `super.onActivityResult()` in a custom `MainActivity` (`android_file_picker/README.md`). Verified `FilePickerDelegate` still relies on `onActivityResult` today.
- Deliberately did **not** carry over the Wiki's iOS `Info.plist` section (`NSAppleMusicUsageDescription`, `UIBackgroundModes`, `NSPhotoLibraryUsageDescription`). Checked `IOSFilePickerHandler.swift`: it uses `PHPickerViewController` and `UIDocumentPickerViewController` with `contentTypes: [.audio]`, neither of which needs those keys. That section was leftover from a much older implementation.

## Test plan

Docs only, no code changes.
- [x] Checked the whole repo for any remaining `flutter_file_picker/wiki` links outside of historical `CHANGELOG.md` entries (left those alone, changelogs are a historical record).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,7 +6,7 @@ labels: new issue
 
 ---
 
-Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened inssues. Also, check if it isn't described on the [Wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki), specially on [Troubleshooting](https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting) page.
+Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened issues.
 
 Also, sometimes a simple `flutter clean` and `flutter build` again with latest file_picker version, may end up by fixing cached issues, so I encourage you to first do so.
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,7 +6,7 @@ labels: new issue
 
 ---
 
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,7 +6,7 @@ labels: new issue
 
 ---
 
-Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened inssues. Also, check if it isn't described on the [Wiki](https://github.com/miguelpruivo/flutter_file_picker/wiki), specially on [Troubleshooting](https://github.com/miguelpruivo/flutter_file_picker/wiki/Troubleshooting) page.
+Before creating an issue, make sure that you are on the latest `file_picker` version and that there aren't already any similar opened issues.
```
navaronbracke: Link to the issues tab on Github?

navaronbracke (approved):
LGTM with nit

.github/ISSUE_TEMPLATE/bug_report.md
vicajilau: Linked to https://github.com/miguelpruivo/flutter_file_picker/issues

PR #2181, opened 2026-08-27
fix(file_picker): stop importing file_picker_darwin so the facade is wasm compatible

Chachuke:
This is purely a pub.dev score fix. No behaviour change on any platform.

Together with #2180 it takes `file_picker` from **120/160 to 160/160**. This PR is worth 30 of those 40 points.

## The problem

`file_picker` is the app-facing package of a federated plugin, so it should only ever import the platform *interface*. It also imports the macOS/iOS *implementation* directly, and only for one thing, the type test in `skipEntitlementsChecks()`:

```dart
import 'package:file_picker_darwin/file_picker_darwin.dart';
...
  }
}
```

That single import drags `dart:io` into the facade's import graph. It costs two separate score sections:

**1. Platform support, 10/20.** pana's trace on 12.1.1:

```
Package not compatible with runtime wasm
Because:
* package:file_picker/file_picker.dart that imports:
* package:file_picker/src/file_picker.dart that imports:
* package:file_picker_darwin/file_picker_darwin.dart that imports:
* package:file_picker_darwin/src/file_picker_darwin.dart that imports:
* dart:io
```

Worth stressing: nothing in this workspace is actually web-unclean. `file_picker_web` already uses `package:web`, and `file_picker_platform_interface` already guards its own `dart:io` behind a conditional export. The facade's import is the only thing standing between this package and `is:wasm-ready`.

**2. Dependency lower bounds, 0/20.** The same type test is the *sole* error in the downgrade analysis:

```
UNDEFINED_METHOD - lib/src/file_picker.dart:261:22
- The method 'skipEntitlementsChecks' isn't defined for the type 'FilePickerDarwin'.
```

## The fix

Route the call through the platform interface, following the `clearTemporaryFiles()` hook that already exists a few lines above:

- `file_picker_platform_interface`: add `Future<void> skipEntitlementsChecks() async {}`, a default no-op
- `file_picker_darwin`: mark its existing method `@override`, body untouched
- `file_picker`: drop the import, dispatch on `FilePickerPlatform.instance`

Behaviour is identical. Before, a non-Darwin platform fell through the type test and did nothing; now it hits the interface no-op and does nothing. On macOS the same `FilePickerDarwin` method runs, and its `if (!Platform.isMacOS) return;` guard is untouched.

## On the version constraints

`file_picker` now requires `file_picker_darwin: ^1.0.4` rather than `^1.0.0`, and that bound is load-bearing. With the old bound, an older `file_picker_darwin` would not override the new hook, so `skipEntitlementsChecks()` would silently fall back to the interface no-op on macOS, which is exactly the regression 12.1.1 fixed. The two implementation packages are bumped to `3.2.0` and `1.0.4` so those constraints resolve; publishing them before `file_picker` is what makes the bound valid on pub.dev.

I left `file_picker`'s own version and all three `CHANGELOG.md` files alone, so this slots into whatever release shape you want.

## Verification

- `dart format --output=none --set-exit-if-changed` on `packages/` — clean
- `flutter analyze` on all three changed packages — no issues. The `@override` resolving is itself proof the new hook is wired correctly.
- `flutter test` on all three — passing, including two new tests: the default no-op on the interface, and that `FilePicker.skipEntitlementsChecks()` actually reaches the platform instance. That second one is a regression test for the bug fixed in 12.1.1, where this method was a permanent no-op.
- WASM: I reproduced pana's reachability analysis locally and validated it by reproducing the published 12.1.1 trace above chain-for-chain on `master`. On this branch `dart:io` is no longer reachable from `package:file_picker/file_picker.dart`.

### DIFF
--- a/packages/file_picker/CHANGELOG.md
+++ b/packages/file_picker/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 12.1.2
+
+### General
+- Stopped importing `file_picker_darwin` directly from the facade package. `skipEntitlementsChecks()` now dispatches through the new `FilePickerPlatform.skipEntitlementsChecks()` hook instead of a type test on `FilePickerPlatform.instance`. That import pulled `dart:io` into `file_picker`'s import graph, making the package incompatible with wasm and breaking analysis when resolved against the lowest allowed dependency versions. No behavior change on any platform.
+
 ## 12.1.1
 
 ### macOS
--- a/packages/file_picker/lib/src/file_picker.dart
+++ b/packages/file_picker/lib/src/file_picker.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 
 import 'package:android_file_picker/android_file_picker.dart';
-import 'package:file_picker_darwin/file_picker_darwin.dart';
 import 'package:file_picker_platform_interface/file_picker_platform_interface.dart';
 import 'package:flutter/foundation.dart';
 
@@ -257,8 +256,6 @@ abstract final class FilePicker {
   /// Note: skipping entitlement checks may lead to unexpected behavior if the
   /// app is actually sandboxed. Use with caution.
   static Future<void> skipEntitlementsChecks() async {
-    if (FilePickerPlatform.instance case final FilePickerDarwin instance) {
-      await instance.skipEntitlementsChecks();
-    }
+    await FilePickerPlatform.instance.skipEntitlementsChecks();
   }
 }
--- a/packages/file_picker/pubspec.yaml
+++ b/packages/file_picker/pubspec.yaml
@@ -21,9 +21,9 @@ dependencies:
   flutter:
     sdk: flutter
 
-  file_picker_platform_interface: ^3.1.0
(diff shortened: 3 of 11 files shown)


vicajilau (approved):

vicajilau (approved):
Solid fix. It follows the exact pattern already used for clearTemporaryFiles(): a no-op default on FilePickerPlatform, an override in FilePickerDarwin, and the facade now dispatches through the interface instead of importing file_picker_darwin directly. That import was the only thing pulling dart:io into file_picker's import graph, so it explains both the wasm incompatibility and the downgrade analysis failure.

Reasoning on the file_picker_darwin: ^1.0.4 bound also checks out, skipEntitlementsChecks() was added in 1.0.2, so a lower bound is technically possible, but pinning to the version this PR itself publishes is the simpler and safer choice.

Added the missing CHANGELOG entries and a version bump for file_picker itself (12.1.2), the rest was already in good shape: tests cover both the new interface no-op and the regression this fixes, and CI is green across every platform.

PR #2184, opened 2026-08-28
fix(linux): stop dropping options passed as plain LinuxOptions

### LINKED ISSUE #2183: [Linux] LinuxOptions.lockParentWindow and acceptLabel are silently ignored
### Describe the bug

`file_picker_linux` silently ignores `lockParentWindow` and `acceptLabel` when they are passed on a plain `LinuxOptions`, which is the API our own deprecation message points people at:

```dart
@Deprecated(
  'Use WindowsOptions.lockParentWindow or LinuxOptions.lockParentWindow instead; ...',
)
```

All three entry points (`pickFiles`, `getDirectoryPath`, `saveFile`) narrow the incoming options and fall back to a default instance when the runtime type is not `FilePickerLinuxOptions`, which throws away whatever the caller set:

```dart
final FilePickerLinuxOptions options = switch (linuxOptions) {
  FilePickerLinuxOptions opts => opts,
  _ => const FilePickerLinuxOptions(),   // caller's values dropped here
};
```

Observed:

```
LinuxOptions(lockParentWindow: true)           -> modal = false
FilePickerLinuxOptions(lockParentWindow: true) -> modal = true
```

`modal` is what gets sent to `org.freedesktop.portal.FileChooser`, so the dialog is simply not modal.

### Second, related problem

`FilePickerLinuxOptions` cannot carry `acceptLabel` at all, because its constructor does not forward it:

```dart
const FilePickerLinuxOptions({this.parentWindow, super.lockParentWindow});
```

`FilePickerLinuxOptions(acceptLabel: 'Choose')` does not compile. So `acceptLabel` and `parentWindow` are mutually exclusive today: you can have one or the other, never both.

### Expected behavior

Options set by the caller reach the portal regardless of which of the two option types is used, and `FilePickerLinuxOptions` can carry every option its supertype defines.

### Platform

- [x] Desktop (Linux)

### Notes

Found while answering a question on #1993 about whether the portal based Linux implementation supports `lockParentWindow`. It does, but only through `FilePickerLinuxOptions`, which is not what the docs tell you to use.

vicajilau:
Fixes #2183. Found while answering a question on #1993 about whether the portal based Linux implementation supports `lockParentWindow`.

## The bug

All three entry points (`pickFiles`, `getDirectoryPath`, `saveFile`) narrow the incoming `LinuxOptions` and fall back to a **default** instance when the runtime type is not `FilePickerLinuxOptions`, throwing away whatever the caller set:

```dart
final FilePickerLinuxOptions options = switch (linuxOptions) {
  FilePickerLinuxOptions opts => opts,
  _ => const FilePickerLinuxOptions(),   // caller's values dropped here
};
```

```
LinuxOptions(lockParentWindow: true)           -> modal = false
FilePickerLinuxOptions(lockParentWindow: true) -> modal = true
```

`modal` is what goes to `org.freedesktop.portal.FileChooser`, so the dialog is simply not modal. And `LinuxOptions.lockParentWindow` is precisely what the deprecation on `FilePicker.pickFiles` tells people to migrate to, so the documented path is the broken one.

What hid this: `acceptLabel` was read from the original `linuxOptions` while everything else read from the narrowed `options`. That asymmetry meant `acceptLabel` kept working and nothing looked wrong.

## Second, related problem

`FilePickerLinuxOptions` could not carry `acceptLabel` at all, because its constructor did not forward it:

```dart
const FilePickerLinuxOptions({this.parentWindow, super.lockParentWindow});
```

So `acceptLabel` and `parentWindow` were mutually exclusive: you could have one or the other, never both. `FilePickerLinuxOptions(acceptLabel: 'Choose')` did not even compile.

## The fix

- The fallback branch carries the incoming values across instead of discarding them.
- `acceptLabel` is read from the narrowed options like every other option, removing the asymmetry.
- `FilePickerLinuxOptions` forwards `acceptLabel` to its supertype.

No behaviour change for callers already using `FilePickerLinuxOptions`, they hit the first branch of the switch either way.

## Tests

Four regression tests in `packages/file_picker_linux/test`, covering both option types and both fields, plus the fact that `FilePickerLinuxOptions` can now hold `acceptLabel` and `parentWindow` at the same time.

Bumped to 1.0.2 with a CHANGELOG entry, since 1.0.1 is already on pub.dev.

## Docs

The package README documented what the package is and how to regenerate the DBus bindings, but not a single option. `lockParentWindow`, `acceptLabel` and `parentWindow` appeared in no README in the repo, while `file_picker`'s own README explicitly points readers at the platform package for this.

Added a "Linux specific options" section covering the three, including the two things you cannot infer from the API:

- `lockParentWindow` does nothing visible on its own, because the portal needs `parentWindow` to know which window to be modal against.
- `acceptLabel` only applies to file picking, not to directory selection or saving.

Plus the accepted `parentWindow` formats, and a note that obtaining the handle is on the caller and is impractical under Wayland.


packages/file_picker_linux/test/file_picker_linux_test.dart
```diff
@@ -22,4 +22,56 @@ void main() {
       expect(filter.info.containsKey('Custom'), isTrue);
     });
   });
+
+  group('LinuxOptions passed through to the portal', () {
+    // The implementation narrows LinuxOptions to FilePickerLinuxOptions. It
+    // used to fall back to a default instance when the incoming value was a
+    // plain LinuxOptions, silently dropping whatever the caller had set, which
+    // is exactly what the deprecation on FilePicker.pickFiles points people at.
+    FilePickerLinuxOptions narrow(LinuxOptions linuxOptions) =>
```
navaronbracke: This is not actually testing the underlying implementation? Shouldn't the narrowing function in the implementation be exposed as visible for testing instead?

navaronbracke (changes requested):

packages/file_picker_linux/test/file_picker_linux_test.dart
vicajilau: You are right, and it was worse than not testing the implementation: the test defined its own copy of the narrowing switch, so breaking the real code left it green. It proved nothing.

Extracted it as you suggested. The switch was also duplicated across all three entry points, so it is now a single `FilePickerLinux.resolveOptions` marked `@visibleForTesting`, shared by `pickFiles`, `getDirectoryPath` and `saveFile`, and the tests call it directly.

I checked that it actually bites now: reintroducing the original bug in the implementation fails two tests, where before it failed none.

Also added a case asserting that a `FilePickerLinuxOptions` comes back identical rather than rebuilt, which the previous shape could not express.

packages/file_picker_linux/lib/src/file_picker_linux.dart
```diff
@@ -302,6 +293,22 @@ class FilePickerLinux extends FilePickerPlatform {
...
   }
 
+  /// Narrows [linuxOptions] to the Linux specific type the portal calls need.
+  ///
+  /// Callers may pass either a plain [LinuxOptions] or a
+  /// [FilePickerLinuxOptions]. Only the latter carries `parentWindow`, but
+  /// everything the supertype defines has to survive the narrowing, otherwise
+  /// options set by the caller are silently dropped on the way to the portal.
+  @visibleForTesting
+  static FilePickerLinuxOptions resolveOptions(LinuxOptions linuxOptions) =>
```
navaronbracke: Could this use a block body? Arrow functions are good if everything fits on one line

navaronbracke (approved):
LGTM with formatting nit

packages/file_picker_linux/lib/src/file_picker_linux.dart
vicajilau: Done!!

PR #2188, opened 2026-08-30
fix(file_picker): require implementations built against the interface it pins

### LINKED ISSUE #2186: You forgot to bump platform implementations version!
https://github.com/miguelpruivo/flutter_file_picker/blob/c75ee793a0d62bd87dbdc9b574f6656685ca0bc6/packages/file_picker/pubspec.yaml#L25-L28

These versions doesn't match with this
https://github.com/miguelpruivo/flutter_file_picker/blob/c75ee793a0d62bd87dbdc9b574f6656685ca0bc6/packages/file_picker/pubspec.yaml#L24

causing the issue below:

```
../../../../../.pub-cache/hosted/pub.dev/windows_file_picker-1.0.1/lib/src/file_picker_windows.dart:54:25: Error: The method 'FilePickerWindows.pickFile' has fewer named arguments than those of overridden method 'FilePickerPlatform.pickFile'.
  Future<PlatformFile?> pickFile({
                        ^
../../../../../.pub-cache/hosted/pub.dev/file_picker_platform_interface-3.2.0/lib/src/file_picker_platform_interface.dart:46:25: Context: This is the overridden method ('pickFile').
  Future<PlatformFile?> pickFile({

```

vicajilau:
Fixes #2186.

## The bug

`file_picker` required `file_picker_platform_interface: ^3.2.0` while still allowing implementations published against `^3.0.0`, which predate the `DarwinOptions` parameter added in 3.2.0. Every constraint was satisfied and the result did not compile.

Confirmed by resolving a clean project at the minimum versions the facade allowed:

```
file_picker_platform_interface   3.2.0   <- forced by ^3.2.0
android_file_picker              1.0.0   <- built against ^3.0.0
file_picker_darwin               1.0.4   <- correct
file_picker_linux                1.0.0   <- built against ^3.0.0
file_picker_web                  3.0.0   <- built against ^3.0.0
windows_file_picker              1.0.0   <- built against ^3.0.0
```

**Four of the five lower bounds were wrong**, not just Windows as reported. `file_picker_darwin` was the only one raised, probably because `DarwinOptions` came from there and the connection was obvious.

## The fix

| Implementation | First version built against 3.2.0 | Was | Now |
| --- | --- | --- | --- |
| `android_file_picker` | 1.0.3 | `^1.0.0` | `^1.0.3` |
| `file_picker_darwin` | 1.0.4 | `^1.0.4` | unchanged |
| `file_picker_linux` | 1.0.2 | `^1.0.0` | `^1.0.2` |
| `file_picker_web` | 3.0.3 | `^3.0.0` | `^3.0.3` |
| `windows_file_picker` | 1.1.0 | `^1.0.0` | `^1.1.0` |

No implementation needs republishing, they are all correct on pub.dev. Only the facade's bounds were stale. Bumped to 12.1.3.

Anyone hitting this today can work around it with `flutter pub upgrade`, since their lockfile is pinning old versions that the loose constraints gave pub no reason to move off.

## The guard

This will recur. Adding a parameter to an abstract method on the platform interface is breaking for implementers even when it is source compatible for callers, and it shipped as a minor. The only safety net is the facade's lower bounds, which is exactly what was missed.

**Updated based on review**: this used to be a custom tool (`tool/check_facade_constraints.dart`) that hand-parsed pubspec YAML and hand-rolled semver comparison to model whether the bug would occur. Dropped that approach, it tested an approximation of the bug rather than the bug itself, a parsing mistake could report green while a real resolution still did not compile.

Replaced it with a CI step that reproduces the actual failure. A standalone Flutter app, created fresh in CI outside this workspace so its dependencies resolve from pub.dev instead of the local checkout, depends on the facade via a path dependency. `pub downgrade` then forces every dependency to the lowest version its constraint allows, exactly what an external app's lockfile could legitimately resolve to.

`flutter analyze` does not catch this on its own, a plugin's platform-specific implementation only enters the type-checked graph through the generated plugin registrant, and only a real build produces that. `flutter build web` does catch it, and empirically catches every platform's implementation at once, not just web's, since dart2js compiles the whole registrant regardless of which platform each implementation targets. No extra SDK setup needed either, unlike building for Windows/Linux/Android/iOS/macOS.

## Verification

`dart format` clean, `flutter analyze` clean across all 8 packages, `melos run test` green, and the new CI step verified in both directions locally: `flutter build web` against a downgraded resolution fails with the exact error from #2186 on master before this PR (across `android_file_picker`, `file_picker_web`, and more, not just the `windows_file_picker` originally reported), and succeeds against this branch.


tool/test/check_facade_constraints_test.dart
```diff
@@ -0,0 +1,112 @@
+import 'package:flutter_test/flutter_test.dart';
```
navaronbracke: I'm not too sure if adding a custom tool for this is the right thing to do? IIRC there is https://pub.dev/documentation/pub_semver/latest/pub_semver/

But I think we should just go with an integration smoke test instead? (since the wrong constraint breaks builds)

navaronbracke (changes requested):
The fix for the pubspec is LGTM (we can release that already) but I'm not too happy with how we will enforce this in tests/tooling

tool/test/check_facade_constraints_test.dart
vicajilau: Agreed, dropped the custom tool entirely rather than swapping in pub_semver, that would have fixed the version-comparison fragility but left the deeper issue: it was still testing a model of the bug (parsed constraints vs pub.dev metadata) instead of the bug itself.

Went with the integration smoke test you suggested. A standalone app outside this workspace depends on the facade, `pub downgrade` forces the lowest allowed versions, then a real `flutter build web` proves whether that combination compiles. Verified locally in both directions: it reproduces the exact #2186 error against master before this PR (turns out it's not just windows_file_picker, android_file_picker and file_picker_web hit it too), and passes against this branch. Details in the updated PR description.

navaronbracke (approved):
LGTM, I think in the future we could set up integration tests, instead of a script in the CI pipeline, to harden this even further (using a way that is more unit-test-like)

That integration test could build the example on web, and test if the buttons in the example work. (using a testWidgets() test)

But this is good for now.

PR #2191, opened 2026-09-01
feat(platform_interface): add PlatformFile.size

### LINKED ISSUE #2187: PlatformFile.size property
**Describe the bug**
Dear team, Thank you for the File picker package.

Release v12.0 silently removes `PlatformFile.size` property without corresponding update in documentation. Also there are no alternative (`PlatformFile.length()` exists, but it is async)

**Platform**

- [ ] Android
- [ ] iOS
- [ ] Web
- [ ] Desktop
- [x] All

Requesting for an alternative which is not async. Something like `PlatformFile.lengthSync()` would be fine.

Happy to post a PR if needed

vicajilau:
Fixes #2187.

## The request

`PlatformFile.size` was removed in the v12 federated rewrite without a synchronous replacement, only `Future<int> length()` remains. That forces an `await` for a value most callers already have on hand.

## The fix

Every platform's `PlatformFile` already carries the size the native picker reports, stored in a private `_bytesLength` field. `length()` already returns it immediately when it's known and only falls back to actually reading the file (`xFile.length()`) when it isn't, so the async signature exists for that fallback path, not because getting the size is inherently async.

Added `int? get size` to `PlatformFile`, returning that already known value synchronously, `null` when it isn't known. Matches the shape `extension` and `path` already use on this class.

Per platform, what `size` actually returns:

| Platform | Source | Known upfront? |
| --- | --- | --- |
| Android | native pick result (`FileInfo.size`) | Yes |
| iOS / macOS | native pick result | Yes |
| Web | browser `File.size`, or loaded bytes | Yes |
| Windows | picker only returns a path | No, `null` |
| Linux | XDG portal only returns a path | No, `null` |

Windows and Linux callers fall back to `await length()`, which still works exactly as before.

## What changed

- `file_picker_platform_interface` 3.2.0 → 3.3.0: added the abstract member.
- `file_picker_darwin`, `android_file_picker`, `file_picker_web` bumped (minor): implemented `size` backed by data they already have.
- `file_picker_linux`, `windows_file_picker` bumped (minor): implemented `size`, returns `null` since neither picker reports it.
- Each implementation's `file_picker_platform_interface` lower bound raised to `^3.3.0`, the version that introduces the abstract member they now implement. Learned this the hard way with #2186, an implementation built against an older interface won't have this override and will fail to compile if pub resolves it alongside a newer interface.
- `file_picker` (the facade) is untouched, `PlatformFile` is used through the platform interface as-is, no facade code references `size` directly.

## Test plan

- [x] `flutter analyze` clean across the whole workspace
- [x] `dart format --output=none --set-exit-if-changed` clean
- [x] `flutter test` across the whole workspace, 57 tests pass, including new coverage for `size` on every platform (`web`'s is `@TestOn('browser')` like the rest of that file, doesn't run under the default `flutter test`)

### DIFF
--- a/packages/file_picker/test/file_picker_test.dart
+++ b/packages/file_picker/test/file_picker_test.dart
@@ -19,7 +19,7 @@ base class TestPlatformFile extends PlatformFile {
   XFile get xFile => XFile(path ?? '');
 
   @override
-  int? get size => 100;
+  int? lengthSync() => 100;
 
   @override
   Future<int> length() async => 100;
--- a/packages/file_picker/test/file_picker_test.dart
+++ b/packages/file_picker/test/file_picker_test.dart
@@ -19,6 +19,9 @@ base class TestPlatformFile extends PlatformFile {
   XFile get xFile => XFile(path ?? '');
 
   @override
+  int? get size => 100;
+
+  @override
   Future<int> length() async => 100;
 
   @override
--- a/packages/file_picker_android/CHANGELOG.md
+++ b/packages/file_picker_android/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.1.0
 
-- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously.
+- Implemented `PlatformFile.lengthSync()`, returning the length Android already reports for a picked file synchronously.
 
 ## 1.0.4
 
--- a/packages/file_picker_android/CHANGELOG.md
+++ b/packages/file_picker_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.0
+
+- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously, without doing any I/O.
+
 ## 1.0.3
 
 - Tightened the `file_picker_platform_interface` lower bound to `^3.2.0`. This package's `pickFile()`/`pickFiles()` signatures reference `DarwinOptions`, added in that version, so resolving against an older one would fail to compile.
--- a/packages/file_picker_android/CHANGELOG.md
+++ b/packages/file_picker_android/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.1.0
 
-- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously, without doing any I/O.
+- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously.
 
 ## 1.0.3
 
--- a/packages/file_picker_linux/CHANGELOG.md
+++ b/packages/file_picker_linux/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.0
+
+- Implemented `PlatformFile.size`. XDG Desktop Portals only return a file path, not its size, so this returns `null` on Linux, use `length()` to actually read it from disk.
+
 ## 1.0.2
 
 - Fixed `LinuxOptions.lockParentWindow` and `LinuxOptions.acceptLabel` being silently ignored. Passing a plain `LinuxOptions` fell back to a default `FilePickerLinuxOptions`, dropping whatever the caller had set, which is the exact API the deprecation on `FilePicker.pickFiles` points to. [#2183](https://github.com/miguelpruivo/flutter_file_picker/issues/2183)
--- a/packages/file_picker_linux/CHANGELOG.md
+++ b/packages/file_picker_linux/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.1.0
 
-- Implemented `PlatformFile.size`. XDG Desktop Portals only return a file path, not its size, so this returns `null` on Linux, use `length()` to actually read it from disk.
+- Implemented `PlatformFile.lengthSync()`. XDG Desktop Portals only return a file path, not its size, so this returns `null` on Linux, use `length()` to actually read it from disk.
 
 ## 1.0.2
 
--- a/packages/file_picker_linux/lib/src/linux_platform_file.dart
+++ b/packages/file_picker_linux/lib/src/linux_platform_file.dart
@@ -50,6 +50,8 @@ base class LinuxPlatformFile extends PlatformFile {
     return XFile(uri.toString(), name: name);
   }
 
+  /// `null` for a picked file. XDG Desktop Portals only return a file path,
+  /// not its size; use [length] to read it from disk instead.
   @override
   int? get size {
     final len = _bytesLength;
--- a/packages/file_picker_linux/lib/src/linux_platform_file.dart
+++ b/packages/file_picker_linux/lib/src/linux_platform_file.dart
@@ -50,8 +50,10 @@ base class LinuxPlatformFile extends PlatformFile {
     return XFile(uri.toString(), name: name);
   }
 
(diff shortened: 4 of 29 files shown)


packages/file_picker/test/file_picker_test.dart
```diff
@@ -18,6 +18,9 @@ base class TestPlatformFile extends PlatformFile {
   @override
   XFile get xFile => XFile(path ?? '');
 
+  @override
+  int? get size => 100;
```
navaronbracke: This should be `lengthSync` ? Otherwise there's `size` and `length()`, which do the same thing, but are named differently

packages/file_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.0
+
+- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously, without doing any I/O.
```
navaronbracke:
```suggestion
- Implemented `PlatformFile.size`, returning the size Android already reports for a picked file synchronously.
```

packages/file_picker_linux/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.0
+
+- Implemented `PlatformFile.size`. XDG Desktop Portals only return a file path, not its size, so this returns `null` on Linux, use `length()` to actually read it from disk.
```
navaronbracke: Should we document little differences like this in the PlatformFIle subclasses as well? (you can put dartdoc on overrides iirc)

packages/file_picker_windows/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.2.0
+
+#### Desktop (Windows)
```
navaronbracke, line 3: Silly remark, but should we clean up changelogs in a new PR, because `Desktop (Windows)` should be removed? It _is_ the Windows implementation. Probably the same thing for the other implementations

navaronbracke (changes requested):

packages/file_picker/test/file_picker_test.dart
vicajilau: Replied on #2187, XFile turned out not to have lengthSync() after all, checked the source. Left the reasoning on why size and length() aren't the same operation there too, will switch to lengthSync() if you still want it after that.

packages/file_picker_android/CHANGELOG.md
vicajilau: Accepted, done.

packages/file_picker_linux/CHANGELOG.md
vicajilau: Good idea, added a doc line on each platform's size override describing what it actually returns there, not just in the changelog.

packages/file_picker_windows/CHANGELOG.md
vicajilau: Agreed it's redundant, but leaving it as is here to keep this PR scoped to size. Happy to clean up the changelog headers across all the platform packages in a separate PR if you want it.

packages/file_picker/test/file_picker_test.dart
navaronbracke: Ah, my bad, yeah it's dart:io File that has both

packages/file_picker_windows/CHANGELOG.md
navaronbracke: Yeah let's do that separately

packages/file_picker/test/file_picker_test.dart
navaronbracke: I would like to keep `length() / lengthSync()`, to keep the naming consistent.

packages/file_picker_linux/lib/src/linux_platform_file.dart
```diff
@@ -50,6 +50,14 @@ base class LinuxPlatformFile extends PlatformFile {
     return XFile(uri.toString(), name: name);
   }
 
+  /// `null` for a picked file. XDG Desktop Portals only return a file path,
```
navaronbracke: This doc line reads a bit weird?

packages/file_picker_windows/lib/src/windows_platform_file.dart
```diff
@@ -50,6 +50,14 @@ base class WindowsPlatformFile extends PlatformFile {
     return XFile(uri.toString(), name: name);
   }
 
+  /// `null` for a picked file. The picker only returns a file path, not its
```
navaronbracke: This doc line reads a bit weird?

packages/file_picker_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.3.0
+
+- Added `PlatformFile.size`, a nullable `int` that returns the file's size in bytes immediately when the native picker already reported it, without doing any I/O. Returns `null` when it isn't known upfront; use `length()` in that case, which falls back to reading the file. Restores a synchronous way to get a file's size, addressing [#2187](https://github.com/miguelpruivo/flutter_file_picker/issues/2187).
```
navaronbracke:
```suggestion
- Added `PlatformFile.size`, a nullable `int` that returns the file's size in bytes, as provided by the underlying file picker result. 
```

navaronbracke (changes requested):

packages/file_picker/test/file_picker_test.dart
vicajilau: Done, renamed to `lengthSync()` across the interface, all five implementations, their tests and the changelogs. You are right that `size` alongside `length()` was two names for one thing, and it also matches what the issue originally asked for.

It is a method now rather than a getter, to line up with `dart:io`'s `File.length()` / `lengthSync()`.

One thing I deliberately left alone: `file.size` inside `file_picker_web` is the browser's `File.size`, not ours, so those stay.

packages/file_picker_linux/lib/src/linux_platform_file.dart
vicajilau: Yeah, it was clunky. Reworded it: says when the value is known first, then why a picked file has none.

packages/file_picker_windows/lib/src/windows_platform_file.dart
vicajilau: Same here, reworded. Thanks for catching both.

packages/file_picker_platform_interface/CHANGELOG.md
vicajilau: Took your wording, just with `lengthSync()` instead of `size` after the rename, plus a note on why it is named that way. Thanks!

packages/file_picker_windows/CHANGELOG.md
vicajilau: Opened #2194 for this. Turns out Windows was the only one left with it, the rest are already clean.

navaronbracke (approved):
LGTM, although I'm beginning to think we should maybe rethink the facade version checks.

Usually what happens is the platform interface package gets published first, on it's own, then a subsequent release of the child packages can depend on the new platform interface.

Is that a solution to the issue?

PR #2201, opened 2026-09-09
fix(android): stop mangling the file name for extensions Android doesn't recognize in saveFile

### LINKED ISSUE #2199: [Android] Saving file as gpx give .gpx.xml.gpx.xml extension
**Describe the bug**
When saving a file with gpx extension, the resulting extension is garbled.

**Platform**
- [X] Android

**Platform OS version**
Android 16

**How are you picking?**
```
      String gpxString = geoXml.toGpxString(pretty: false);
      Uint8List bytes = Uint8List.fromList(utf8.encode(gpxString));
        await FilePicker.saveFile(
          dialogTitle: "Save as GPX",
          fileName: 'trace.gpx',
          type: FileType.custom,
          allowedExtensions: ["gpx"],
          bytes: bytes,
        );
```

I tried setting the mimeType to "*/*", etc... to no avail.

**Error Log**
The event as in console :
{
W/FilePickerUtils(12872):    _id=47
W/FilePickerUtils(12872):    local_filename=/storage/emulated/0/Download/trace.gpx.xml.gpx
W/FilePickerUtils(12872):    mediaprovider_uri=null
W/FilePickerUtils(12872):    destination=6
W/FilePickerUtils(12872):    title=trace.gpx.xml.gpx
W/FilePickerUtils(12872):    description=trace.gpx.xml
W/FilePickerUtils(12872):    uri=non-dwnldmngr-download-dont-retry2download
W/FilePickerUtils(12872):    status=200
W/FilePickerUtils(12872):    hint=null
W/FilePickerUtils(12872):    media_type=text/xml
W/FilePickerUtils(12872):    total_size=774
W/FilePickerUtils(12872):    last_modified_timestamp=1788855962326
W/FilePickerUtils(12872):    bytes_so_far=0
W/FilePickerUtils(12872):    allow_write=1
W/FilePickerUtils(12872):    local_uri=file:///storage/emulated/0/Download/trace.gpx.xml.gpx
W/FilePickerUtils(12872):    reason=placeholder
W/FilePickerUtils(12872): }

**Screenshots and/or video**
If applicable, add screenshots or video to help explain your problem.

**Flutter Version details**
[✓] Flutter (Channel stable, 3.47.2, on Ubuntu 24.04.4 LTS 7.0.0-31-generic, locale
    fr_FR.UTF-8) [217ms]
    • Flutter version 3.47.2 on channel stable at /home/nicolasp/_apps/flutter
    • Framework revision d3b14c8769 (il y a 12 jours), 2026-08-26 16:07:51 -0700
    • Engine revision a804b26164
    • Dart version 3.13.2
    • DevTools version 2.60.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop,
      enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-native-assets, enable-record-use, enable-swift-package-manager,
      omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
    [2,5s]
    • Android SDK at /home/nicolasp/_apps/AndroidSDK/
    • Emulator version 36.6.11.0 (build_id 15507667) (CL:N/A)
    • Platform android-37.0, build-tools 36.1.0
    • Java binary at: /home/nicolasp/_apps/android-studio/jbr/bin/java
      This is the JDK bundled with the latest Android Studio installation on this
      machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14196175-b1038.72)
    • All Android licenses accepted.

[✓] Chrome - develop for the web [16ms]
    • Chrome at google-chrome

[✗] Linux toolchain - develop for Linux desktop [1 217ms]
    • Ubuntu clang version 18.1.3 (1ubuntu1)
    • cmake version 3.28.3
    ✗ ninja is required for Linux development.
      It is likely available from your distribution (e.g.: apt install ninja-build),
      or can be downloaded from https://github.com/ninja-build/ninja/releases
    • pkg-config version 1.8.1
    ✗ GTK 3.0 development libraries are required for Linux development.
      They are likely available from your distribution (e.g.: apt install
      libgtk-3-dev)
    • OpenGL core renderer: Mesa Intel(R) Iris(R) Xe Graphics (RPL-P)
    • OpenGL core version: 4.6 (Core Profile) Mesa 25.2.8-0ubuntu0.24.04.2
    • OpenGL core shading language version: 4.60
    • OpenGL ES renderer: Mesa Intel(R) Iris(R) Xe Graphics (RPL-P)
    • OpenGL ES version: OpenGL ES 3.2 Mesa 25.2.8-0ubuntu0.24.04.2
    • OpenGL ES shading language version: OpenGL ES GLSL ES 3.20
    • GL_EXT_framebuffer_blit: yes
    • GL_EXT_texture_format_BGRA8888: yes

[✓] Connected device (3 available) [148ms]
    • sdk gphone64 x86 64 (mobile) • emulator-5554 • android-x64    • Android 16 (API
      36) (emulator)
    • Linux (desktop)              • linux         • linux-x64      • Ubuntu 24.04.4
      LTS 7.0.0-31-generic
    • Chrome (web)                 • chrome        • web-javascript • Google Chrome
      152.0.7977.82

[✓] Network resources [270ms]

**Additional context**

vicajilau:
Fixes #2199.

## The bug

Saving a file with an extension unregistered in the device's `MimeTypeMap` (e.g. `.gpx`, a niche XML based format) produces a mangled name. The report's example: saving `trace.gpx` produced `trace.gpx.xml.gpx`.

## Root cause, traced through the actual save flow

1. `getMimeTypeForBytes` (`FileUtils.kt`) doesn't find `gpx` in the device's `MimeTypeMap`, so it falls back to sniffing the file's byte content. A GPX file is valid XML, so the sniff correctly detects `text/xml`, a real but wrong answer for what the caller asked for.
2. That mime type is set on the `ACTION_CREATE_DOCUMENT` intent alongside the suggested title `trace.gpx`. The system's document picker honors the declared type and "corrects" the suggested name by appending its own default extension for `text/xml`, `.xml`, producing the actual created document `trace.gpx.xml`.
3. `maybeRenameGenericMimeDuplicate` runs afterward to append the extension when the picker didn't add one (the case it was written for, generic mime types like `application/octet-stream`). It only checks whether the current name already ends with the extension we asked for, `gpx`, never whether it ends with a *different* one. `trace.gpx.xml` doesn't end in `.gpx`, so the whole string is treated as the base name and `.gpx` gets appended on top, producing `trace.gpx.xml.gpx`.

Not specific to `.gpx`, any extension unknown to the device whose content happens to sniff as a recognizable generic format (XML, in this case, likely others too) can hit the same chain.

## The fix

- `getMimeTypeForBytes`: when the extension is present but unknown to `MimeTypeMap`, use a wildcard mime type (`*/*`) instead of falling through to content sniffing. A wildcard has no default extension of its own, so the picker has nothing to "correct" the name with.
- `extractBaseNameAndSuffix` (used by `maybeRenameGenericMimeDuplicate`): added a case that recognizes our extension immediately followed by exactly one more extension, e.g. `trace.gpx.xml`, and strips both, recovering the true base name (`trace`) instead of only handling the case where no extension is present at all.

Verified the new regex's matching behavior directly (a Kotlin `Regex` and Python's `re` share the same backtracking semantics for this pattern) against the exact reported name plus a few edge cases (a base name containing a literal dot, the existing Android collision suffix style, a name that already has the right extension), all resolve to the expected base name.

## What I could not verify myself

There's no existing Kotlin unit test setup in this module (only Dart tests), and reproducing this needs an actual Android device/emulator plus a real "create document" system picker interaction, which isn't something I can drive from here. @nicolaspernoud if you're able to test this branch against your actual repro (saving with `fileName: 'trace.gpx'`, `allowedExtensions: ["gpx"]`), that would be the most reliable confirmation before merging. You can point at it with:

```yaml
dependency_overrides:
  android_file_picker:
    git:
      url: https://github.com/vicajilau/flutter_file_picker.git
      ref: fix/android-savefile-unknown-extension-mimetype
      path: packages/file_picker_android
```

## Test plan

- [x] `flutter analyze` clean
- [x] Bumped `android_file_picker` to 1.1.1 with a changelog entry
- [x] Confirmation against the real repro (see above)


packages/file_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.1
+
+- Fixed `saveFile()` garbling the file name for extensions Android doesn't recognize, e.g. saving `trace.gpx` produced `trace.gpx.xml.gpx`. The extension was unknown to the device's `MimeTypeMap`, so the mime type fell back to sniffing the file's content, which detected generic XML and returned `text/xml`. The system's document picker then "corrected" the suggested file name to match that mime type's own default extension, and a later cleanup step meant to append a missing extension appended the original one on top instead of replacing the mismatched one. Extensions unknown to the device now get a wildcard mime type instead of a sniffed one, and the cleanup step now replaces a mismatched extension rather than assuming there is none. [#2199](https://github.com/vicajilau/flutter_file_picker/issues/2199)
```
navaronbracke:
```suggestion
- Fixed `saveFile()` garbling the file name for extensions Android doesn't recognize, e.g. saving `trace.gpx` produced `trace.gpx.xml.gpx`. [#2199](https://github.com/vicajilau/flutter_file_picker/issues/2199)
```

navaronbracke (approved):
LGTM with nit

Issue #2202, opened 2026-09-09
Scope the next major version (13.0.0)


vicajilau:
We agreed in #2197 that PlatformFile.length() returning 0 for both a genuinely empty file and a failed lookup should be fixed by matching lengthSync(), Future<int?> length() instead of Future<int> length(). That is a real breaking change under sound null safety, confirmed with dart analyze against the common int len = await file.length() pattern, not just a warning.

#2197 explicitly deferred it to batch with the next breaking change rather than cut a major on its own. Looked through the current backlog for something to batch it with and found nothing else queued that is actually breaking, most open proposals are additive.

There is a real second candidate already sitting in the codebase though: the parameters deprecated when file_picker 12.0.0 shipped the federated rewrite, never removed.

## Proposal for 13.0.0

1. PlatformFile.length() becomes Future<int?> length(), matching lengthSync(). (#2197)
2. Remove the parameters deprecated since 12.0.0, all on pickFile()/pickFiles()/saveFile()/getDirectoryPath() in packages/file_picker/lib/src/file_picker.dart:
   - allowMultiple, use pickFile() for single file selection
   - withData, use PlatformFile.readAsBytes()
   - withReadStream, use PlatformFile.readAsByteStream()
   - readSequential, use PlatformFile.readAsByteStream()
   - lockParentWindow, use WindowsOptions.lockParentWindow or LinuxOptions.lockParentWindow
   - cancelUploadOnWindowBlur, use WebOptions.cancelUploadOnWindowBlur
   - androidSafOptions, use androidOptions

Both are already decided in isolation, this issue is just to record the scope so we ship them together instead of one major for #2197 alone followed by another one later for the deprecated parameter cleanup.

Open to adding anything else that turns out to need a major, but nothing else in the current backlog looks like it does.

vicajilau:
What do you think @navaronbracke?

navaronbracke:
Yes this looks good for now. Maybe there will be other things that get added, though.

vicajilau:
Scope implemented in #2203.

Also reviewed every other open feature request against this before closing, to check whether anything else warranted riding along in a major bump: #786, #1258, #1705, #1732, #1861, #2036, #2037, #2155. None of them require a breaking change, they can all ship independently as minors whenever they are ready, so there is nothing else to batch here. Closing this as settled.

PR #2203, opened 2026-09-10
Scope 13.0.0: nullable PlatformFile.length() and removal of parameters deprecated since 12.0.0

### LINKED ISSUE #2197: PlatformFile.length() returns 0 when it fails, which is also a valid length
## The problem

`PlatformFile.length()` returns `0` when it cannot determine the length, in all five implementations:

```dart
Future<int> length() async {
  final len = _bytesLength;
  if (len != null && len > 0) return len;
  try {
    return await xFile.length();
  } catch (_) {
    return 0;   // <-- "I do not know" and "the file is empty" become the same value
  }
}
```

A caller cannot tell a genuinely empty file from a failed lookup. Both come back as `0`, with no exception and nothing logged. The interface documents it as "Get the length of the file in bytes", which says nothing about this.

That matters for the obvious things: skipping upload of a "0 byte" file that actually has content and could not be measured, progress bars dividing by a length that is wrong rather than absent, size checks silently passing.

## Why now

`lengthSync()` shipped in `file_picker` 12.2.0 returning `int?`, using `null` to mean "not known without doing I/O". So the two sibling APIs now disagree about how to say the same thing:

```dart
file.lengthSync()   // int?  -> null means "I do not know"
await file.length() // int   -> 0    means "I do not know", or genuinely empty
```

The idiom the docs suggest, `file.lengthSync() ?? await file.length()`, handles the `null` carefully and then walks straight into the `0`.

## Options

1. **`Future<int?> length()`**, matching `lengthSync()`. Consistent, and forces callers to face the case. Breaking for everyone who calls it.
2. **Let it throw**, since it already sits behind `await` and the caller is prepared for failure there. Breaking behaviourally, silent today, exception tomorrow.
3. **Leave the signature, document the `0`.** Not breaking, and not really a fix, since a caller who reads the docs still cannot distinguish the two cases.

I lean towards 1, for the same reason `lengthSync()` returns `int?`: `0` is a real, valid length, so it cannot also be the sentinel. Option 2 has the advantage that failures stop being silent, but it turns a swallowed error into a crash in code that never expected one.

## Not urgent

Either fix is breaking, which means `file_picker_platform_interface` 4.0.0 and a major on all five implementations, right after 12.2.0 went out. This is worth batching with the next breaking change rather than cutting a major for it alone.

Opening it now so the reasoning is recorded while it is fresh, from the discussion in #2187.

vicajilau:
Implements the scope agreed in #2202.

## PlatformFile.length() becomes Future<int?>

Fixes #2197.

`length()` returned `0` both for a genuinely empty file and for a failed lookup, with no way to tell them apart. It now returns `null` in the failure case, matching `lengthSync()`, across the platform interface and all five platform implementations (Android, Darwin, Linux, Web, Windows).

## Removal of parameters deprecated since 12.0.0

Removed from `FilePicker.pickFiles()`/`pickFile()`/`getDirectoryPath()`/`saveFile()`:

- `allowMultiple`, `withData`, `withReadStream`, `readSequential`
- `lockParentWindow` (the top level one, not `WindowsOptions.lockParentWindow`/`LinuxOptions.lockParentWindow`)
- `cancelUploadOnWindowBlur`
- `androidSafOptions`

None of these were ever forwarded to the platform implementations except `androidSafOptions`, which is now folded into `androidOptions` directly (matching what the deprecation message already pointed callers to).

## Versioning

- `file_picker_platform_interface`: 3.3.0 to 4.0.0
- `android_file_picker`, `file_picker_darwin`, `file_picker_linux`, `windows_file_picker`: to 2.0.0
- `file_picker_web`: 3.1.0 to 4.0.0
- `file_picker`: 12.2.0 to 13.0.0

Also updated the example app and its tests, and added a "Migrating to v13" section to the README.

## Test plan

- [x] `melos exec -- flutter analyze .` clean across all 8 packages
- [x] `melos exec -- flutter test` all green across all 8 packages
- [x] `dart format --output=none --set-exit-if-changed .` clean


example/lib/src/file_picker_demo.dart
```diff
@@ -116,8 +112,8 @@ class _FilePickerDemoState extends State<FilePickerDemo> {
             acceptLabel: _acceptLabelFromInput(),
           ),
           darwinOptions: DarwinOptions(acceptLabel: _acceptLabelFromInput()),
-          withData: _withData,
-          androidSafOptions: _androidSafOptionsFromFlags(),
+          androidOptions:
+              _androidSafOptionsFromFlags() ?? const AndroidOptions(),
```
navaronbracke, line 116: Could `_androidSafOptionsFromFlags()` default to returning `const AndroidOptions()` as fallback, so it is a non null result? That would avoid some repetition

navaronbracke (approved):

vicajilau:
Done in 36aaf0e: removed the withData native code paths (Android and iOS), dead since 12.0.0 stopped forwarding that facade parameter. No behavior change, since loadDataToMemory was already permanently false on both platforms, just deleting the now unreachable eager read branches, the loadData() helper, and the bytes wire key.

vicajilau:
Correction: I pushed that directly to main by mistake, reverted it (18ea28b) and moved it to a proper PR instead: #2205

PR #2207, opened 2026-09-11
Fix FilePickerWebOptions.readSequential having no effect

### LINKED ISSUE #2206: file_picker_web: FilePickerWebOptions.readSequential has no effect
### The bug

`FilePickerWebOptions.readSequential` is declared and documented ("Whether to read multiple files sequentially instead of in parallel.") but nothing in `file_picker_web` ever reads it. Setting it to `true` or `false` changes nothing.

`FilePickerWeb._processSelectedFiles` (`packages/file_picker_web/lib/src/file_picker_web.dart`) processes the `FileList` with a single `for` loop that `await`s each file before moving to the next one, unconditionally:

```dart
Future<List<PlatformFile>> _processSelectedFiles(
  FileList files,
  FilePickerWebOptions webOptions,
) async {
  final List<PlatformFile> pickedFiles = [];

  for (int i = 0; i < files.length; i++) {
    final file = files.item(i);
    if (file == null) continue;

    if (webOptions.withReadStream) {
      pickedFiles.add(_createWebPlatformFile(file: file, readStream: _openFileReadStream(file)));
      continue;
    }

    if (!webOptions.withData) {
      pickedFiles.add(_createWebPlatformFile(file: file));
      continue;
    }

    final bytes = await _readSingleFileBytes(file);
    pickedFiles.add(_createWebPlatformFile(file: file, bytes: bytes));
  }

  return pickedFiles;
}
```

There is no branch on `webOptions.readSequential` anywhere, and no parallel code path exists at all today (no `Future.wait`, no concurrent scheduling). The loop is always effectively sequential, regardless of what the caller passes.

### Confirmation

Reproduced the exact loop structure (same `for` + `await`, `readSequential` accepted as a parameter but never consulted, exactly mirroring the real code) against 5 fake files with a 20ms artificial read delay each, tracking the maximum number of concurrent reads observed:

```
readSequential=true  -> maxConcurrency=1, elapsedMs=112
readSequential=false -> maxConcurrency=1, elapsedMs=108
```

Max concurrency is 1 either way, and elapsed time is the same, confirming the flag has zero effect on the current implementation.

### History

This flag predates the 12.0.0 federated rewrite (originally added per the `file_picker` CHANGELOG: "Add `readSequential` flag for web. If `readSequential` is true, order of picked files will be preserved. If flag is false, files will be read parallel."). The parallel code path documented there does not exist in the current `file_picker_web` implementation, it looks like it was dropped during the rewrite without reconnecting the flag.

Note: `withData`/`withReadStream` on the same `FilePickerWebOptions` class are correctly wired and do change behavior, this issue is specific to `readSequential`.

### Expected behavior

`readSequential: false` (the default) should read multiple files concurrently. `readSequential: true` should read them one at a time. Selection order should be preserved in the result either way.

### Proposed fix

Track it in a PR that reads files concurrently (e.g. via `Future.wait`) when `readSequential` is `false`, writing results into an index-based slot so output order matches selection order regardless of concurrency, and keeps the current one-at-a-time loop when `readSequential` is `true`.

vicajilau:
Fixes #2206.

## The bug

`FilePickerWeb._processSelectedFiles` processed the picked `FileList` with a single `for` loop that awaited each file in turn, with no branch on `readSequential` anywhere and no parallel code path at all. The flag never did anything regardless of its value. Confirmed by reproducing the exact loop structure against fake delayed reads: max observed concurrency was 1 whether `readSequential` was `true` or `false`.

`withData`/`withReadStream` on the same `FilePickerWebOptions` class were already correctly wired, this was specific to `readSequential`.

## The fix

Extracted the per-file work into an indexed task and added `runIndexedTasks` (`indexed_task_runner.dart`), a small internal helper that runs those tasks either one at a time or concurrently via `Future.wait`, writing results into index-based slots so the output always matches selection order regardless of completion order or concurrency mode.

`readSequential: true` now genuinely reads one file at a time. `readSequential: false` (the default) reads them concurrently. Selection order is preserved in the result either way.

## Test plan

- [x] Added `indexed_task_runner_test.dart`, a plain Dart test (no browser needed) exercising both modes directly: sequential caps concurrency at 1, concurrent allows more than one in flight, both preserve order even when completion order is reversed
- [x] `melos exec -- flutter analyze .` clean across all 8 packages
- [x] `melos exec -- flutter test` all green across all 8 packages
- [x] `dart format --output=none --set-exit-if-changed .` clean


packages/file_picker_web/lib/src/indexed_task_runner.dart
```diff
@@ -0,0 +1,25 @@
...
+/// ([sequential] `true`) or concurrently ([sequential] `false`).
+///
+/// Returns the results in index order regardless of completion order, so
+/// callers can pick concurrency without also giving up on ordering.
+Future<List<T>> runIndexedTasks<T>(
+  int length,
+  Future<T> Function(int index) task, {
+  required bool sequential,
+}) async {
+  final results = List<T?>.filled(length, null);
```
navaronbracke: I suppose the `null` is to have a ready list so that the `then()` can put the results in the right slot, but can we avoid the cast at the bottom? `whereNotNull()` ?

packages/file_picker_web/lib/src/indexed_task_runner.dart
```diff
@@ -0,0 +1,25 @@
...
+
+  if (sequential) {
+    for (var i = 0; i < length; i++) {
+      results[i] = await task(i);
+    }
+  } else {
+    await Future.wait([
```
navaronbracke: Since Future.wait() returns List<T>, could you just return the result of Future.wait() ?

packages/file_picker_web/lib/src/indexed_task_runner.dart
```diff
@@ -0,0 +1,25 @@
...
+/// Returns the results in index order regardless of completion order, so
+/// callers can pick concurrency without also giving up on ordering.
+Future<List<T>> runIndexedTasks<T>(
+  int length,
+  Future<T> Function(int index) task, {
+  required bool sequential,
+}) async {
+  final results = List<T?>.filled(length, null);
+
+  if (sequential) {
```
navaronbracke, line 13: Could we use an if with return, to avoid the else ?

packages/file_picker_web/lib/src/indexed_task_runner.dart
navaronbracke: Could this abort early when length == 0?

navaronbracke (changes requested):

packages/file_picker_web/lib/src/indexed_task_runner.dart
vicajilau: Good catch, Future.wait() already preserves input order regardless of completion order, so this was unnecessary. Returned it directly now, which also drops the trailing cast.

packages/file_picker_web/lib/src/indexed_task_runner.dart
vicajilau: Done, same fix as above.

packages/file_picker_web/lib/src/indexed_task_runner.dart
vicajilau: Done, early return in the sequential branch instead.

packages/file_picker_web/lib/src/indexed_task_runner.dart
vicajilau: Added an explicit early return for length == 0.

navaronbracke (approved):

Issue #2192, opened 2026-09-01
Add integration_test coverage for the example app


vicajilau:
Follow-up from the discussion in #2188 (comment: https://github.com/miguelpruivo/flutter_file_picker/pull/2188#issuecomment-5490402786).

## What

An `integration_test` suite that actually drives `example/`, presses the pick/save buttons, and asserts the flow completes, running against a real browser/device rather than the headless `flutter_test` harness.

## Why

This repo's CI already proves the example app *builds* on every platform, but nothing currently exercises it at runtime. A regression that compiles fine but breaks the actual pick/save flow (a bad method channel argument, a broken native callback, and so on) would currently only surface from a user report.

## Why this is not the same as the facade constraint smoke test from #2188

That check targets a different failure mode entirely: a compile time mismatch between the platform interface and an implementation, caught by `pub downgrade` + a real build outside the workspace (`example/` can't reproduce that scenario itself, since its dependencies always resolve to the local checkout regardless of downgrade). This issue is about runtime behavior of the current code, not version resolution, so it needs different tooling (`integration_test` against a real browser/device instead of a build step) and would live in the repo differently (likely under `example/integration_test/`).

## Scope to figure out

- Which platform(s) first, web is the most CI-friendly (an actual browser, no emulator/simulator needed).
- How to interact with a native file picker dialog from an automated test (mocking the platform channel defeats the purpose of testing the real native code; not mocking it means scripting an OS-level file dialog, which is the hard part most plugins struggle to automate).

PR #2208, opened 2026-09-12
Add a web integration_test suite for the example app

vicajilau:
## Summary

- Adds a web `integration_test` that taps the "Pick file" button in the example app and lets a real browser file chooser answer it, verified end to end through the actual `<input type=file>`, the plugin code, and the UI (asserts the byte count read back matches a fixture file).
- The test driver attaches its own Chrome DevTools Protocol connection to the same browser `flutter drive` already controls through chromedriver, both to intercept the native file chooser and to issue a real click so Chrome grants the "user activation" the file chooser requires (Flutter's own `tester.tap()` does not count as a trusted gesture).
- Adds a new `web_integration_test` CI job and a `melos run test:integration_web` script for running it locally.

## Why

CI already proves the example app builds on every platform, but nothing exercised it at runtime. A regression that compiles fine but breaks the actual pick flow (a bad method channel argument, a broken native callback) would previously only surface from a user report.

## Scope

This covers Web only, and only the single file pick flow, since Web is the only platform where a headless CI runner can drive a real file chooser without an emulator, simulator, or OS level dialog scripting. Multi pick, save file, folder pick, and non web platforms are left to follow up issues.

Towards #2192

## Test plan

- [x] Ran the new flow locally against a real chromedriver and Chrome, confirmed it passes
- [x] Confirmed it actually catches a real regression (temporarily broke the change handling in `web_file_input_session.dart`, confirmed the test failed, then reverted)
- [x] `flutter analyze .` and `dart format --output=none --set-exit-if-changed .` clean across the workspace
- [x] New `web_integration_test` CI job green on this PR


example/integration_test/pick_file_web_test.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:file_picker_example/src/file_picker_demo.dart';
+import 'package:flutter/material.dart';
```
navaronbracke, line 2: Can we migrate the example app to material_ui in a new PR?

example/test_driver/integration_test.dart
```diff
@@ -0,0 +1,141 @@
...
+/// Uses chromedriver's `GET /sessions` endpoint, which predates the W3C
+/// WebDriver spec but every chromedriver release still serves for
+/// compatibility; it is the only way to read back a session neither started
+/// nor otherwise addressable by this script.
+Future<String> _waitForChromeDriverSession(String chromedriverPort) async {
+  final sessionsUrl = Uri.parse('http://localhost:$chromedriverPort/sessions');
+  final deadline = DateTime.now().add(const Duration(seconds: 30));
+  while (DateTime.now().isBefore(deadline)) {
+    final response = await http.get(sessionsUrl);
+    final sessions = (jsonDecode(response.body) as Map)['value'] as List;
```
navaronbracke: Could this ise if case expressions to decode the nested maps/lists? Then we don't need the casts

navaronbracke (changes requested):

example/integration_test/pick_file_web_test.dart
vicajilau: That's a good idea but out of scope for this PR, which is only about adding the web integration_test suite. Happy to open a separate PR or issue for migrating the example app's Material usage if you want to track it.

example/test_driver/integration_test.dart
vicajilau: Done in 6109730, switched to pattern matching on the decoded JSON so we no longer need the manual Map/List casts.

example/test_driver/integration_test.dart
```diff
@@ -0,0 +1,144 @@
...
+/// Uses chromedriver's `GET /sessions` endpoint, which predates the W3C
+/// WebDriver spec but every chromedriver release still serves for
+/// compatibility; it is the only way to read back a session neither started
+/// nor otherwise addressable by this script.
+Future<String> _waitForChromeDriverSession(String chromedriverPort) async {
+  final sessionsUrl = Uri.parse('http://localhost:$chromedriverPort/sessions');
+  final deadline = DateTime.now().add(const Duration(seconds: 30));
+  while (DateTime.now().isBefore(deadline)) {
+    final response = await http.get(sessionsUrl);
+    if (jsonDecode(response.body) case {'value': List<dynamic> sessions}) {
```
navaronbracke:
```suggestion
    if (jsonDecode(response.body) case {'value': List<Object?> sessions}) {
```

navaronbracke (approved):
LGTM with nit

PR #2214, opened 2026-09-15
UIScene support

### LINKED ISSUE #2212: [iOS 27] Support UIScene
Since iOS 27 / Xcode is now officially available, we should support UIScene

https://docs.flutter.dev/release/breaking-changes/uiscenedelegate

cc @vicajilau

navaronbracke:
This PR adds UIScene support for iOS 27.

It also removes a stale Swift file in the example app and cleans up the plugin setup on iOS / MacOS to reduce some duuplication

Fixes #2212


vicajilau (approved):
LGTM!

navaronbracke:
Xcode 27 only supports iOS 15 and later and iOS 27 and later

PR #2217, opened 2026-09-21
fix(darwin): give each picked item its own temp directory

EdinTech:
## The bug

Picking **multiple items on iOS can silently return nothing**. A single item always works.

## Why

`IOSFilePickerHandler.copyToTemporaryDirectory` derives its destination from the source's `lastPathComponent` alone:

```swift
let destinationURL = URL(fileURLWithPath: NSTemporaryDirectory())
    .appendingPathComponent(sourceURL.lastPathComponent)

if FileManager.default.fileExists(atPath: destinationURL.path) {
    try FileManager.default.removeItem(at: destinationURL)
}
try FileManager.default.copyItem(at: sourceURL, to: destinationURL)
```

That name is not unique. Photos edited in the Photos app are exported under a fixed name, and Live Photos arrive as `.pvt` packages. And these copies run **concurrently** — `loadFileRepresentation` calls its completion handler on an arbitrary queue, once per selected item (`picker(_:didFinishPicking:)`).

So colliding names race:

```
A: fileExists? no  -> starts copying to /tmp/NAME
B: fileExists? yes -> removes /tmp/NAME, deleting A's in-flight copy
A: copyItem throws -> returns nil
B: copyItem throws -> returns nil
```

Both items are then dropped by the `compactMap` over `resolved`. When **every** item loses the race, the handler hits:

```swift
currentResult(orderedResolved.isEmpty ? nil : orderedResolved)
```

`nil` reaches Dart as an empty list — **indistinguishable from the user cancelling**. Nothing is logged, nothing throws.

## Reproduction

On a physical iPhone (iOS 26), with `FileType.image` and `DarwinOptions(assetRepresentationMode: .current)`:

1. Select **two or more Live Photos** → `pickFiles` returns an empty list. No error, no partial result.
2. Select **one** photo → works every time.

A single item has no race to lose, which is why this only shows up with multi-select.

## The fix

Give each item its own freshly created subdirectory, so the copies can no longer observe or delete each other:

```swift
let directoryURL = URL(fileURLWithPath: NSTemporaryDirectory())
    .appendingPathComponent(UUID().uuidString, isDirectory: true)
let destinationURL = directoryURL
    .appendingPathComponent(sourceURL.lastPathComponent)

try FileManager.default.createDirectory(
    at: directoryURL,
    withIntermediateDirectories: true
)
try FileManager.default.copyItem(at: sourceURL, to: destinationURL)
```

The existence check and removal go away with the shared destination — the directory is new every time. The returned path keeps the original file name, so `PlatformFile.name` and the extension are unchanged, and `resolveActualFile` still resolves `.pvt` packages the same way.

`FilePicker.clearTemporaryFiles()` is unaffected: it removes the contents of `NSTemporaryDirectory()` recursively, which now includes these subdirectories.

## Notes

- Native-only change, so per `CONTRIBUTING.md` there are no Dart unit tests to add.
- `packages/file_picker_darwin` bumped to 2.1.1 (patch, bug fix) with a changelog entry.
- Verified in production: we have been running this patch as a fork for our app's photo upload path, where the silent failure was losing users' photo selections.


vicajilau (approved):

vicajilau:
Many thanks for this @EdinTech !!

PR #2218, opened 2026-09-21
Forward initialDirectory and dialogTitle to native on Darwin

### LINKED ISSUE #2216: initialDirectory has no effect, dialog opens at last‑used directory instead of specified path
**Describe the bug**
The `initialDirectory` parameter does not take effect on desktop (Windows). When calling `FilePicker.pickFiles()` with a custom `initialDirectory` path provided, the file picker dialog still opens at the last‑used directory instead of jumping to the specified initial directory.

I have confirmed that the provided directory path exists and is accessible, folder permissions are normal. `flutter clean` and rebuild have been performed, but the problem still persists.

**Platform**
- [ ] Android
- [ ] iOS
- [ ] Web
- [x] Desktop

**Platform OS version**
Mac 26.5.1 (25F80)

**How are you picking?**
```dart
final String targetDir = r"C:\MyWork\Data";
  initialDirectory: targetDir,
  type: FileType.custom,
  allowedExtensions: ['txt', 'json'],
);

### PR DESCRIPTION
**Pull request by vicajilau**
## Summary
- `pickFile()`, `pickFiles()`, and `getDirectoryPath()` in `file_picker_darwin` accepted `initialDirectory` (and `dialogTitle` for `getDirectoryPath()`) but never included them in the method channel arguments map, so they never reached native code.
- The native macOS implementation already read `args["initialDirectory"]`/`args["dialogTitle"]` correctly in all four of its handlers, it was simply never receiving them.
- Added Dart unit tests asserting both arguments are forwarded for `pickFile()`, `pickFiles()`, and `getDirectoryPath()`.
- Bumped `file_picker_darwin` to 2.1.2 (patch, bug fix) with a changelog entry.

Fixes #2216

## Test plan
- [x] `flutter test test/file_picker_darwin_test.dart` (12/12 passing, including the 2 new tests)
- [x] `dart analyze lib test` (no issues)
- [x] `dart format` applied

### CODE REVIEW

File: packages/file_picker_darwin/test/file_picker_darwin_test.dart [hunk 1]
```diff
@@ -115,6 +115,60 @@ void main() {
       expect(receivedLabel, isNull);
     });
 
+    test(
+      'pickFile and pickFiles send dialogTitle and initialDirectory',
+      () async {
+        final picker = FilePickerDarwin();
+        Map? receivedArguments;
```
navaronbracke:
```suggestion
        Map<Object?, Object?>? receivedArguments;
```

packages/file_picker_darwin/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.2
+
+- Fixed `initialDirectory` having no effect on macOS and iOS for `pickFile()`, `pickFiles()`, and `getDirectoryPath()`. The Dart implementation never forwarded `initialDirectory` (and `dialogTitle` for `getDirectoryPath()`) to the native method channel call, so the native side, which already read those arguments correctly, never received them. [#2216](https://github.com/vicajilau/flutter_file_picker/issues/2216)
```
navaronbracke: You don't have to explain the reasoning behind a bug in the changelog :P

```suggestion
- Fixed `initialDirectory` having no effect on macOS and iOS for `pickFile()`, `pickFiles()`, and `getDirectoryPath()`. [#2216](https://github.com/vicajilau/flutter_file_picker/issues/2216)
```

navaronbracke (approved):
LGTM, although I would also cross check Windows / Linux / Android

vicajilau:
Good call @navaronbracke!! I checked all three.

Windows and Linux are fine end to end. `initialDirectory` reaches the actual native call in both (`arena.pcwstr(initialDirectory)` on Windows, the D-Bus request encoding on Linux), and Windows was already confirmed working on a real machine earlier in #2216.

Android has the same Dart side gap this PR fixes for Darwin (`initialDirectory`/`dialogTitle` aren't added to the method channel arguments for pick/getDirectoryPath), but it goes further there. The native Kotlin side never reads `initialDirectory` for `ACTION_OPEN_DOCUMENT`, `ACTION_OPEN_DOCUMENT_TREE`, or the `ACTION_GET_CONTENT` variants either, only `saveFile` supports it today. Making it work on Android means adding native support for turning a plain path into a SAF document tree URI, not just forwarding an argument, so it's a bigger, separate piece of work rather than something this PR should take on. I'll file a follow up issue for that instead of scope creeping this one.

PR #2219, opened 2026-09-21
Document initialDirectory limitation for picking on Android

vicajilau:
## Summary
- Follow up from the review discussion on #2218: Android has a similar looking gap to the Darwin one fixed there, but it goes deeper.
- The native Kotlin side never sets `DocumentsContract.EXTRA_INITIAL_URI` for `ACTION_OPEN_DOCUMENT`, `ACTION_OPEN_DOCUMENT_TREE`, or the `ACTION_GET_CONTENT` variants, so `initialDirectory` has no effect on `pickFile()`, `pickFiles()`, `pickFileAndDirectoryPaths()`, or `getDirectoryPath()`. Only `saveFile()` honors it today.
- Making it work for picking needs native support for turning a plain path into a SAF document tree URI, which isn't a direct mapping, so it's documented as a known limitation for now rather than fixed in this PR.

No code changes, documentation only.


vicajilau:
Wrong target, moved this commit onto #2218 instead.

